PR #6, opened 2024-04-23
Add Inverted Images for Light/Dark Mode

0xnim:


CyanVoxel:
Awesome! The inverted masks, thumbnail border, and highlight border aren't necessary though, as those are only used for their shapes and are never displayed anywhere. Could you add a commit that removes those 4 new files?

PR #24, opened 2024-04-23
Added multiselection adding of tags

ookami125:
This only does the most basic version and doesn't currently handle anything special with multiple tag boxes.


ookami125:
And I borked it. I'm just gonna close this and redo it...

PR #32, opened 2024-04-24
python 3.12 support

Thesacraft:
Enables the program to be run with python3.12 aswell as supporting older version.
Pyside6 is purposly limited to 3.6.3.1 because 3.7 seems to have a bug that breaks the programm(should be [this ](https://bugreports.qt.io/browse/PYSIDE-2675) )that will hopefully be fixed in 3.7.1. 
All other versions are corectly selected by pip
(Tested with 3.10 and 3.12)
it should also work with 3.9.6 because all original versions of the packages are in the range described in this file.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,8 +2,8 @@ humanfriendly==10.0
 opencv_python>=4.8.0.74,<=4.9.0.80
 Pillow==10.3.0
 pillow_avif_plugin>=1.3.1,<=1.4.3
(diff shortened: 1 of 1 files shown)


Trevo525:
It looks like there is already a branch for this. Maybe that would be a better target? 

https://github.com/CyanVoxel/TagStudio/tree/py-3.12

Thesacraft:
I guess but I don't think it's necessary because my edits should allow all versions of python to run without any other custom changes to the

PR #34, opened 2024-04-24
Mutli-Select Tagging

DrRetro2033:
Adds some basic multi-select tagging for Tag Studio

https://youtu.be/xGxPZ0_GBI0

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,12 +6,12 @@
 	"configurations": [
 		{
 			"name": "TagStudio",
-			"type": "debugpy",
+			"type": "python",
 			"request": "launch",
 			"program": "${workspaceRoot}/tagstudio/tag_studio.py",
 			"console": "integratedTerminal",
 			"justMyCode": true,
-			"args": ["--debug"]
+			"args": []
 		}
 	]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,7 +8,7 @@
 			"name": "TagStudio",
 			"type": "debugpy",
 			"request": "launch",
-			"program": "${workspaceRoot}\\TagStudio\\tagstudio.py",
+			"program": "${workspaceRoot}/tagstudio/tag_studio.py",
 			"console": "integratedTerminal",
 			"justMyCode": true,
 			"args": ["--debug"]
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,12 +6,12 @@
 	"configurations": [
 		{
 			"name": "TagStudio",
-			"type": "python",
+			"type": "debugpy",
 			"request": "launch",
 			"program": "${workspaceRoot}\\TagStudio\\tagstudio.py",
 			"console": "integratedTerminal",
 			"justMyCode": true,
-			"args": []
+			"args": ["--debug"]
 		}
 	]
 }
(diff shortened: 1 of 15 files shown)


BananaChad:
https://github.com/CyanVoxel/TagStudio/assets/105674111/2b905a99-d7b8-4b1e-929f-a55d7d014018

works with latest commit(e69e499) using python 3.12.3

CyanVoxel:
A few things I've found so far:
- When applying tags to a multiselected group of entries, if the entries have multiple tag field types (content, meta, etc) and they're in a different order across the selected entries, this will ignore the intended tag field type and instead apply it to a different tag field (maybe the first one it finds?)
- Favorite and Archived badges are not able to be used to apply their respective tags to multiple entries at once, and when applying these tags normally via this multiselection process, the badges are not updated if the favorite/archived tags are applied until clicking on those entries. I believe I have some Qt singals for this, or at the very least some old callbacks that really *should* be signals.
- Bonus: The first time I tried this, I think I tried it with the badges first, and ended up corrupting the library I was using. I think some fields and values got their wires crossed, which caused errors that persist when loading the library again. I'm trying to replicate this, and will provide more information if I'm able to successfully to that.
Pasted image 20240424133137

CyanVoxel:
After further investigation, here's what's going on inside the library file that was corrupted:
multiselection error

I'm still working on reproducing it

DrRetro2033:
> After further investigation, here's what's going on inside the library file that was corrupted: multiselection error
> 
> I'm still working on reproducing it

Alright, thanks for letting me know.

DrRetro2033:
So I rewrote how Multi-Select actually gets the field id, and now it works perfectly (Tested the bugs from earlier commits, and should not corrupt your library now, hopefully.)

CyanVoxel:
Glad to see it working a lot better now! Tags seem to be applied to the correct fields now and no more library corruption ;)

Unfortunately I found a bug involving badges: By selecting one or more entries, then clicking the favorite or archived badge on an *unselected* entry, it instead tries applying the tag badge that was clicked to the *selected entrie(s)* instead of the entry that was clicked.
So the intended behavior would be for any selected entries to have this badge applied when the badge of one of these *selected entries* is clicked, but for entries *outside* of this selection then the clicked badge would only ever apply a tag to whichever entry is clicked (which is the current behavior in the main branch).

Less immediately important, but I'm also noticing the thumbnails re-rendering whenever a badge updates, instead of just calling the `update_badges` method on the ItemThumbs. This also helped me find a bug on the main branch where badges don't update after fields are removed, but that's not a result of this PR.

Feel free to keep the PR open or convert it to a draft - closing and reopening it messed with my git and I had to re-clone the repo in order to test the branch out again :)

DrRetro2033:
So I fixed the badge problem, and it works just fine now. Although, I did notice that when removing the meta field instead of removing the tag, the badges on the ItemThumbs do not update and they require a second click to hide them.

CyanVoxel:
Really excellent work!! This is an incredible feature to finally have implemented! I'm so sorry for the mess of my code that needed sorting out in order for this to work... I really appreciate you doing this though, and continuing to keep coming back with new fixes!

In the meantime, stress testing this PR has led me to discover a couple preexisting bugs regarding the multi-selection metadata system, one being the mentioned "field deletion not updating the badge" (#56), and a much more critical one (#55) that can cause library corruption. Anyway, I'm mentioning these because I'm going to investigate the cause of those before pulling in this branch, since this makes significant changes to the same system. I plan on pulling this one first before any potential fix to the outside bugs, however I just wanted to update you on why I haven't pulled it in the meantime. Thanks again for all your hard work on this!

PR #35, opened 2024-04-24
Add NixOS Dev Environment Support

dakota-marshall:
- Add necessary flake files to allow for a NixOS development environment.
- Update the `TagStudio.sh` bash script to use `/usr/bin/env bash` instead of `/bin/bash` to allow NixOS to run it.
- Update `README.md` with documentation on running the Nix environment.

Can now run `nix develop` on Nix, and then can run the TagStudio script and the application will launch as expected. 

Let me know if anything needs fixing!

_Note to any Nix people: The way I setup the environment is probably needlessly obtuse, but I got the app launching. This can probably be a lot simpler if someone has a smarter way of doing it._


dakota-marshall:
Broken when attempting to use python 3.12 after the update, will fix then it will be ready again.

dakota-marshall:
Had to update the pinned hash to use QT 6.6.3 instead, and installed python 3.12 into the environment, now working as expected.

PR #39, opened 2024-04-24
Update library.py to not duplicate default tags on save

### LINKED ISSUE #38: [BUG] Favorite and Archive tags are duplicated inside library files
As the title says, duplicate tag data for the built-in tags are accruing over time inside of library files. This is likely due to a failure in the check system for built-in tags, thinking that the built-in tags are missing from the library when they're not.
image

Thesacraft:
Fixes #38 
You checked for default tags before adding any tags to check to it (in the to_json method).
Also fixed wrong return type of verify_default_tags


CyanVoxel:
Looks like that fixes the issue! Thanks for the PR!

PR #40, opened 2024-04-24
🏷️ Add strict typing dicts for ts_library.json

chao-master:
Wip:
Adds a new file `json_typing.py` that includes `TypedDict` declarations for the json file format
Then uses those thoughout core where nessesary.

The main benefit is strict typing checking then making and reading the json

Unfinished work:
 - `fields`, I couldn't work out that is being stored or loaded here
 - `macros`, I can't actually tell if these are used

### DIFF
--- a/tagstudio/src/core/json_typing.py
+++ b/tagstudio/src/core/json_typing.py
@@ -1,0 +1,33 @@
+from typing import TypedDict
+
+class Json_Libary(TypedDict("",{"ts-version":str})):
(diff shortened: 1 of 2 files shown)


CyanVoxel (approved):
Looks good! I'll go ahead and pull this, but in the future make sure to use PascalCase/CapWords for Python class names, as described the the [PEP 8 Style Guide](https://peps.python.org/pep-0008/#class-names).

PR #42, opened 2024-04-25
Allow files to be opened in their respective programs in the background for Windows, MacOS, and Linux

xarvex:
Uses subprocess.Popen() along with start/open/xdg-open to open a file in its respective program. This process will last even after the TagStudio process exits.

Note: this has only been tested on Linux!


xarvex:
Did not realize previous commit was a part of this, will create new PR.

PR #58, opened 2024-04-26
Add folders to tags tool

Creepler13:
Adds a new button to the tools menu that creates tags based on the folder structure and applies them to entries

### DIFF
--- a/tagstudio/src/core/ts_core.py
+++ b/tagstudio/src/core/ts_core.py
@@ -31,11 +31,11 @@ SPREADSHEET_TYPES: list[str] = ['csv', 'xls', 'xlsx', 'numbers', 'ods']
 PRESENTATION_TYPES: list[str] = ['ppt', 'pptx', 'key', 'odp']
 ARCHIVE_TYPES: list[str] = ['zip', 'rar', 'tar', 'tar.gz', 'tgz', '7z']
 PROGRAM_TYPES: list[str] = ['exe', 'app']
-SHORTCUT_TYPES: list[str] = ['lnk', 'desktop']
+SHORTCUT_TYPES: list[str] = ['lnk', 'desktop', 'url']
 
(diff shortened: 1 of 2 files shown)


CyanVoxel (changes requested):
A very nice feature! I've just got a few requested changes after testing it out:

- Move the feature from the “Tools” menu to the “Macros” menu. This feature is similar to how Macros will function in the future, so it’ll make more sense to have it there. 
- Change the default Tag color to be none ("") instead of black.
- Add a confirmation dialog before performing the action. Something that explains to the user what is about to happen to their library, and gives them the option to cancel if they don’t want it performed. Ideally it could preview how many and which tags are about to be created, but that’s not strictly necessary.
- Instead of having a tag with no name for the root folder, instead either give it the same name as the root folder or allow the option to omit this root tag entirely.
	- Also I’m not sure what the intended behavior for this empty tag is supposed to be, since the other subfolder tags inherit from each other but not this one. If this tag was never intended, then that’s fine as well.
	
*Feel free to keep the PR open or convert it to a draft, either way just please don't close it and re-open it with the same PR number :)*

Creepler13:
I made it more complicated that needed but its a good visualisation of what happening

- Moved the feature from the “Tools” menu to the “Macros” menu
- Changed the default Tag color to be none ("") instead of black
- Added a confirmation dialog before performing the action. That displays what is going to happen (might need a makeover, i'm  not good with UI)
- Fixed the empty Tag for the root folder

CyanVoxel (approved):
Thanks for making the changes! Great work on this feature, and don't worry about the UI - the whole app needs a pass to make the UI more usable and consistent. I really appreciate the time and effort, and having a full preview of the tags being created is awesome! 👍 

I'll go ahead and pull this, but for the future just make sure that variable names (the `Apply_button` specifically) are using `snake_case`, although from the rest of your code it looks like that was just a one-off that slipped though.

PR #61, opened 2024-04-26
A simple Tag Database.

DrRetro2033:
To access, go to Edit > Tag Database.
image


Loran425 (commented):
All looks to function as intended, 
an expected side effect is that it gets a bit slower as the number of tags increases 
loading took a few seconds with 10,000 tags and editing a tag then saving in that scenario took almost a couple of minutes but this would be a good tool to support viewing the tag database.

Issues with searching or adding to search seem to be an issue with the TagWidget or possibly the original TagSearch Modal

CyanVoxel:
While I have a more comprehensive Tag panel/tab planned, this is a great stopgap until that's implemented. Performance issues with large numbers of tags likely stem from the TagSearch Modal implementation, which populates the screen with *many* TagWidgets.

PR #69, opened 2024-04-27
Ability to open file or file location using context menu

cirillom:
Added the ability to open the file or view it in the file explorer using a context menu or by clicking directly on path.
Should work with in Unix, but I didn't test it (my linux died).

https://github.com/TagStudioDev/TagStudio/assets/39923113/b3e9f636-6caa-4026-8621-aa3c0921811d


CyanVoxel (approved):
Looks pretty good! The only real issue I have is the way the file explorer is selected and ran on Windows vs other platforms, which currently wouldn't work on Linux or macOS due to the nautilus command plus the use of `os.startfile`, which is a Windows-only specific command.

Normally I would request changes for the PR, but we're working on pulling in the rest of the new feature changes today in order to refactor this specific file. I'll likely open an issue for it after the refactor if you'd like to address it once the classes are broken up into separate files.

PR #72, opened 2024-04-27
improve code readability

yedpodtrzitko:
Hi,
this PR aims to simplify some pieces of code to improve readability. Beside that there's a few tiny fixes for potential bugs (see my comments on given lines)

### DIFF
--- a/tagstudio/src/cli/ts_cli.py
+++ b/tagstudio/src/cli/ts_cli.py
@@ -1556,7 +1556,7 @@ class CliDriver:
 
 				print(self.format_title(title))
 				
-				if len(self.filtered_entries) > 0:
+				if self.filtered_entries:
 					# entry = self.lib.get_entry_from_index(
 					# 	self.filtered_entries[index])
 					entry = self.lib.get_entry(self.filtered_entries[index][1])
@@ -1581,7 +1581,7 @@ class CliDriver:
 
 					self.print_fields(self.filtered_entries[index][1])
 				else:
-					if len(self.lib.entries) > 0:
+					if self.lib.entries:
 						print(self.format_h1('No Entry Results for Query', color=BRIGHT_RED_FG))
 						self.set_external_preview_default()
 					else:
--- a/tagstudio/src/cli/ts_cli.py
+++ b/tagstudio/src/cli/ts_cli.py
@@ -493,7 +493,7 @@ class CliDriver:
 		# print(f'Char Limit: {char_limit}, Len: {len(text)}')
 		return char_limit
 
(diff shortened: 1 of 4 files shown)


tagstudio/src/cli/ts_cli.py
```diff
@@ -2069,13 +2071,13 @@ def scr_choose_option(self, subtitle: str, choices: list, prompt: str = '', requ
 					# 	# self.cleanup()
 					# 	sys.exit()
 					# Cancel =======================================================
-					if (com[0].lower() == 'cancel' or com[0].lower() == 'c' or com[0] == '0') and required==False:
+					if com_name in ('cancel', 'c', '0') and not required:
 						clear()
 						return -1
 					# Selection ====================================================
-					elif int(com[0]) > 0 and int(com[0]) <= len(choices):
+					elif com_name.isdigit() and 0 < int(com_name) <= len(choices):
```
yedpodtrzitko, line 2078: here I added condition `.isdigit()`, otherwise it could end up with `ValueError` when calling `int()` on non-numeric string.

tagstudio/src/cli/ts_cli.py
```diff
@@ -2625,14 +2627,14 @@ def scr_edit_entry_text(self, entry_index, field_index, allow_newlines=True, cle
 			self.lib.update_entry_field(
 				entry_index, field_index, new_content.rstrip('\n').rstrip('\r'), 'replace')
 
-	def scr_list_tags(self, query: str = '', tag_ids: list[int] = [], clear_scr=True) -> None:
+	def scr_list_tags(self, query: str = '', tag_ids: list[int] = None, clear_scr=True) -> None:
```
yedpodtrzitko, line 2630: if the empty list as a default parameter isnt intentional, then it might be causing some issues, see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

tagstudio/src/qt/ts_qt.py
```diff
@@ -4601,7 +4601,7 @@ def expand_collation(self, collation_entries: list[tuple[int, int]]):
 						 for x in collation_entries])
 		# self.update_thumbs()
 
-	def get_frame_contents(self, index=0, query=str):
+	def get_frame_contents(self, index=0, query: str = None):
```
yedpodtrzitko, line 4604: I assume `query=str` was supposed to be `query:str`, right? And since it's after a parameter with a default value, it needs some default value as well.

CyanVoxel: Yes, that looks to be like the original intention. Surprised my linter didn't pick that up...

PR #78, opened 2024-04-28
Build Script for windows

Thesacraft:
Adds a build script for a windows executable
It intentionally doesn't hide the console for the exe because windows defender then complains that it is a malicious programm.
Also i oriented the build script of of the mac os build script from @williamtcastro

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,4 @@ PySide6_Addons>=6.5.1.1,<=6.6.3.1
 PySide6_Essentials>=6.5.1.1,<=6.6.3.1
 typing_extensions>=3.10.0.0,<=4.11.0
 ujson>=5.8.0,<=5.9.0
+Pyinstaller
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,4 +7,4 @@ PySide6_Addons>=6.5.1.1,<=6.6.3.1
 PySide6_Essentials>=6.5.1.1,<=6.6.3.1
 typing_extensions>=3.10.0.0,<=4.11.0
 ujson>=5.8.0,<=5.9.0
-Pyinstaller
+Pyinstaller==6.6.0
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -7,3 +7,4 @@ PySide6_Addons>=6.5.1.1,<=6.6.3.1
 PySide6_Essentials>=6.5.1.1,<=6.6.3.1
 typing_extensions>=3.10.0.0,<=4.11.0
 ujson>=5.8.0,<=5.9.0
+Pyinstaller
```
williamtcastro, line 10:
```suggestion
Pyinstaller==6.6.0
```

Thesacraft: I'll do that tomorrow thanks

Thesacraft:
So im considering to change --onefile to --onedir because this reduces the start process from 12s to under 2s but the disadvantage to that is that, we have a executable and a directory (and not just 1 file) that have to be in the same parent directory for it to work.
Maybe someone can give me input on this

williamtcastro (approved):
Its a great optiion to have a faster start but if is possible to run as a option for the user witch it could build either a portable one or a folder structure

williamtcastro (approved):

CyanVoxel:
So I was able to get it to build the .exe, but it won't launch and I get this error:
```
Error loading Python DLL 'F:\GitHub\TagStudio\build\TagStudio\_internal\python312.dll'.
LoadLibrary: The specified module could not be found.
```
I have no "_internal" folder inside the build directory. Is there something I should be doing differently?

CyanVoxel:
> So I was able to get it to build the .exe, but it won't launch and I get this error:
> 
> ```
> Error loading Python DLL 'F:\GitHub\TagStudio\build\TagStudio\_internal\python312.dll'.
> LoadLibrary: The specified module could not be found.
> ```
> 
> I have no "_internal" folder inside the build directory. Is there something I should be doing differently?

Ah, I needed to launch the .exe in the dist folder, not the build folder...

PR #79, opened 2024-04-28
Copy and Paste + Shortcuts

Creepler13:
Added buttons to copy and paste the fields of entries and their shortcuts (Ctrl+C, Ctrl+V)

It would be good to add the ability to unselected selected entries by clicking at the background or other elements in order to avoid accidents.


tagstudio/src/qt/ts_qt.py
```diff
@@ -356,6 +356,34 @@ def start(self):
...
+                QtCore.Qt.Key.Key_C,
+            )
+        )
+        copy_entry_fields_action.setToolTip("Ctrl+C")
+        edit_menu.addAction(copy_entry_fields_action)
+
+        paste_entry_fields_action = QAction("&Paste", menu_bar)
+        paste_entry_fields_action.triggered.connect(
+            lambda: self.paste_entry_fields_action_callback()
+        )
```
yedpodtrzitko, line 375:
```suggestion
        paste_entry_fields_action.triggered.connect(
            self.paste_entry_fields_action_callback
        )
```

tagstudio/src/qt/ts_qt.py
```diff
@@ -841,6 +869,20 @@ def run_macro(self, name: str, entry_id: int):
                             mode="replace",
                         )
 
+    def copy_entry_fields_action_callback(self):
+        for item in self.selected.__reversed__():
+            if item[0] == ItemType.ENTRY:
+                entry = self.lib.get_entry(item[1])
```
yedpodtrzitko, line 875: - `__reversed__` is a private method, so ideally we shouldnt use it
- `item` can be unpacked directly into `item_type` and `item_id`, so it's more obvious what values we're working with

```suggestion
        for item_type, item_id in reversed(self.selected):
            if item_type == ItemType.ENTRY:
                entry = self.lib.get_entry(item_id)
```

tagstudio/src/qt/ts_qt.py
```diff
@@ -841,6 +869,20 @@ def run_macro(self, name: str, entry_id: int):
...
+                self.copied_fields = entry.fields.copy()
+
+    def paste_entry_fields_action_callback(self):
+        if self.copied_fields != None:
+            for item in self.selected:
+                if item[0] == ItemType.ENTRY:
+                    entry = self.lib.get_entry(item[1])
```
yedpodtrzitko, line 882: similar as above
```suggestion
            for item_type, item_id in self.selected:
                if item_type == ItemType.ENTRY:
                    entry = self.lib.get_entry(item_id)
```

yedpodtrzitko:
when I select multiple items, it copies the fields from the first item selected only. Is that intended behaviour? Eg.

selected item 1:
tags: [foo, bar]
metatags: []

selected item 2:
tags: [aaa]
metatags: [bbb]


target item:
tags: [foo, bar]
metatags: []

yedpodtrzitko (commented):
nice done overall, but there's a few things which could improve the code readability.

Creepler13:
> when I select multiple items, it copies the fields from the first item selected only. Is that intended behaviour? Eg.
> 
> selected item 1: tags: [foo, bar] metatags: []
> 
> selected item 2: tags: [aaa] metatags: [bbb]
> 
> target item: tags: [foo, bar] metatags: []

I intended to copy the last item from the selection, but forgot the break

CyanVoxel:
I've made some significant changes to this, including:
- Handling state of menu text and action availability
- Fix entanglement with shallow copying
- Fix paste overwriting existing data
- Multi-selection copying now merges all fields of selected entries in the buffer instead of only copying the last entry in the selection, which is now reflected in action command text.

I'll be labeling this as "review needed" to get some extra eyes on these changes. While there can be some funkiness with the copy paste behavior at times, I believe most of that is due to preexisting bugs with the field system and just the general awkwardness of it compared to the upcoming changes to it. If there are any bugs specifically related to my changes here then I'll be happy to address them.

I think the copy/paste actions should also be in the context menu for the ItemThumbs themselves, however I most likely won't be able to implement that until tomorrow.

PR #80, opened 2024-04-28
Qt refactor

Loran425:
Refactoring work of ts_qt.py

Pardon the messy start gotta copy everything over through some rather odd methods to maintain history. 
This will be trimmed down a lot

Final intention is for widgets/modals/helpers to be imported to the __init__.py of each folder so that something like this works
```python 
from src.qt.helpers import open_file 
```
rather than 
```python
from src.qt.helpers.open_file import open_file
```


Loran425:
Some of the larger remaining functions or modals are tied pretty closely with the driver, rather than lock this all down while I take a look at it tomorrow this should be good to check for merge.

tagstudio/src/qt/modals/fix_unlinked.py
```diff
@@ -0,0 +1,179 @@
...
+from PySide6.QtWidgets import QWidget, QVBoxLayout, QHBoxLayout, QLabel, QPushButton
+
+from src.core.library import Library
+from src.qt.helpers import FunctionIterator, CustomRunnable
+from src.qt.modals import DeleteUnlinkedEntriesModal, RelinkUnlinkedEntries
+from src.qt.widgets	import ProgressWidget
+
+# Only import for type checking/autocompletion, will not be imported at runtime.
+if typing.TYPE_CHECKING:
+	from src.qt.ts_qt import QtDriver
```
Loran425, line 19: This type of importing appears to be a way around the way around the apparent coupling between the Widget and the QtDriver as the widget doesn't need to know anything about the driver, only the IDE for type checking and auto completion

tagstudio/src/qt/modals/fix_unlinked.py
```diff
@@ -0,0 +1,179 @@
...
+WARNING = f'[WARNING]'
+INFO = f'[INFO]'
+
+logging.basicConfig(format="%(message)s", level=logging.INFO)
+
+
+class FixUnlinkedEntriesModal(QWidget):
+	# done = Signal(int)
+	def __init__(self, library:'Library', driver:'QtDriver'):
+		super().__init__()
```
Loran425, line 32: Only side effect is that the type hint has to be a string type, this appears to be fairly well accepted though so I don't see an issue with this side effect.

Issue #81, opened 2024-04-28
Fix Open In File Manager for Linux


Icosahunter:
The current implementation assumes the file manager is Nautilus. As a Linux fallback we can use "xdg-open" passing in the parent folder for the file in question, this won't select the file but it will open the folder the file is located in, in a file manager agnostic way. It looks like KDE Dolphin supports the "--select" option so we can add support for that too.
Also I recommend changing the tooltip from "open in explorer" to "open in file manager" or "open in file browser" so the language is platform agnostic.

xarvex:
It looks like this also causes a freeze, I assume because a blocking API being used. I will have a fix ready shortly.

PR #84, opened 2024-04-28
Allow opening and selecting files in file managers for all platforms

xarvex:
Fixes: #81
Refactored file opener helper to use already present open_file
Refactored open_file to allow option for file manager

For Linux systems, should use default file manager.

**IMPORTANT: only tested on Linux!**

This as well prevents a freeze from the previous implementation that used subprocess.run, which is a blocking operation.

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -83,7 +83,7 @@ def open_file(path: str, file_manager: bool = False):
 				command_args = [path]
 				if file_manager:
 					# will reveal in Finder
-					command_args = command_args.append("-R")
+					command_args.append("-R")
 			else:
 				if file_manager:
 					command_name = "dbus-send"
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -76,7 +76,7 @@ def open_file(path: str, file_manager: bool = False):
 				command_name = "start"
 				# first parameter is for title, NOT filepath
 				command_args = ["", normpath]
-			subprocess.Popen([command_name] + command_args, shell=True, close_fds=True, creationflags=subprocess.DETACHED_PROCESS | subprocess.CREATE_NEW_PROCESS_GROUP | subprocess.CREATE_BREAKAWAY_FROM_JOB)
+			subprocess.Popen([command_name] + command_args, shell=True, close_fds=True, creationflags=subprocess.CREATE_NEW_PROCESS_GROUP | subprocess.CREATE_BREAKAWAY_FROM_JOB)
 		else:
 			if sys.platform == "darwin":
 				command_name = "open"
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -65,20 +65,38 @@ logging.basicConfig(format="%(message)s", level=logging.INFO)
 QSettings.setPath(QSettings.IniFormat, QSettings.UserScope, os.getcwd())
 
 
-def open_file(path: str):
+def open_file(path: str, file_manager: bool = False):
 	try:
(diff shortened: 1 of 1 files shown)


tagstudio/src/qt/ts_qt.py
```diff
@@ -65,20 +65,38 @@
...
-			subprocess.Popen(["start", "", os.path.normpath(path)], shell=True, close_fds=True, creationflags=subprocess.DETACHED_PROCESS)
+			normpath = os.path.normpath(path)
+			if file_manager:
+				command_name = "explorer"
+				command_args = [f"/select,{normpath}"]
+			else:
+				command_name = "start"
+				# first parameter is for title, NOT filepath
+				command_args = ["", normpath]
+			subprocess.Popen([command_name] + command_args, shell=True, close_fds=True, creationflags=subprocess.DETACHED_PROCESS)
```
cirillom, line 79: `creationflags=subprocess.DETACHED_PROCESS` makes the file manager open behind TS.

Like this:
image

Removing it makes the Windows Explorer open in front of TS, which, in my opinion, makes more sense.

Also, thanks for refactoring my code, didn't realize that there was an open_file function already. Great job!

cirillom (commented):
Tested and validated on Windows, works almost as expected. Except that **I** think that the file manager should open in front of TS, not behind it.

tagstudio/src/qt/ts_qt.py
xarvex: DETACHED_PROCESS is actually what makes the window keep open after TagStudio closes, which is especially something intended whenever opening in its respective program. These windows should not suddenly close when TagStudio is closed.
I'm not sure if there's any additional Windows flag for that, though.

xarvex: What happens when `subprocess.DETACHED_PROCESS | subprocess.CREATE_NEW_PROCESS_GROUP | subprocess.CREATE_BREAKAWAY_FROM_JOB` is used instead? That should further isolate these program launches.

xarvex: Resolved in 737be61.

tagstudio/src/qt/ts_qt.py
```diff
@@ -65,20 +65,38 @@
...
+				# first parameter is for title, NOT filepath
+				command_args = ["", normpath]
+			subprocess.Popen([command_name] + command_args, shell=True, close_fds=True, creationflags=subprocess.CREATE_NEW_PROCESS_GROUP | subprocess.CREATE_BREAKAWAY_FROM_JOB)
 		else:
 			if sys.platform == "darwin":
 				command_name = "open"
+				command_args = [path]
+				if file_manager:
+					# will reveal in Finder
+					command_args = command_args.append("-R")
```
williamtcastro, line 86: I tested on macOS and this was causing the command_args to become `None` due to the append function return breaking the string concatenation.

By doing this changed it works at intended


`# subprocess.Popen([command] + command_args, close_fds=True)
                     ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: can only concatenate list (not "NoneType") to list`
 

```suggestion
					command_args.append("-R")
```

williamtcastro (changes requested):

tagstudio/src/qt/ts_qt.py
xarvex: Oh whoops, my LSP acts up in areas where it thinks code is impossible to run, which happens in these kind of blocks. Thank you for spotting this!

cirillom (approved):
Working perfectly on Windows 11!

williamtcastro (approved):

williamtcastro (approved):
Working perfectly on macOS test with local and remote libraries

PR #87, opened 2024-04-28
display all tags by default in Add Tag window

yedpodtrzitko:
When I want to add a new tag to item, the current flow is slightly suboptimal with some unnecessary friction.

For now, this PR displays all tags in the "Add Tag" window by default instead of the mandatory step of typing something into the search field to show filtered results.

Ideally I would also like to see all available fields in the right sidebar where "Add Field" is, so I dont have to open new dialog for it. Does it sound like something what's compatible with your vision?

If not, then at least I have suggestion for the  "Add Field" window. Right now I have to expand the dropdown with all available fields first. That's one extra click in comparison with a list of all available fields displayed as buttons or checkboxes or whatever UI element is available in PySide. Ideally this would allow me to select multiple Fields at the same time (so maybe the checkboxes?).


Loran425:
#80 Has moved the preview panel to https://github.com/TagStudioDev/TagStudio/blob/c9c399faa9c492d91d2388662b3aea3b4d989fb8/tagstudio/src/qt/widgets/preview_panel.py#L1-L15

Sorry for any headache on this

PR #88, opened 2024-04-28
Add testing support

Hidorikun:
Added a way to run all tests. 
Added tests for the remove punctuation utility as well as refactored it. 
Moved constants in separate folder because they caused a circular dependency between ts_core and library.

### DIFF
--- a/tagstudio/src/core/library.py
+++ b/tagstudio/src/core/library.py
@@ -16,10 +16,10 @@ import xml.etree.ElementTree as ET
 from enum import Enum
 import ujson
 
-from src.core.json_typing import JsonCollation, JsonEntry, JsonLibary, JsonTag
-from src.core import ts_core
-from src.core.utils.str import strip_punctuation
-from src.core.utils.web import strip_web_protocol
+from tagstudio.src.core.json_typing import JsonCollation, JsonEntry, JsonLibary, JsonTag
+from tagstudio.src.core.constants import TS_FOLDER_NAME, BACKUP_FOLDER_NAME, COLLAGE_FOLDER_NAME, VERSION, TEXT_FIELDS
+from tagstudio.src.core.utils.str import strip_punctuation
+from tagstudio.src.core.utils.web import strip_web_protocol
 
 TYPE = ['file', 'meta', 'alt', 'mask']
 # RESULT_TYPE = Enum('Result', ['ENTRY', 'COLLATION', 'TAG_GROUP'])
@@ -154,8 +154,6 @@ class Entry:
 			self.fields[field_index][field_id] = sorted(tags, key=lambda t: library.get_tag(t).display_name(library))
 
 		# logging.info(f'Tags: {self.fields[field_index][field_id]}')
-		
-
 
 class Tag:
 	"""A Library Tag Object. Referenced by ID."""
@@ -546,8 +544,8 @@ class Library:
 		path = os.path.normpath(path).rstrip('\\')
 
 		# If '.TagStudio' is included in the path, trim the path up to it.
-		if ts_core.TS_FOLDER_NAME in path:
-			path = path.split(ts_core.TS_FOLDER_NAME)[0]
+		if TS_FOLDER_NAME in path:
+			path = path.split(TS_FOLDER_NAME)[0]
 
 		try:
 			self.clear_internal_vars()
@@ -565,11 +563,11 @@ class Library:
 		"""Verifies/creates folders required by TagStudio."""
 
 		full_ts_path = os.path.normpath(
-			f'{self.library_dir}/{ts_core.TS_FOLDER_NAME}')
+			f'{self.library_dir}/{TS_FOLDER_NAME}')
 		full_backup_path = os.path.normpath(
-			f'{self.library_dir}/{ts_core.TS_FOLDER_NAME}/{ts_core.BACKUP_FOLDER_NAME}')
+			f'{self.library_dir}/{TS_FOLDER_NAME}/{BACKUP_FOLDER_NAME}')
 		full_collage_path = os.path.normpath(
-			f'{self.library_dir}/{ts_core.TS_FOLDER_NAME}/{ts_core.COLLAGE_FOLDER_NAME}')
+			f'{self.library_dir}/{TS_FOLDER_NAME}/{COLLAGE_FOLDER_NAME}')
 
 		if not os.path.isdir(full_ts_path):
 			os.mkdir(full_ts_path)
@@ -606,13 +604,13 @@ class Library:
 		path = os.path.normpath(path).rstrip('\\')
 
 		# If '.TagStudio' is included in the path, trim the path up to it.
-		if ts_core.TS_FOLDER_NAME in path:
-			path = path.split(ts_core.TS_FOLDER_NAME)[0]
+		if TS_FOLDER_NAME in path:
+			path = path.split(TS_FOLDER_NAME)[0]
 
-		if os.path.exists(os.path.normpath(f'{path}/{ts_core.TS_FOLDER_NAME}/ts_library.json')):
+		if os.path.exists(os.path.normpath(f'{path}/{TS_FOLDER_NAME}/ts_library.json')):
 
 			try:
-				with open(os.path.normpath(f'{path}/{ts_core.TS_FOLDER_NAME}/ts_library.json'), 'r', encoding='utf-8') as f:
+				with open(os.path.normpath(f'{path}/{TS_FOLDER_NAME}/ts_library.json'), 'r', encoding='utf-8') as f:
 					json_dump: JsonLibary = ujson.load(f)
 					self.library_dir = str(path)
 					self.verify_ts_folders()
@@ -788,9 +786,9 @@ class Library:
 		if return_code == 1:
 
 			if not os.path.exists(os.path.normpath(
-					f'{self.library_dir}/{ts_core.TS_FOLDER_NAME}')):
+					f'{self.library_dir}/{TS_FOLDER_NAME}')):
 				os.makedirs(os.path.normpath(
-					f'{self.library_dir}/{ts_core.TS_FOLDER_NAME}'))
+					f'{self.library_dir}/{TS_FOLDER_NAME}'))
 
 			self._map_filenames_to_entry_ids()
 
@@ -832,7 +830,7 @@ class Library:
 		Used in saving the library to disk.
 		"""
 
-		file_to_save: JsonLibary = {"ts-version": ts_core.VERSION,
+		file_to_save: JsonLibary = {"ts-version": VERSION,
 				  	"ignored_extensions": [],
 						"tags": [],
 						"collations": [],
@@ -869,7 +867,7 @@ class Library:
 
 		self.verify_ts_folders()
 
-		with open(os.path.normpath(f'{self.library_dir}/{ts_core.TS_FOLDER_NAME}/{filename}'), 'w', encoding='utf-8') as outfile:
+		with open(os.path.normpath(f'{self.library_dir}/{TS_FOLDER_NAME}/{filename}'), 'w', encoding='utf-8') as outfile:
 			outfile.flush()
 			ujson.dump(self.to_json(), outfile, ensure_ascii=False, escape_forward_slashes=False)
 			# , indent=4 <-- How to prettyprint dump
@@ -886,7 +884,7 @@ class Library:
 		filename = f'ts_library_backup_{datetime.datetime.utcnow().strftime("%F_%T").replace(":", "")}.json'
 
 		self.verify_ts_folders()
-		with open(os.path.normpath(f'{self.library_dir}/{ts_core.TS_FOLDER_NAME}/{ts_core.BACKUP_FOLDER_NAME}/{filename}'), 'w', encoding='utf-8') as outfile:
+		with open(os.path.normpath(f'{self.library_dir}/{TS_FOLDER_NAME}/{BACKUP_FOLDER_NAME}/{filename}'), 'w', encoding='utf-8') as outfile:
 			outfile.flush()
 			ujson.dump(self.to_json(), outfile, ensure_ascii=False, escape_forward_slashes=False)
 		end_time = time.time()
@@ -932,11 +930,11 @@ class Library:
 		# Scans the directory for files, keeping track of:
 		#   - Total file count
 		#   - Files without library entries
-		# for type in ts_core.TYPES:
+		# for type in TYPES:
 		start_time = time.time()
 		for f in glob.glob(self.library_dir + "/**/*", recursive=True):
 			# p = Path(os.path.normpath(f))
-			if ('$RECYCLE.BIN' not in f and ts_core.TS_FOLDER_NAME not in f 
+			if ('$RECYCLE.BIN' not in f and TS_FOLDER_NAME not in f 
        				and 'tagstudio_thumbs' not in f and not os.path.isdir(f)):
 				if os.path.splitext(f)[1][1:].lower() not in self.ignored_extensions:
 					self.dir_file_count += 1
@@ -2034,7 +2032,7 @@ class Library:
 		# entry = self.entries[entry_index]
 		entry = self.get_entry(entry_id)
 		field_type = self.get_field_obj(field_id)['type']
-		if field_type in ts_core.TEXT_FIELDS:
+		if field_type in TEXT_FIELDS:
 			entry.fields.append({int(field_id): ''})
 		elif field_type == 'tag_box':
 			entry.fields.append({int(field_id): []})
--- a/tagstudio/src/core/ts_core.py
+++ b/tagstudio/src/core/ts_core.py
@@ -7,50 +7,8 @@
 import json
 import os
 
-from src.core.library import Entry, Library
-
-VERSION: str = '9.2.0'  # Major.Minor.Patch
-VERSION_BRANCH: str = 'Alpha'  # 'Alpha', 'Beta', or '' for Full Release
-
-# The folder & file names where TagStudio keeps its data relative to a library.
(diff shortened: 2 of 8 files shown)


tagstudio/src/core/library.py
```diff
@@ -16,10 +16,10 @@
...
 import ujson
 
-from src.core.json_typing import JsonCollation, JsonEntry, JsonLibary, JsonTag
-from src.core import ts_core
-from src.core.utils.str import strip_punctuation
-from src.core.utils.web import strip_web_protocol
+from tagstudio.src.core.json_typing import JsonCollation, JsonEntry, JsonLibary, JsonTag
+from tagstudio.src.core.constants import TS_FOLDER_NAME, BACKUP_FOLDER_NAME, COLLAGE_FOLDER_NAME, VERSION, TEXT_FIELDS
+from tagstudio.src.core.utils.str import strip_punctuation
+from tagstudio.src.core.utils.web import strip_web_protocol
```
Loran425, line 22: `Tagstudio.src...` import format breaks at runtime for vscode and when running from the terminal

```suggestion
from src.core.json_typing import JsonCollation, JsonEntry, JsonLibary, JsonTag
from src.core.constants import TS_FOLDER_NAME, BACKUP_FOLDER_NAME, COLLAGE_FOLDER_NAME, VERSION, TEXT_FIELDS
from src.core.utils.str import strip_punctuation
from src.core.utils.web import strip_web_protocol
```

tagstudio/src/core/ts_core.py
```diff
@@ -7,50 +7,8 @@
...
-
-TAG_COLORS = ['', 'black', 'dark gray', 'gray', 'light gray', 'white', 'light pink',
-	      'pink', 'red', 'red orange', 'orange', 'yellow orange', 'yellow', 
-		  'lime', 'light green', 'mint', 'green','teal', 'cyan', 'light blue', 
-		  'blue', 'blue violet', 'violet', 'purple', 'lavender', 'berry', 
-		  'magenta', 'salmon', 'auburn', 'dark brown', 'brown', 'light brown', 
-		  'blonde', 'peach', 'warm gray', 'cool gray', 'olive']
-
+from tagstudio.src.core.constants import TEXT_FIELDS
+from tagstudio.src.core.library import Entry, Library
```
Loran425, line 11: `Tagstudio.src...` import format breaks at runtime for vscode and when running from the terminal

```suggestion
from src.core.constants import TEXT_FIELDS
from src.core.library import Entry, Library
```

tagstudio/src/qt/ts_qt.py
```diff
@@ -36,16 +36,17 @@
...
 										SHORTCUT_TYPES, PROGRAM_TYPES, ARCHIVE_TYPES, PRESENTATION_TYPES,
 										SPREADSHEET_TYPES, DOC_TYPES, AUDIO_TYPES, VIDEO_TYPES, IMAGE_TYPES,
 										LIBRARY_FILENAME, COLLAGE_FOLDER_NAME, BACKUP_FOLDER_NAME, TS_FOLDER_NAME,
 										VERSION_BRANCH, VERSION)
-from src.core.utils.web import strip_web_protocol
-from src.qt.flowlayout import FlowLayout, FlowWidget
-from src.qt.main_window import Ui_MainWindow
+from tagstudio.src.core.utils.web import strip_web_protocol
+from tagstudio.src.qt.flowlayout import FlowLayout, FlowWidget
+from tagstudio.src.qt.main_window import Ui_MainWindow
```
Loran425: `Tagstudio.src...` import format breaks at runtime for vscode and when running from the terminal

tagstudio/tests/main.py
Loran425, line 1: I'm having luck getting `python.exe -m unittest discover tagstudio.tests -t "tagstudio"` to run from the `<local_repo>` directory but this file is throwing a fit working between all the different use cases. 
Something like `unittest.TestSuite()` or [unittest load test protocol](https://docs.python.org/3/library/unittest.html#load-tests-protocol) might be more appropriate for the longer term and avoid needing to change the import format in the src library.

requirements.txt
```diff
@@ -7,3 +7,4 @@ PySide6_Addons>=6.5.1.1,<=6.6.3.1
 PySide6_Essentials>=6.5.1.1,<=6.6.3.1
 typing_extensions>=3.10.0.0,<=4.11.0
 ujson>=5.8.0,<=5.9.0
+parametrized==0.9.0
```
Loran425, line 10:
```suggestion
parameterized==0.9.0
```

tagstudio/src/qt/ts_qt.py
```diff
@@ -27,21 +27,22 @@
 from PySide6.QtWidgets import (QApplication, QWidget, QHBoxLayout, QPushButton, QLineEdit, QScrollArea, QFileDialog,
 							   QSplashScreen, QMenu)
 from humanfriendly import format_timespan
-
-from src.core.library import ItemType
-from src.core.ts_core import (PLAINTEXT_TYPES, TagStudioCore, TAG_COLORS, DATE_FIELDS, TEXT_FIELDS, BOX_FIELDS, ALL_FILE_TYPES,
+from tagstudio.src.core.library import Collation, Entry, ItemType, Library, Tag
+from tagstudio.src.core.palette import ColorType, get_tag_color
+from tagstudio.src.core.ts_core import TagStudioCore
+from tagstudio.src.core.constants import (PLAINTEXT_TYPES, TAG_COLORS, DATE_FIELDS, TEXT_FIELDS, BOX_FIELDS, ALL_FILE_TYPES,
```
Loran425, line 33: `Tagstudio.src...` import format breaks at runtime for vscode and when running from the terminal

```suggestion
from src.core.library import Collation, Entry, ItemType, Library, Tag
from src.core.palette import ColorType, get_tag_color
from src.core.ts_core import TagStudioCore
from src.core.constants import (PLAINTEXT_TYPES, TAG_COLORS, DATE_FIELDS, TEXT_FIELDS, BOX_FIELDS, ALL_FILE_TYPES,
```

tagstudio/src/qt/ts_qt.py
```diff
@@ -27,21 +27,22 @@
...
-from src.qt.main_window import Ui_MainWindow
-from src.qt.helpers import FunctionIterator, CustomRunnable
-from src.qt.widgets import CollageIconRenderer, ThumbRenderer, PanelModal, ProgressWidget, PreviewPanel, ItemThumb
-from src.qt.modals import (BuildTagPanel, TagDatabasePanel, FileExtensionModal, FixUnlinkedEntriesModal,
+from tagstudio.src.core.utils.web import strip_web_protocol
+from tagstudio.src.qt.flowlayout import FlowLayout
+from tagstudio.src.qt.main_window import Ui_MainWindow
+from tagstudio.src.qt.helpers import FunctionIterator, CustomRunnable
+from tagstudio.src.qt.widgets import CollageIconRenderer, ThumbRenderer, PanelModal, ProgressWidget, PreviewPanel, ItemThumb
+from tagstudio.src.qt.modals import (BuildTagPanel, TagDatabasePanel, FileExtensionModal, FixUnlinkedEntriesModal,
```
Loran425, line 43: `Tagstudio.src...` import format breaks at runtime for vscode and when running from the terminal

```suggestion
from src.core.utils.web import strip_web_protocol
from src.qt.flowlayout import FlowLayout
from src.qt.main_window import Ui_MainWindow
from src.qt.helpers import FunctionIterator, CustomRunnable
from src.qt.widgets import CollageIconRenderer, ThumbRenderer, PanelModal, ProgressWidget, PreviewPanel, ItemThumb
from src.qt.modals import (BuildTagPanel, TagDatabasePanel, FileExtensionModal, FixUnlinkedEntriesModal,
```

tagstudio/src/qt/ts_qt.py
```diff
@@ -27,21 +27,22 @@
...
 						   FixDupeFilesModal, FoldersToTagsModal)
-import src.qt.resources_rc
+import tagstudio.src.qt.resources_rc
```
Loran425, line 45: `Tagstudio.src...` import format breaks at runtime for vscode and when running from the terminal

```suggestion
import src.qt.resources_rc
```

tagstudio/tests/src/core/test_tags.py
```diff
@@ -1,12 +1,14 @@
-from src.core.library import Tag
+import unittest
 
+from tagstudio.src.core.library import Tag
```
Loran425, line 3: `Tagstudio.src...` import format breaks at runtime for vscode and when running from the terminal

```suggestion
from src.core.library import Tag
```

tagstudio/tests/src/utils/test_str.py
```diff
@@ -0,0 +1,16 @@
+import unittest
+
+from parameterized import parameterized
+
+from tagstudio.src.core.utils.str import strip_punctuation
```
Loran425, line 5: `Tagstudio.src...` import format breaks at runtime for vscode and when running from the terminal

```suggestion
from src.core.utils.str import strip_punctuation
```

tagstudio/src/qt/ts_qt.py
```diff
@@ -27,21 +27,22 @@
 from PySide6.QtWidgets import (QApplication, QWidget, QHBoxLayout, QPushButton, QLineEdit, QScrollArea, QFileDialog,
 							   QSplashScreen, QMenu)
 from humanfriendly import format_timespan
-
-from src.core.library import ItemType
-from src.core.ts_core import (PLAINTEXT_TYPES, TagStudioCore, TAG_COLORS, DATE_FIELDS, TEXT_FIELDS, BOX_FIELDS, ALL_FILE_TYPES,
+from tagstudio.src.core.library import Collation, Entry, ItemType, Library, Tag
+from tagstudio.src.core.palette import ColorType, get_tag_color
```
Loran425, line 31: This line was removed in a previous commit and is no longer used in `ts_qt.py`

tagstudio/src/qt/ts_qt.py
```diff
@@ -27,21 +27,22 @@
 from PySide6.QtWidgets import (QApplication, QWidget, QHBoxLayout, QPushButton, QLineEdit, QScrollArea, QFileDialog,
 							   QSplashScreen, QMenu)
 from humanfriendly import format_timespan
-
-from src.core.library import ItemType
-from src.core.ts_core import (PLAINTEXT_TYPES, TagStudioCore, TAG_COLORS, DATE_FIELDS, TEXT_FIELDS, BOX_FIELDS, ALL_FILE_TYPES,
+from tagstudio.src.core.library import Collation, Entry, ItemType, Library, Tag
```
Loran425, line 30: This line was previously reduced to only import ItemType as it is the only item used in `ts_qt.py`

Loran425 (changes requested):
Glad to see someone interested in the testing side of thing, many new projects (or newly open source projects) suffer due to lack of testing, and having robust tests makes everyone's lives easier. 

A few other changes that would be needed before this could be considered for a merge.
The following files all reference `ts_core.py` looking for items now in the `src.core.constants` file.

tagstudio\src\qt\modals\build_tag.py
tagstudio\src\qt\widgets\collage_icon.py
tagstudio\src\qt\widgets\item_thumb.py
tagstudio\src\qt\widgets\preview_panel.py
tagstudio\src\qt\widgets\thumb_renderer.py

requirements.txt
yedpodtrzitko, line 10: this shouldnt be a _production_ depedency, I'd suggest to create `requirements-dev.txt` or similar and put it there instead.

yedpodtrzitko:
just out of curiosity - why `unittest` instead of `pytest`? The latter is considered defacto the go-to testing library. It has fixtures and parametrization support out of the box, and it's less verbose, eg. no need to wrap every single test into a class, and you can use plain `assert` instead of writing `self.assertEqual` etc. everywhere.

requirements.txt
Hidorikun: That's a good point. By the way, do you guys manually add those dependencies to the requirements.txt or can pip do that automtically like npm can?

Hidorikun:
@yedpodtrzitko I was looking for some native support for testing in Python and came across unittest. 

My thinking was that I would want to avoid adding extra dependencies if I can. 
Saying that, to create a parameterized test I also needed to add a dependency, so if pytest can do that and much more, I agree that it would be a better alternative. 

Thank you for pointing it out. I am not accustomed with the defacto conventions in python projects so far, and I appreciate the input.

requirements.txt
yedpodtrzitko: There's `pip freeze` which lists all currently installed packages, so you can do `pip freeze > requirements.txt` but it lists even the transitional dependencies which is suboptimal.

Ideally the project should use some advanced dep. management tool like Poetry or pdm, but seems like that isnt/wasnt a priority yet

Loran425: Fair point, 
I don't think there's currently a plan to switchover to a more robust dependency manager, not saying a switch shouldn't happen but I believe the energy is more focused on swapping the backend from a JSON file to a database. 

Are there any recommendations aside from something like a requirements-dev.txt file to be a stopgap on this kind of thing?

Example
```
-r requirements.txt
parameterized==0.9.0
```

yedpodtrzitko: I'd go with `requirements-dev.txt` and `requirements-dev.txt` without referencing each other, any dev will know how to deal with it. But I dont have strong opinion about it.

tagstudio/src/core/ts_core.py
Hidorikun: Thank you for the feedback! I have made the changes.

tagstudio/src/qt/ts_qt.py
Hidorikun: Thank you for the feedback! I have made the changes.

tagstudio/tests/main.py
Hidorikun: I have added pytest instead of unittest and I found that running `python -m pytest tests/` from the tagstudio folder works. Does not seem to work any other way though.

requirements.txt
Hidorikun: Thank you for the feedback! I have added a requirements-dev.txt as discussed in this MR with pytest.

tagstudio/tests/src/core/test_tags.py
Hidorikun: Thank you for the feedback! I have made the changes.

tagstudio/tests/src/utils/test_str.py
Hidorikun: Thank you for the feedback! I have made the changes.

tagstudio/src/qt/ts_qt.py
Hidorikun: Thank you for the feedback! I have made the changes.

tagstudio/src/qt/ts_qt.py
Hidorikun: Thank you for the feedback! I have made the changes.

requirements.txt
yedpodtrzitko: I'd keep the original name `requirements.txt` instead of adding the `-prod` suffix there, as that's the most common name everyone looks for by default. Other than that it looks good to me, nice job.

Hidorikun: Done, thank you for the review 😄

tagstudio/src/core/ts_core.py
```diff
@@ -7,51 +7,9 @@
 import json
 import os
 
+from src.core.constants import TEXT_FIELDS
```
Loran425, line 10: Also needs `TS_FOLDER_NAME`

```suggestion
from src.core.constants import TEXT_FIELDS, TS_FOLDER_NAME
```

tagstudio/src/qt/ts_qt.py
```diff
@@ -323,7 +323,7 @@ def start(self):
 		# self.entry_panel.setWindowIcon(icon)
 
 		if os.name == 'nt':
-			appid = "cyanvoxel.tagstudio.9"
+			appid = "cyanvoxel.9"
```
Loran425, line 326: not sure if this was intentional or a replace all gone wrong.

Loran425 (changes requested):
Great work, I'm only seeing these two comments that should be addressed.

tagstudio/src/qt/ts_qt.py
Hidorikun: It was a replace gone wrong, good catch. I reverted it.

requirements-dev.txt
```diff
@@ -0,0 +1,10 @@
+humanfriendly==10.0
+opencv_python>=4.8.0.74,<=4.9.0.80
+Pillow==10.3.0
+pillow_avif_plugin>=1.3.1,<=1.4.3
+PySide6>=6.5.1.1,<=6.6.3.1
+PySide6_Addons>=6.5.1.1,<=6.6.3.1
+PySide6_Essentials>=6.5.1.1,<=6.6.3.1
+typing_extensions>=3.10.0.0,<=4.11.0
+ujson>=5.8.0,<=5.9.0
```
yedpodtrzitko, line 9:
```suggestion
```

There was probably some misunderstanding, but these dependencies should remain only in `requirements.txt`, otherwise these files will get out of sync sooner or later and it will cause problems.

If you really want to have everything installed via a single dependency file, put here `-r requirements.txt` isntead. Otherwise I think both these files should be independent from each other. Ie. `requirements.txt` for installing production dependencies, `requirements-dev.txt` for developers' dependencies.

gabrieljreed:
Just wanted to add my 2 cents here - this looks great! One small thing I might change is that I usually see the `tests` folder live outside the source code directory (in this case, the `tagstudio` folder.) Everything inside the source code folder should be code for the package, since tests are not actual application code, they can be separate. This is actually recommended in the [pytest documentation](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code) as well.

Issue #7, opened 2024-04-23
[BUG] Refresh Directory adds Current files onto the old ones, no clearing of the old files.


0xnim:
When you open a directory in Application and create a new library it adds the files but when you try to refresh it adds them on top of the old files.

After one refresh and a new file between adding and refreshing:
![image](https://github.com/CyanVoxel/TagStudio/assets/57798165/0dd97253-9f0c-402c-a36f-a844208f4dbe)
After another refresh:
![image](https://github.com/CyanVoxel/TagStudio/assets/57798165/9a3dec66-93f8-4624-98c3-542131bb6e65)

As you can see the old files are not clearing.

CyanVoxel:
Does the number of files listed in the status bar at the bottom also go up with each refresh?

![image](https://github.com/CyanVoxel/TagStudio/assets/46939827/40477de8-3b13-4a7d-ab33-1ef496e7dda6)

0xnim:
Yes, the number goes up.

0xnim:
I have determined it does not get added to the file. Instead just a frontend issue.

ShadowJonathan:
Same issue here, it just basically duplicates the library

CyanVoxel:
Fixed in #89

PR #89, opened 2024-04-29
fix: strip leading slash in library items

yedpodtrzitko:
I dont know where exactly the issue occurs, but for now this is good enough to fix the issue.

The issue is where files are inserted into library with leading slash, eg. "/photo.jpg" instead of "photo.jpg". That's because when `entry.path` is empty, the result is just the slash + filename.

Then when I refresh the library, `photo.jpg` isnt there apparently there (only `/photo.jpg`), so it gets inserted again as a duplicity.


CyanVoxel (approved):
Fixes #7! This has been a longstanding issue I was never able to replicate myself. Thanks for the fix!

PR #99, opened 2024-04-29
added action to close a library

TechCrafterGaming:
I don't know if there is anything to change about the way I currently reset the window.
I've tried to get some idea of the function of reopening a Libaray when one was already open.


CyanVoxel (changes requested):
Pretty good! I think the only thing that's missing from the `close_library()` method is a couple lines that set the location the last opened library, found in the `shutdown()` method:
```py
self.settings.setValue("last_library", self.lib.library_dir)
self.settings.sync()
```

CyanVoxel (approved):

PR #103, opened 2024-04-30
show all tags in Add Tags panel by default

yedpodtrzitko:
Right now when I want to add new tag to file via the "Add Tag" panel window, I have to trigger update of tags manually via typing into the search input. Is there any actual reason behind this? If I have a trivial number of tags, I want to see them straight away without typing anything. That's the change this PR introduces.

Beside that the empty value for `self.first_tag_id` is now `None` instead of `-1` to see better it's actually non-tag value, plus it will fail loudly in case the it would get passed somewhere it shouldnt.

### DIFF
--- a/tagstudio/src/core/library.py
+++ b/tagstudio/src/core/library.py
@@ -1577,7 +1577,8 @@ class Library:
 
 		# NOTE: I'd expect a blank query to return all with the other implementation, but
 		# it misses stuff like Archive (id 0) so here's this as a catch-all.
-		if query == '':
+		query = query.strip()
+		if not query:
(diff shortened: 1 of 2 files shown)


Loran425 (commented):
Code works and and behave well up to 1000 tags on my system @CyanVoxel this seems like your call on what the behavior should be for this. 
A think a better method for the DB system would be to allow the user to set default or most frequent tags but this works for at least modest-moderate tag libraries (<1500 tags has about a 0.5-1 second delay loading the tag database view on my machine)

CyanVoxel (approved):

PR #105, opened 2024-04-30
Add "Delete Tag" menu option to tags in the Tag Database and tag box

Darxoon:
I added this menu option to the tag database and tag boxes. It will prompt the user and if they say yes, delete the tags and remove all references to it

grafik

One thing I added is that when you hold shift while clicking Delete Tag, the prompt and popup after it was deleted will be skipped. I figured this would be useful when deleting a lot of tags at the same time.

### DIFF
--- a/tagstudio/src/qt/modals/tag_database.py
+++ b/tagstudio/src/qt/modals/tag_database.py
@@ -4,7 +4,7 @@
 
 
 from PySide6.QtCore import Signal, Qt, QSize
-from PySide6.QtWidgets import QWidget, QVBoxLayout, QHBoxLayout, QLineEdit, QScrollArea, QFrame
+from PySide6.QtWidgets import QWidget, QVBoxLayout, QHBoxLayout, QLineEdit, QScrollArea, QFrame, QMessageBox, QApplication
 
 from src.core.library import Library
 from src.qt.widgets import PanelWidget, PanelModal, TagWidget
@@ -92,8 +92,10 @@ class TagDatabasePanel(PanelWidget):
 				l = QHBoxLayout(c)
 				l.setContentsMargins(0,0,0,0)
 				l.setSpacing(3)
-				tw = TagWidget(self.lib, self.lib.get_tag(tag_id), True, False)
+				tag_deletable = tag_id not in [0, 1] # [0, 1] should probably be extracted as a constant during refactor
+				tw = TagWidget(self.lib, self.lib.get_tag(tag_id), True, False, tag_deletable)
 				tw.on_edit.connect(lambda checked=False, t=self.lib.get_tag(tag_id): (self.edit_tag(t.id)))
+				tw.on_delete.connect(lambda checked=False, t=self.lib.get_tag(tag_id): (self.delete_tag(t.id)))
 				l.addWidget(tw)
 				self.scroll_layout.addWidget(c)
 		else:
@@ -106,8 +108,10 @@ class TagDatabasePanel(PanelWidget):
 				l = QHBoxLayout(c)
 				l.setContentsMargins(0,0,0,0)
 				l.setSpacing(3)
-				tw = TagWidget(self.lib, tag, True, False)
+				tag_deletable = tag_id not in [0, 1] # [0, 1] should probably be extracted as a constant during refactor
+				tw = TagWidget(self.lib, tag, True, False, tag_deletable)
 				tw.on_edit.connect(lambda checked=False, t=tag: (self.edit_tag(t.id)))
+				tw.on_delete.connect(lambda checked=False, t=tag: (self.delete_tag(t.id)))
 				l.addWidget(tw)
 				self.scroll_layout.addWidget(c)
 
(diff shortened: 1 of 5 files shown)


tagstudio/src/qt/modals/tag_database.py
```diff
@@ -92,8 +92,9 @@ def update_tags(self, query:str):
 				l = QHBoxLayout(c)
 				l.setContentsMargins(0,0,0,0)
 				l.setSpacing(3)
-				tw = TagWidget(self.lib, self.lib.get_tag(tag_id), True, False)
+				tw = TagWidget(self.lib, self.lib.get_tag(tag_id), True, False, True)
```
Loran425, line 95: Should probably remove the ability to delete "System Tags" (archived & favorite) since TagStudio recreates those tags on system shutdown if they are missing. 
```suggestion
				if tag_id in [0, 1]:  # [0, 1] should probably be extracted as a constant during refactor
					tag_deletable = False
				else:
					tag_deletable = True
				tw = TagWidget(self.lib, self.lib.get_tag(tag_id), True, False, tag_deletable )
```

tagstudio/src/qt/modals/tag_database.py
```diff
@@ -106,8 +107,9 @@ def update_tags(self, query:str):
 				l = QHBoxLayout(c)
 				l.setContentsMargins(0,0,0,0)
 				l.setSpacing(3)
-				tw = TagWidget(self.lib, tag, True, False)
+				tw = TagWidget(self.lib, tag, True, False, True)
```
Loran425, line 110: Should probably remove the ability to delete "System Tags" (archived & favorite) since TagStudio recreates those tags on system shutdown if they are missing. 
```suggestion
				if tag.id in [0, 1]:  # [0, 1] should probably be extracted as a constant during refactor
					tag_deletable = False
				else:
					tag_deletable = True
				tw = TagWidget(self.lib, self.lib.get_tag(tag_id), True, False, tag_deletable )
```

tagstudio/src/qt/widgets/tag_box.py
```diff
@@ -120,6 +126,24 @@ def edit_tag(self, tag_id:int):
...
+		
+		tag = self.lib.get_tag(tag_id)
+		shift_held = Qt.KeyboardModifier.ShiftModifier in QApplication.keyboardModifiers()
+		
+		if shift_held or show_delete_prompt():
+			self.lib.remove_tag(tag_id)
+			self.update_tags()
+			
+			if not shift_held:
+				QMessageBox.information(self, "Delete Tag", "Tag deleted.")
```
Loran425, line 146: Having to replicate this here and in tag_database makes me think there should be a more reusable way to implement these dialogs but I think that will be a whole codebase sweep that needs to happen so I don't think that's an issue for this PR.

tagstudio/src/qt/widgets/tag_box.py
```diff
@@ -71,36 +71,42 @@ def __init__(self, item, title, field_index, library:Library, tags:list[int], dr
...
 			# tw = TagWidget(self.lib, self.lib.get_tag(tag), True, True, 
 			# 							on_remove_callback=lambda checked=False, t=tag: (self.lib.get_entry(self.item.id).remove_tag(self.lib, t, self.field_index), self.updated.emit()), 
 			# 							on_click_callback=lambda checked=False, q=f'tag_id: {tag}': (self.driver.main_window.searchField.setText(q), self.driver.filter_items(q)),
 			# 							on_edit_callback=lambda checked=False, t=tag: (self.edit_tag(t))
 			# 							)
-			tw = TagWidget(self.lib, self.lib.get_tag(tag), True, True)
-			tw.on_click.connect(lambda checked=False, q=f'tag_id: {tag}': (self.driver.main_window.searchField.setText(q), self.driver.filter_items(q)))
-			tw.on_remove.connect(lambda checked=False, t=tag: (self.remove_tag(t)))
-			tw.on_edit.connect(lambda checked=False, t=tag: (self.edit_tag(t)))
+			tw = TagWidget(self.lib, self.lib.get_tag(tag_id), True, True, True)
```
Loran425, line 92: Should probably remove the ability to delete "System Tags" (archived & favorite) since TagStudio recreates those tags on system shutdown if they are missing. 
```suggestion
				if tag_id in [0, 1]:  # [0, 1] should probably be extracted as a constant during refactor
					tag_deletable = False
				else:
					tag_deletable = True
				tw = TagWidget(self.lib, self.lib.get_tag(tag_id), True, False, tag_deletable )
```

Loran425 (changes requested):
Looks good overall but there are some guardrails for system tags that I think should be implemented to prevent errors. 

Also Highlights some other refactoring that should probably happen in the tag_database.py file and further supports the use use of a constants file recommended in #88

Archina:
Would it be possible to add something like a reactive store to handle updates of the tags and similar structures instead of pushing signals back and forth.

This would relax the coupling between the TagDatabase and the TagWidget. It would also make it easier to listen to such changes in other places, like under the image preview in the main tag browser.

However, I am more of a Frontend Dev than a python dev.

tagstudio/src/qt/widgets/tag_box.py
Darxoon: I considered putting this function in a place where I could reuse it without repeating myself. the other event listeners don't do this either though so I chose to be consistent with them

Darxoon:
I implemented your suggestions that system tags can't be deleted now

CyanVoxel:
Thank you for the work put into this PR, but unfortunately due to the significant library backend changes in #332 I'll have to close this as it's no longer compatible. I apologize for the long turnaround time and and lack of input on my side about this.

#569 is currently taking up the task of implementing this feature.

PR #106, opened 2024-05-01
Video Player

DrRetro2033:
https://youtu.be/sWr5EN5hRI8
Added a simple video player for Tag Studio. Currently, the player will automatically mute itself when switching to another video. This is by design, as playing a really loud video before someone is ready will probably cause heart-attacks.

Also, I noticed the big refactor, and I am not fully familiar with the structure. I currently have VideoPlayer imported like this.
image
Could someone tell me the correct way of importing my video player in preview_panel.py? Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -251,4 +251,7 @@ compile_commands.json
 .TagStudio
 TagStudio.ini
 
+
+
+
 # End of https://www.toptal.com/developers/gitignore/api/visualstudiocode,python,qt
--- a/Build_win.bat
+++ b/Build_win.bat
@@ -1,0 +1,31 @@
+@echo off
+set TAGSTUDIO_NAME=TagStudio
+set TAGSTUDIO_DIR=tagstudio
+set TAGSTUDIO_DIR_RESOURCES=%TAGSTUDIO_DIR%/resources
+set TAGSTUDIO_ICON=%TAGSTUDIO_DIR%/resources/icon.ico
+set TAGSTUDIO_SRC=%TAGSTUDIO_DIR%/src
+set TAGSTUDIO_MAIN=%TAGSTUDIO_DIR%/tag_studio.py
+set BUILD_MODE=--onedir
+
+
+if "%1" == "--help" (
+    echo run "%~nx0" for normal Build
+    echo run "%~nx0 --portable" for Build packaged into one file
+    goto end
+)
+if "%1" == "--portable" (
+    echo Building portable executable...
+    set BUILD_MODE=--onefile
+    goto run
+)
+if not "%1" == "" (
+    echo Invalid argument run "%~nx0 --help" for help
+    goto end
+)
+:run
+echo Building executable...
+set COMMAND=PyInstaller --name "%TAGSTUDIO_NAME%" --icon "%TAGSTUDIO_ICON%" --add-data "%TAGSTUDIO_DIR_RESOURCES%:./resources" --add-data "%TAGSTUDIO_SRC%:./src" -p "%TAGSTUDIO_DIR%" --console %BUILD_MODE% "%TAGSTUDIO_MAIN%" -y
+call .venv\Scripts\activate.bat
+%COMMAND%
+deactivate
+:end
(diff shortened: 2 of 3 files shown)


tagstudio/src/qt/widgets/video_player.py
```diff
@@ -0,0 +1,205 @@
...
+				self.hover_fix_timer.start(10)
+		elif obj == self.video_preview and event.type() == QEvent.Type.GraphicsSceneHoverLeave or event.type() == QEvent.Type.HoverLeave:
+			if not self.video_preview.isUnderMouse():
+				self.hover_fix_timer.stop()
+				self.releaseMouse()
+
+
+		return super().eventFilter(obj, event)
+	
+	def checkIfStillHovered(self) -> None:
```
yedpodtrzitko, line 117: please use `snake_case` instead
```suggestion
	def check_if_still_hovered(self) -> None:
```

yedpodtrzitko:
I dont have strong opinion about the feature itself, and it's not up to me to decide, but the [README mentions the code should follow pep-8](https://github.com/TagStudioDev/TagStudio?tab=readme-ov-file#can-i-contribute). That means the camelCases names in code should rather be snake_cases.

tagstudio/src/qt/widgets/__init__.py
```diff
@@ -10,4 +10,5 @@
 from .tag_box import TagBoxWidget
 from .text import TextWidget
 from .item_thumb import ItemThumb
-from .preview_panel import PreviewPanel
\ No newline at end of file
+from .preview_panel import PreviewPanel
+from .video_player import VideoPlayer
```
Loran425, line 14: All good here

tagstudio/src/qt/widgets/video_player.py
Loran425, line 1: Might want to change some of these `logging.info` calls to `logging.debug` call to avoid overwhelming the other terminal information.

DrRetro2033:
> I dont have strong opinion about the feature itself, and it's not up to me to decide, but the [README mentions the code should follow pep-8](https://github.com/TagStudioDev/TagStudio?tab=readme-ov-file#can-i-contribute). That means the camelCases names in code should rather be snake_cases.
Alright, thanks for telling me.

tagstudio/src/qt/widgets/preview_panel.py
```diff
@@ -23,6 +23,7 @@
 from src.qt.modals import AddFieldModal
 from src.qt.widgets import (ThumbRenderer, FieldContainer, TagBoxWidget, TextWidget, PanelModal, EditTextBox,
 							EditTextLine, ItemThumb)
+from .video_player import VideoPlayer
```
Loran425, line 26:
```suggestion
from src.qt.widgets import (ThumbRenderer, FieldContainer, TagBoxWidget, TextWidget, PanelModal, EditTextBox,
							EditTextLine, ItemThumb, VideoPlayer)
```
Should just need to add `VideoPlayer` to the `src.qt.widgets` import on line 24-25

Loran425:
Code review aside, I'm leaning towards a thumbnail with a play button on it being the default and then clicking it starting the preview but I'm open to what the community wants, I'd just be worried about responsiveness in remote storage cases which might not be the typical setup.

DrRetro2033:
> Code review aside, I'm leaning towards a thumbnail with a play button on it being the default and then clicking it starting the preview but I'm open to what the community wants, I'd just be worried about responsiveness in remote storage cases which might not be the typical setup.

tagstudio/src/qt/widgets/preview_panel.py
DrRetro2033: Tried that, but python gave me a circular dependency error.

tagstudio/src/qt/widgets/__init__.py
Loran425: I didn't think about the import order, if preview panel relies on video player, then the video player needs to be before the preview panel in the __init__ file
```suggestion
from .video_player import VideoPlayer
from .preview_panel import PreviewPanel
```

tagstudio/src/qt/widgets/preview_panel.py
Loran425: Tossed in another comment, I didn't think about import order, `VideoPlayer` needs to be imported in the `__init__` file before `PreviewPanel`

PR #107, opened 2024-05-01
add code formatting & Github Action via ruff

yedpodtrzitko:
The current PR is quite huge, as the code has totally different styling all across the places (mixing tabs/spaces, single/double quotes, inconsistent spacing etc.), but having unified coding style is benefit in a long term.

Currently there are no enforced code styling rules other than a mild recommendation "something like pep8" in README. 
Nitpicking and bikeshedding about code formatting in PRs is not very productive, and if something can be automated then let's automate it. 

This PR is adding a job for automatic check of code formatting. If the style doesnt follow what ruff expects, it will show failing badge in PR. See example here: https://github.com/yedpodtrzitko/TagStudio/pull/1

But to make the action pass, we need to format all existing files to match the style.

To format the code, dev only needs to run `ruff format` before commiting their changes.

### DIFF
--- a/.github/workflows/ruff.yaml
+++ b/.github/workflows/ruff.yaml
@@ -1,0 +1,11 @@
+name: Ruff
+on: [ push, pull_request ]
+jobs:
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: chartboost/ruff-action@v1
+        with:
+          version: 0.4.2
+          args: 'format --check'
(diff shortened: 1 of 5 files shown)


CyanVoxel (changes requested):
Could the commit with the workflow and readme change (10303284205971cf67341b88296f0a171a25a7f9) be made the only commit in this PR, or split of into its own PR? I would love to get that in but the commit with the reformatting itself is quite a lot to look over, and is currently having several merge conflicts. It would be much easier if I could format the codebase on my end and push it at an opportune time knowing that no other code was changed. Thank you for putting the work into this!

yedpodtrzitko:
yeah that makes sense. Done.

I also added (optional) pre-commit hook to format the code automatically when making new commit.

CyanVoxel (approved):
Awesome, thanks for doing this!

PR #109, opened 2024-05-01
fix file opening on posix

yedpodtrzitko:
`os.startfile()` is implemented only on Windows. This add support on posix systems.


xarvex:
It looks like this stemmed from #84 being refactored out of usage in #80. I went ahead and reapplied the fix that was present there into what you had.
Thank you very much for spotting this, so we could address it!

yedpodtrzitko:
I havent seen that PR, otherwise I'd probably wait until it will get merged or suggested it there via comment.

xarvex:
It was already merged, it just got lost somewhere in the recent refactor and the old code got put back in, that's what happened.

PR #110, opened 2024-05-01
fix image thumbnails rotation

yedpodtrzitko:
this fixes images rotated via EXIF orientation tag, see [ImageOps.exif_transpose](https://pillow.readthedocs.io/en/latest/reference/ImageOps.html#PIL.ImageOps.exif_transpose):

> If an image has an EXIF Orientation tag, other than 1, transpose the image accordingly, and remove the orientation data.


CyanVoxel:
Thanks for the fix! I went ahead and added the same line to the method that handles the smaller thumbnails as well, since it was a quick addition and fixes it for those thumbnails as well.

PR #111, opened 2024-05-01
feat: add hand cursor for active sidebar elements

yedpodtrzitko:
Add hand cursor (👆) for thumbnail and file path in sidebar to indicate these elements are interactive.


Loran425:
This one looks good to me, I like the hand context clue that something is clickable.

PR #113, opened 2024-05-01
minor bug fixes to Tag Search Box

TechCrafterGaming:
There was a bug if you added a tag and then directly afterwards (without selecting another element) wanted to add a tag to a field, the tag was not displayed at first, only when the `update_tag` function was called once.

and I gave the `update_tag` function an empty query as default so you don`t have to pass an empty string when you call it.


Loran425:
Nice tweak for allowing the tags to update in real time rather than having to close the AddTag modal. 
I like it.

PR #114, opened 2024-05-01
Remove functionally duplicated code in update_tags

Loran425:
Remove Commented lines
Modify update_tags to always use tag ids (this might be considered for changes later)
Modify update_tags to avoid possibility of different behavior based on if a query was entered or not. 
Remove unused `panel: BuildTagPanel`


CyanVoxel (approved):
Seems to work just fine! Thanks for the cleanup and additional comments!

PR #117, opened 2024-05-02
Improved TagStudio.sh

abby-freakazoid:
"set -e" to abort on error
"cd …" to prevent creating .venv if script is launched from outside git dir
"[ … ]" to skip creating venv when it already exists


xarvex:
Can confirm working on Linux, working across different directories with and without `.venv`. This can be merged now, thanks!

PR #121, opened 2024-05-02
Bugfix/right click file path/#116

### LINKED ISSUE #116: UI/UX Bug, Right Clicking the File Path
## Expected Behavior
Right Clicking the file path under the preview shows a context menu

## Current Behavior
When right clicking the file path a context menu is briefly shown before the file manger opens

## Possible Solution
Check how the `file_label` being clicked on opens explorer and only trigger opening the file in the file manager when the left mouse button is clicked, on right click open the existing context menu.

## Steps to Reproduce
1. Open a library
2. Select an item
3. Right Click the file path

image

gabrieljreed:
Add button handling to `FileOpenerLabel.mousePressEvent` to only call `open_explorer` when left button is clicked.

Resolves #116


CyanVoxel (approved):
Looks good! Thank you for the extra doc comments as well! While we don't have a current [docstring](https://www.geeksforgeeks.org/python-docstrings/) format at the moment, this will still be useful!

PR #123, opened 2024-05-02
Fix opening to a file with spaces in path on windows / #120

### LINKED ISSUE #120: [Windows Bug] Open file in explorer fails if space is in the filepath
## Expected Behavior
`Open file in explorer` should open a the file explorer at the file path

## Current Behavior
If the file path contains spaces explorer opens to another window (Documents on my machine)

## Possible Solution
Check filepath for whitespace when utilizing `open file in explorer` 
Only on windows, wrap the path with quotes to open the correct folder

## Steps to Reproduce
1. Create a library in a folder with whitespace in the path
2. Right Click on a file
3. Select `Open file in explorer`
4. Check what folder it opened

SylviaSK:
Unsure if this is exactly to style, but it should resolve #120 

Simply wrapping the path in quotes did not suffice, there seems to be some weird edge case in `.Popen()` with args containing quotes

### DIFF
--- a/tagstudio/src/qt/helpers/file_opener.py
+++ b/tagstudio/src/qt/helpers/file_opener.py
@@ -35,13 +35,15 @@ def open_file(path: str, file_manager: bool = False):
 		if sys.platform == "win32":
 			normpath = os.path.normpath(path)
 			if file_manager:
-				command_name = "explorer"
-				command_args = [f"/select,{normpath}"]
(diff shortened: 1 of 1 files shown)


CyanVoxel (approved):
Nice catch! This seems to be the issue that keeps coming back... Not sure why exactly it doesn't like the list, but thank you for finding a fix!

PR #126, opened 2024-05-03
Refactor strip_web_protocol()

JinguBangWest:
This allows for more prefixes to be added in the future if needed without repeating code multiple times.

### DIFF
--- a/tagstudio/src/core/utils/web.py
+++ b/tagstudio/src/core/utils/web.py
@@ -4,9 +4,8 @@
 
 def strip_web_protocol(string: str) -> str:
     """Strips a leading web protocol (ex. \"https://\") as well as \"www.\" from a string."""
-    new_str = string
-    new_str = new_str.removeprefix('https://')
-    new_str = new_str.removeprefix('http://')
-    new_str = new_str.removeprefix('www.')
-    new_str = new_str.removeprefix('www2.')
-    return new_str
+    prefixes = ['https://','http://','www.','www2.']
+    for prefix in prefixes:
+        if string.startswith(prefix):
+            string = string.replace(prefix, '')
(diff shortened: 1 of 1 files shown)


tagstudio/src/core/utils/web.py
```diff
@@ -4,9 +4,8 @@
...
-    new_str = string
-    new_str = new_str.removeprefix('https://')
-    new_str = new_str.removeprefix('http://')
-    new_str = new_str.removeprefix('www.')
-    new_str = new_str.removeprefix('www2.')
-    return new_str
+    prefixes = ['https://','http://','www.','www2.']
+    for prefix in prefixes:
+        if string.startswith(prefix):
+            string = string.replace(prefix, '')
```
yedpodtrzitko, line 10: consider following example:

    >>> 'www.foowww.com'.replace('www.', '')
    'foocom'

Al3bad: This should handle it. Basically replacing only the first occurance of the prefix

```
>>> 'www.foowww.com'.replace('www.', '', 1)
'foowww.com'
```

yedpodtrzitko: ...why not keep it simple?

```diff
-       if string.startswith(prefix):
-            string = string.replace(prefix, '')
+       string = string.removeprefix(prefix)
```

JinguBangWest: I meant to leave it as `string.removeprefix()` but didn't change it back for my commit. Also @yedpodtrzitko your right we should keep it simple. I've went ahead and changed this in my latest commit.

Loran425:
Functionally identical, looks good.

PR #127, opened 2024-05-03
use list of libraries as a starting point

yedpodtrzitko:
This changes the behaviour when opening the application - instead of opening the last library (`settings.last_library`), it shows a list of all previously opened libraries:
Screenshot 2024-05-03 at 12 44 09

plus the setting to make it optional (default false)

Screenshot 2024-05-09 at 23 21 46

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -8,6 +8,7 @@
 """A Qt driver for TagStudio."""
 
 import ctypes
+import enum
 import logging
 import math
 import os
@@ -15,13 +16,14 @@ import sys
 import time
 import webbrowser
 from datetime import datetime as dt
+from hashlib import sha1
 from pathlib import Path
 from queue import Empty, Queue
 from typing import Optional
 
 from PIL import Image
 from PySide6 import QtCore
-from PySide6.QtCore import QObject, QThread, Signal, Qt, QThreadPool, QTimer, QSettings
+from PySide6.QtCore import QObject, QThread, Signal, Qt, QThreadPool, QSettings
 from PySide6.QtGui import (
     QGuiApplication,
     QPixmap,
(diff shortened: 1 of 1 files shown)


Loran425:
I like the conceptual idea of recent libraries, but this is a fairly large departure from the current user experience. 
A sub menu from an open recent would be what I'd expect from software like this (at least from most of the windows programs I've seen), but we're in alpha so worth discussing.

yedpodtrzitko:
>  but this is a fairly large departure from the current user experience.

the current state is that when I open the app, it starts dozen of threads crunching thumbnails. The chance I really want to do it is (1 / number_of_libraries). With this change I can avoid doing that until I explicitly select the library I  want to work with. Menu with list of recent libraries doesnt solve this issue. 
If we're talking UX, the UX of given menu is generally horrible. Selecting library is something I do very often, so having such action available within a single click cant really be compared with a tiny nested menu a few clicks away.

Loran425:
> the current state is that when I open the app, it starts dozen of threads crunching thumbnails. The chance I really want to do it is (1 / number_of_libraries). With this change I can avoid doing that until I explicitly select the library I want to work with. Menu with list of recent libraries doesnt solve this issue.

Ahhh I see the difference in what you're looking for, looking to avoid the threading spool up just for a library that might be closed shortly after starting the program. 

I know the open last library is intended to be a temporary option where the user would be allowed to have a preference of where they could select the starting behavior similar to how browsers do it. 
- Open last library
- Open default library
- Start Empty

Could add a preference for a library picker, I think one of the design decisions that's still up in the air is allowing multiple directories to feed a single `library`

> If we're talking UX, the UX of given menu is generally horrible. Selecting library is something I do very often, so having such action available within a single click cant really be compared with a tiny nested menu a few clicks away.

Not sure I'd call it horrible, if someone is using a single library this is directly adding a step between starting the program and using the program. That's why I'm thinking it should be an option but not the only option, because I know I'd be annoyed by another click every time I start the program to test something or use it in my singular library, but I can see that opening the last library only to immediately open another library is causing frustration with the program on your end.

yedpodtrzitko:
Thanks for feedback. Choice what should happen when app is opened sounds reasonable.

PR #128, opened 2024-05-03
Update README.md

javacaliente:
Added a ## Tech-Stack section.


Loran425:
Not sure how much of a tech stack is present for a client application, only tech utilized for this TagStudio, really SQLite/Local Database (future release), Python & Qt via PySide, but I'd be interested to see if there's an approach I'm missing.

CyanVoxel:
Closing this since there isn't much in the project warranting a section like this, plus the section has remained blank.

PR #135, opened 2024-05-04
misc: remove duplicate code for tags updating

yedpodtrzitko:
- code for toggling archived and favorite tag is kinda' duplicate, plus each of these methods has some redundant code within as well.

- use identifiers for the archived/favorite tag instead of non-descriptive int values.


Loran425:
looks good

CyanVoxel (approved):
Looks great, thank you!

PR #136, opened 2024-05-04
Fix for #119, Prompt for new location user if library cannot be saved

### LINKED ISSUE #119: If tag studio cannot find the library to save the library it cannot close the library
## Expected Behavior
If TagStudio cannot find the save directory for the library it should prompt the user to close the library or save in a new location

## Current Behavior
Saving/Closing errors and the library remains open

## Possible Solution
During save or close library, check that the save location is valid, if not, prompt the user to close the library or select a new location to save the library (should be the root of wherever the library folder moved)

## Steps to Reproduce (Edit: no need to close/reopen TagStudio)
1. Open a library
2. Delete or rename the library folder from the file manager
3. Attempt to use `File > Save Library` or `File > Close Library`

```
[LIBRARY] Saving Library to Disk...
  File "C:\Users\Loran425\PycharmProjects\TagStudio\tagstudio\src\qt\ts_qt.py", line 254, in <lambda>
    close_library_action.triggered.connect(lambda: self.close_library())
                                                   ^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Loran425\PycharmProjects\TagStudio\tagstudio\src\qt\ts_qt.py", line 439, in close_library
    self.lib.save_library_to_disk()
  File "C:\Users\Loran425\PycharmProjects\TagStudio\tagstudio\src\core\library.py", line 870, in save_library_to_disk
    self.verify_ts_folders()
  File "C:\Users\Loran425\PycharmProjects\TagStudio\tagstudio\src\core\library.py", line 575, in verify_ts_folders
    os.mkdir(full_ts_path)
FileNotFoundError: [WinError 3] The system cannot find the path specified: 'C:\\Users\\Loran425\\Pictures\\Item Icons\\tradingicons\\Testing\\.TagStudio'
```

SylviaSK:
This should resolve #119 

Simply prompts the user to select a new directory if `self.lib.save_library_to_disk()` throws a `FileNotFoundError`

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -551,12 +551,13 @@ class QtDriver(QObject):
             thread.wait()
         QApplication.quit()
 
-    def save_library(self):
+    def save_library(self, show_status=True):
         logging.info(f"Saving Library...")
-        self.main_window.statusbar.showMessage(f"Saving Library...")
-        start_time = time.time()
-        # This might still be able to error, if the selected directory deletes in a race condition or something silly like that
-        # Hence the loop, but if this is considered overkill, thats fair.
+        if show_status:
+            self.main_window.statusbar.showMessage(f"Saving Library...")
+            start_time = time.time()
+        # This might still be able to error, if the selected directory deletes in a race condition
+        # or something silly like that. Hence the loop, but if this is considered overkill, thats fair.
         while True:
             try:
                 self.lib.save_library_to_disk()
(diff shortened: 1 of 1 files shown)


SylviaSK:
The only section I changed is passing Ruff on my end, and I pulled the Ruff commit before making my edit, so its not reverting to pre-ruff. If anyone knows what could be causing this I'd appreciate any insight.

CyanVoxel:
> The only section I changed is passing Ruff on my end, and I pulled the Ruff commit before making my edit, so its not reverting to pre-ruff. If anyone knows what could be causing this I'd appreciate any insight.

When I try copying the raw file and formatting it with Ruff on my local machine, it seems to want to try formatting a couple method calls to try and take up less width:
```py
logging.info(
    "Library parent directory not found, prompting user to select the directory"
)
dir = QFileDialog.getExistingDirectory(
    None,
    "Library Location not found, please select location to save Library",
    "/",
    QFileDialog.ShowDirsOnly,
)
```

Ruff should be using the default configuration. If yours is also using the default, then I'm not sure why it also wouldn't try to format it in the same way. Sorry for the inconvenience by the way, the ruff workflow is new and is sure to have some growing pains.

SylviaSK:
My mistake, I installed it as a VS Code extension and took its linting as what it wanted. Seeing no issues, I put in the pull request. I now installed it for CLI, and the formater puts out the same as what you posted.

That said, I didn't realize until now that `close_library()` has the same bug separate from `save_library()`, so I'll work on getting that fixed, and update this PR with one that fixes the bug in both, and fits to Ruff's formatting. If I should instead set up a different PR, or something like that, let me know, I'm somewhat new to writing PRs.

CyanVoxel:
No worries! You can keep it in this PR and I'll only pull it when you're ready. You could also turn this PR into a draft which won't let me pull until you undraft it, but in this case I'll remember not to pull is early :) And thank you for your contribution!

SylviaSK:
Thanks for your help! It should be ready now.

CyanVoxel (approved):
Looks good! Good catch with `close_library()` as well, it's better to have all that saving logic in the same method anyway!

PR #142, opened 2024-05-06
Add pytest support

Hidorikun:
This is a rewrite of https://github.com/TagStudioDev/TagStudio/pull/88 because it was faster and less error prone to do so than to solve the merge conflicts.

I would like to keep this smaller so merge conflicts don't affect it as much so I omitted the method refactoring from the first pull request with the intention of adding it in a separate pull request.


gabrieljreed (approved):

gabrieljreed:
This looks great!

Hidorikun:
@CyanVoxel @Loran425  @gabrieljreed Kind reminder that this is ready to be merged whenever you see fit 😄

tagstudio/src/core/ts_core.py
```diff
@@ -9,143 +9,6 @@
 
 from src.core.library import Entry, Library
```
CyanVoxel, line 11: The MyPy check is failing due to missing imports in the `ts_core.py` file. Seems like they were accidentally missed when moving the constants out into their own file.
```suggestion
from src.core.constants import TS_FOLDER_NAME, TEXT_FIELDS

```

Issue #86, opened 2024-04-28
[Bugfix Request] Prevent crash alert on exit on macOS


Xpl0itU:
Currently, when quitting on macOS, a popup will appear that the application has crashed, this might be due to threads not being terminated gracefully, from reading the QThread documentation, one solution might be this https://doc.qt.io/qt-6/qthread.html#isInterruptionRequested

PR #145, opened 2024-05-08
dont run job threads needlessly

yedpodtrzitko:
Currently the worker threads are running in a loop and raise `Empty` error each time when `queue.get()` in there timeouts (ie. every 0.2 seconds) and repeat the same thing again, which is not very clean.

This PR makes call of `queue.get()` blocking, so there's no need for the exception-raising mumbo jumbo. Previously there was problem with this approach on application shutdown. So now when the app is quitting, there's a marker put into the queue to notify the thread it should terminate itself.

Tested on macOS only.

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -549,6 +549,11 @@ class QtDriver(QObject):
         for _ in self.thumb_threads:
             self.thumb_job_queue.put(Consumer.MARKER_QUIT)
 
+        # wait for threads to quit
+        for thread in self.thumb_threads:
+            thread.quit()
+            thread.wait()
+
         QApplication.quit()
 
     def save_library(self, show_status=True):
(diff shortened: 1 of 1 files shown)


CyanVoxel:
Currently I'm able to consistently get a `QThread: Destroyed while thread is still running` warning when loading my video library and closing the program before the thumbnails have finished loading, which regresses the fix for #86 applied in #94. I'm assuming the threads are being terminated before reaching the markers in the queues?

Here's the full order of the console log:
```
Saving Library...
[LIBRARY] Saving Library to Disk...
[LIBRARY] Formatting Tags to JSON...
[LIBRARY] Formatting Entries to JSON...
[LIBRARY] Formatting Collations to JSON...
[LIBRARY] Done Formatting to JSON!
[LIBRARY] Library saved to disk in 5.633 seconds
[SHUTDOWN] Ending Thumbnail Threads...
QThread: Destroyed while thread is still running
```

yedpodtrzitko:
@CyanVoxel should be fixed now

PR #147, opened 2024-05-08
Add action to help menu

arthniwa:
Add "Go to GitHub Repository" to the help menu.


CyanVoxel (changes requested):
Looks great! Until there's a proper help resource, this will work nicely as a stopgap. All it needs is to be formatted via Ruff, with instructions on how to do that [over here](https://github.com/TagStudioDev/TagStudio?tab=readme-ov-file#can-i-contribute) on the readme.

However, for this PR, since all Ruff wants to do is format this one line to take up less width, you could get away with just manually formatting it as follows and it should pass the Ruff check:
Line 421:
```
self.repo_action.triggered.connect(
    lambda: webbrowser.open("https://github.com/TagStudioDev/TagStudio")
)
```

PR #149, opened 2024-05-09
Video Player

DrRetro2033:
Did a redo on the video player, as it was in GitHub conflict hell. So it should work now, plus I even added some rounded corners to it as well.


CyanVoxel (changes requested):
Love the functionality! And sorry for all the code changes in the meantime, there was a recent formatting refactor that probably messed with stuff pretty bad. Thank you for your patience!

Overall I think this is going in a good direction, there's just a few things I've run into or would like to see done:
- When clicking on a new video file, the preview size is always incorrect until I manually grab the resize bar to "refresh" the panel (pictured below)
image
- Rarely I'm encountering an issue with vertical videos where the only about the middle part of the video is updating during playback, while the rest of the frame area only updates when the panel size changes. This may be related to the first issue.
- The video playback can be kind of intensive sometimes and slows down the program. This is more so an issue stemming from having to stream a potentially high bitrate video file from disk while the program runs. Because of this, I would really like to see an autoplay toggle to stop videos from automatically playing in the preview window. Maybe this could be included in the right-click context menu for the video preview.
- And lastly I'd like to see the mute/unmute status carry over from video to video instead of always defaulting to mute. This doesn't need to persist between sessions, but I think it makes sense for each next video to remember your audio preference.

Thank you again for working on this and getting a pretty cool feature going!

DrRetro2033 (commented):
Fixed bugs and added autoplay toggle.

CyanVoxel (changes requested):
Awesome!! The persistent volume settings seem to work fine, but the autoplay toggle seems to be inconsistent. If I toggle autoplay off followed by clicking on another entry and then clicking back to the first video, then it will autoplay regardless of the setting.

Other than that, just a couple general housekeeping things:
- Remove or comment out the testing log messages (under mouse, over mouse, etc.). Ideally the ffmpeg logs would also be removed, but *I'm* not even sure if it's possible to turn those off, so I wouldn't worry about those.
- On startup I get the following warning logs:
```
QGraphicsScene::addItem: item has already been added to this scene
QGraphicsProxyWidget::setWidget: cannot embed widget 0x27f44083a30 which is not a toplevel widget, and is not a child of an embedded widget
QGraphicsProxyWidget::setWidget: cannot embed widget 0x27f44083d90 which is not a toplevel widget, and is not a child of an embedded widget
```
The first warning seems to stem from line 85 in `video_player.py`, however I'm not sure about the other two warnings. I would suggest investigating the causes of these.

tagstudio/src/qt/widgets/video_player.py
```diff
@@ -0,0 +1,400 @@
...
+            # self.video_preview.setSize(self.resolution)
+            self.player.setPosition(0)
+            # logging.info(f'Set muted to true.')
+            extension = os.path.splitext(self.filepath)[1][1:].lower()
+            filename = os.path.basename(self.filepath)
+            if old_filename == filename:
+                self.player.play()
+            elif extension in VIDEO_TYPES:
+                # self.player.setVideoOutput(self.video_preview)
+                self.album_art.setOpacity(0.0)
```
CyanVoxel, line 189: This line throws an error as of ec0d4cb7089a8d225ba2fe3573602bc240b68d77 since the `album_art` variable was removed

CyanVoxel:
> Awesome!! The persistent volume settings seem to work fine, but the autoplay toggle seems to be inconsistent. If I toggle autoplay off followed by clicking on another entry and then clicking back to the first video, then it will autoplay regardless of the setting.

I'm also still having this same autoplay issue where it has inconsistency following the setting.

DrRetro2033:
> > Awesome!! The persistent volume settings seem to work fine, but the autoplay toggle seems to be inconsistent. If I toggle autoplay off followed by clicking on another entry and then clicking back to the first video, then it will autoplay regardless of the setting.
> 
> I'm also still having this same autoplay issue where it has inconsistency following the setting.

I can't seem to recreate the autoplay issue. It is working just fine on my end after multiple tries.

CyanVoxel:
> I can't seem to recreate the autoplay issue. It is working just fine on my end after multiple tries.

Here's a clip of me recreating the issue on my end:

https://github.com/TagStudioDev/TagStudio/assets/46939827/2e534205-969c-4331-8200-ac1603e10192

It usually takes the form of the video autoplaying regardless of the setting when clicking off then back to the file. It seems to also be affected by whether or not the last file clicked on was also a video.

DrRetro2033:
> > I can't seem to recreate the autoplay issue. It is working just fine on my end after multiple tries.
> 
> Here's a clip of me recreating the issue on my end:
> 2024-05-10.12-57-17-1.mp4
> 
> It usually takes the form of the video autoplaying regardless of the setting when clicking off then back to the file. It seems to also be affected by whether or not the last file clicked on was also a video.

Alright, I see why I did not get the bug, I was switching between video to video and not video to image. It was because of code that was checking to see if the file is the same as the last video it played (for old logic I forgot to remove.) So, I had the check for autoplay, but the logic was skipping over it. Now it works just fine. ;)

tagstudio/src/qt/widgets/video_player.py
```diff
@@ -0,0 +1,382 @@
...
+            + (self.contentsRect().size().width(), self.contentsRect().size().height()),
+            radius=20,
+            fill=(0, 0, 0, 0),
+        )
+        mask.resize(
+            (
+                int(self.contentsRect().size().width()),
+                int(self.contentsRect().size().height()),
+            )
+        )
```
CyanVoxel, line 345: Recommend ensuring that the width and height of the video are always at least 0 to prevent a `ValueError: Width and height must be >= 0` error which prevents any new video from loading (I actually ran into this). The changes to this method also include removing the unnecessary resize step as well as decreasing the border radius to better match the image thumbnails.
```suggestion
        width: int = int(max(self.contentsRect().size().width(), 0))
        height: int = int(max(self.contentsRect().size().height(), 0))
        mask = Image.new(
            "RGBA",
            (
                width,
                height,
            ),
            (0, 0, 0, 255),
        )
        draw = ImageDraw.Draw(mask)
        draw.rounded_rectangle(
            (0, 0) + (width, height),
            radius=12,
            fill=(0, 0, 0, 0),
        )
```

tagstudio/src/qt/widgets/video_player.py
```diff
@@ -0,0 +1,382 @@
...
+
+    def close(self, *args, **kwargs) -> None:
+        self.player.stop()
+        super().close(*args, **kwargs)
+
+    def toggleAutoplay(self) -> None:
+        self.driver.settings.setValue("autoplay_videos", self.autoplay.isChecked())
+
+    def checkMediaStatus(self, media_status: QMediaPlayer.MediaStatus) -> None:
+        logging.info(media_status)
```
CyanVoxel, line 166: Recommend commenting out this log
```suggestion
        # logging.info(media_status)
```

tagstudio/src/qt/widgets/video_player.py
```diff
@@ -0,0 +1,382 @@
...
+            # Even if I stop the player before switching, it breaks.
+            # On the plus side, this adds infinite looping for the video preview.
+            self.player.stop()
+            self.player.setSource(QUrl().fromLocalFile(self.filepath))
+            # logging.info(f'Set source to {self.filepath}.')
+            # self.video_preview.setSize(self.resolution)
+            self.player.setPosition(0)
+            # logging.info(f'Set muted to true.')
+            extension = os.path.splitext(self.filepath)[1][1:].lower()
+            filename = os.path.basename(self.filepath)
```
CyanVoxel, line 178: Recommend removing unused code here
```suggestion
```

CyanVoxel:
Looking good! Besides the suggested changes above, there's only a couple issues I'm still running into. One is that the play/pause button seems to get desynced and doesn't update to the correct "pause" state when the videos start autoplaying.

The other issue might be tougher to track down, and seems to only affect certain video files I've tested it on. This is the issue I mentioned where only the center part of the video frame updates, unless there's mouse activity over the video where that'll force a frame update. So far I've only found 3 video files that do this, and they're all .mov files.

https://github.com/TagStudioDev/TagStudio/assets/46939827/154c9210-ca06-4d84-a197-d60aa81a0d63

tagstudio/src/qt/widgets/video_player.py
```diff
@@ -0,0 +1,382 @@
...
+            if self.player.hasVideo():
+                self.pauseToggle()
+
+        if (
+            obj == self.mute_button
+            and event.type() == QEvent.Type.MouseButtonPress
+            and event.button() == Qt.MouseButton.LeftButton
+        ):
+            if self.player.hasAudio():
+                self.muteToggle()
```
yedpodtrzitko, line 217:
```suggestion
        if event.type() == QEvent.Type.MouseButtonPress and event.button() == Qt.MouseButton.LeftButton:
            if obj == self.play_pause and self.player.hasVideo():
                self.pauseToggle()
            if obj == self.mute_button and self.player.hasAudio():
                self.muteToggle()
```

DrRetro2033:
> Looking good! Besides the suggested changes above, there's only a couple issues I'm still running into. One is that the play/pause button seems to get desynced and doesn't update to the correct "pause" state when the videos start autoplaying.
> 
> The other issue might be tougher to track down, and seems to only affect certain video files I've tested it on. This is the issue I mentioned where only the center part of the video frame updates, unless there's mouse activity over the video where that'll force a frame update. So far I've only found 3 video files that do this, and they're all .mov files.
> 2024-05-10.17-58-05-1.mp4

Alright, So I do not have any vertical videos in the mov format to test; BUT, I have a hunch it has something to do with how Qt is rendering the video. I noticed in the video you provided, the screen was glitching the most with videos that were higher than 1440p. So I tried playing the GTA 6 trailer at 2160p in the video player and it had artifacts and freezes the program. I may have a solution, but I do not know how long it will take to implement it. Hopefully I will have the solution in a day or two. But I am not holding my breath.

CyanVoxel:
> Alright, So I do not have any vertical videos in the mov format to test; BUT, I have a hunch it has something to do with how Qt is rendering the video. I noticed in the video you provided, the screen was glitching the most with videos that were higher than 1440p. So I tried playing the GTA 6 trailer at 2160p in the video player and it had artifacts and freezes the program. I may have a solution, but I do not know how long it will take to implement it. Hopefully I will have the solution in a day or two. But I am not holding my breath.

Here are the videos I've found that result in the issue for me. One of them isn't even vertical and is pretty low resolution (it's also a bizarre video so I'm sorry in advance), so that may (or may not) narrow down the issue...
[Videos for PR 149.zip](https://github.com/TagStudioDev/TagStudio/files/15289495/Videos.for.PR.149.zip)

DrRetro2033:
> > Alright, So I do not have any vertical videos in the mov format to test; BUT, I have a hunch it has something to do with how Qt is rendering the video. I noticed in the video you provided, the screen was glitching the most with videos that were higher than 1440p. So I tried playing the GTA 6 trailer at 2160p in the video player and it had artifacts and freezes the program. I may have a solution, but I do not know how long it will take to implement it. Hopefully I will have the solution in a day or two. But I am not holding my breath.
> 
> Here are the videos I've found that result in the issue for me. One of them isn't even vertical and is pretty low resolution (it's also a bizarre video so I'm sorry in advance), so that may (or may not) narrow down the issue... [Videos for PR 149.zip](https://github.com/TagStudioDev/TagStudio/files/15289495/Videos.for.PR.149.zip)

Thanks for the Videos, they are extremely helpful. It appears that I am getting a slightly different, more stranger bug now.
image
It appears that the video is rotated 90 degrees, and it is for every video you sent. Extremely weird. I am starting to think PySide6 hates me.

CyanVoxel:
> Thanks for the videos, they are extremely helpful. It appears that I am getting a slightly different, more stranger bug now. ! It appears that the video is rotated 90 degrees, and it is for every video you sent. Extremely weird. I am starting to think PySide6 hates me.

Seeing how they're playing for you has potentially led me to the solution: These videos are actually encoded as having their height and widths swapped, with an EXIF rotation flag being set to correct for them. This is something I already have to manually account for when loading in image previews. I'll dig around the video player code to see where I can account for this flag and correct the resolution.
*Actual dimensions for the low-res landscape video:*
image
image

CyanVoxel:
Upon further testing this issue may lay deeper within Qt or possibly FFmpeg. That fact that we already had different behaviors should've tipped me off. While I'm certain that the rotation flag is the root of this behavior (also it wasn't EXIF but rather just QuickTime metadata), I'm not sure if it's easily fixable on our end.

yedpodtrzitko:
it will be slightly PITA with mypy & QT, please add `# type: ignore` on the lines where it reported error and but dont see any obvious error there, thanks.

yedpodtrzitko:
I did checkout the branch to see why is it failing and I noticed one unrelated bug:

- select any video in library
- video starts playing in the thumbnail
- hover over the video and it shows play icon (▶️ ), but the video is playing already so it should show pause icon instead.

yedpodtrzitko:
to fix the `PySide App Test` CI job, please add the following change into your PR: https://github.com/TagStudioDev/TagStudio/commit/57f64e97cf34c5d81398d69f8a5aa459cf7d6861

yedpodtrzitko:
here is the example of `# type: ignore` usage:

https://github.com/TagStudioDev/TagStudio/blob/1461f2ee70be4437dfc3b49f86f42bef80899574/tagstudio/src/qt/widgets/tag_box.py#L81

yedpodtrzitko:
you can check this commit how to fix the mypy issues: https://github.com/TagStudioDev/TagStudio/commit/8e62147e24b4685271e69ea27bb1a79175651f89

tagstudio/src/qt/widgets/video_player.py
```diff
@@ -0,0 +1,377 @@
...
+            ),
+            (0, 0, 0, 255),
+        )
+        draw = ImageDraw.Draw(mask)
+        draw.rounded_rectangle(
+            (0, 0) + (width, height),
+            radius=12,
+            fill=(0, 0, 0, 0),
+        )
+        mask = mask.getchannel("A").toqpixmap()
```
yedpodtrzitko, line 336: `mask` has originally type `PIL.Image`, but here it's changed to `QPixmap`, so it would be really better to use different variable for one of these, so the types wont clash

tagstudio/src/qt/widgets/video_player.py
```diff
@@ -0,0 +1,377 @@
...
+            QSize(
+                int(self.video_preview.size().width()),
+                int(self.video_preview.size().height()),
+            )
+        )
+        return
+        # return super().resizeEvent(event)\
+
+    def inputMethodEvent(self, event: QInputMethodEvent) -> None:
+        return
```
yedpodtrzitko, line 364: is this intentional?

CyanVoxel:
The play/pause indicator not updating after autoplaying seems to be *mostly* fixed, except in the case of when the video autoplays while the user's mouse is hovered over the indicator.

Shirkit:
This seems like a nice feature to add to this project, looking forward to it!

CyanVoxel (approved):

CyanVoxel:
@DrRetro2033 Thank you again for all your work on this!

PR #151, opened 2024-05-09
add list of libraries into sidebar

yedpodtrzitko:
superseding #127

- show list of previously opened libraries on bottom of sidebar:
Screenshot 2024-05-10 at 1 57 38

- add option (default True) to decide if previous library should be opened on app start
Screenshot 2024-05-10 at 2 03 26

- there's a minor refactoring to keep working with variables localized. So instead of this:
```
root_layout = QHBoxLayout(self)
# -50 various lines of code-
root_layout.setContentsMargins(0, 0, 0, 0)
# -more unrelated code-
root_layout.addWidget(splitter)
```

...working with the variable is now moved into the same place.

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -505,7 +505,11 @@ class QtDriver(QObject):
             )
             self.open_library(lib)
 
-        app.exec_()
+        if self.args.ci:
+            # gracefully terminate the app in CI environment
+            self.thumb_job_queue.put((self.SIGTERM.emit, []))
+
+        app.exec()
 
         self.shutdown()
 
@@ -1329,6 +1333,12 @@ class QtDriver(QObject):
             # logging.info(f'Done Filtering! ({(end_time - start_time):.3f}) seconds')
 
             # self.update_thumbs()
+
+    def remove_recent_library(self, item_key: str):
+        self.settings.beginGroup(SettingItems.LIBS_LIST)
+        self.settings.remove(item_key)
+        self.settings.endGroup()
+        self.settings.sync()
 
     def update_libs_list(self, path: str | Path):
         """add library to list in SettingItems.LIBS_LIST"""
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -1330,29 +1330,27 @@ class QtDriver(QObject):
 
             # self.update_thumbs()
 
-    def update_libs_list(self, path: str):
-        # add library to list in SettingItems.LIBS_LIST
+    def update_libs_list(self, path: str | Path):
+        """add library to list in SettingItems.LIBS_LIST"""
         ITEMS_LIMIT = 5
+        path = Path(path)
 
         self.settings.beginGroup(SettingItems.LIBS_LIST)
 
-        current_time = str(time.time())
-        all_libs = {current_time: path}
+        all_libs = {str(time.time()): str(path)}
 
         for item_key in self.settings.allKeys():
             item_path = self.settings.value(item_key)
-            if item_path != path:
+            if Path(item_path) != path:
                 all_libs[item_key] = item_path
 
-        # sort items by the most recent first
+        # sort items, most recent first
         all_libs = sorted(all_libs.items(), key=lambda item: item[0], reverse=True)
 
         # remove previously saved items
         self.settings.clear()
 
-        for idx, (item_key, item_value) in enumerate(all_libs, start=1):
-            if idx > ITEMS_LIMIT:
-                break
+        for item_key, item_value in all_libs[:ITEMS_LIMIT]:
             self.settings.setValue(item_key, item_value)
 
         self.settings.endGroup()
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -16,7 +16,7 @@ import time
 import webbrowser
 from datetime import datetime as dt
 from pathlib import Path
-from queue import Empty, Queue
+from queue import Queue
 from typing import Optional
 
 from PIL import Image
@@ -46,6 +46,7 @@ from PySide6.QtWidgets import (
 )
 from humanfriendly import format_timespan
 
+from src.core.enums import SettingItems
 from src.core.library import ItemType
 from src.core.ts_core import (
     PLAINTEXT_TYPES,
@@ -88,7 +89,9 @@ from src.qt.modals.file_extension import FileExtensionModal
 from src.qt.modals.fix_unlinked import FixUnlinkedEntriesModal
 from src.qt.modals.fix_dupes import FixDupeFilesModal
 from src.qt.modals.folders_to_tags import FoldersToTagsModal
-import src.qt.resources_rc
+
+# this import has side-effect of import PySide resources
+import src.qt.resources_rc  # pylint: disable=unused-import
 
 # SIGQUIT is not defined on Windows
 if sys.platform == "win32":
@@ -282,6 +285,7 @@ class QtDriver(QObject):
         edit_menu = QMenu("&Edit", menu_bar)
         tools_menu = QMenu("&Tools", menu_bar)
         macros_menu = QMenu("&Macros", menu_bar)
+        window_menu = QMenu("&Window", menu_bar)
         help_menu = QMenu("&Help", menu_bar)
 
         # File Menu ============================================================
@@ -376,6 +380,18 @@ class QtDriver(QObject):
         tag_database_action.triggered.connect(lambda: self.show_tag_database())
         edit_menu.addAction(tag_database_action)
 
+        check_action = QAction("Open library on start", self)
+        check_action.setCheckable(True)
+        check_action.setChecked(
+            self.settings.value(SettingItems.START_LOAD_LAST, True, type=bool)
+        )
+        check_action.triggered.connect(
+            lambda checked: self.settings.setValue(
+                SettingItems.START_LOAD_LAST, checked
+            )
+        )
+        window_menu.addAction(check_action)
+
         # Tools Menu ===========================================================
         fix_unlinked_entries_action = QAction("Fix &Unlinked Entries", menu_bar)
         fue_modal = FixUnlinkedEntriesModal(self.lib, self)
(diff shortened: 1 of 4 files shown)


CyanVoxel:
I like the ability to toggle the automatic library loading on start. I saw in #127 you said you motivation for this was the thumbnail threads spooling up, which is more of an issue with the current lack of thumbnail caching than anything else. That being said, there will always be some resources being used when opening a library and having the choice to toggle that off is valuable regardless.

While I largely agree with @Loran425 about the approach of having an "Open Recent Libraries" menu over the idea of having a persistent panel on screen, I can see the desire to be able to switch between libraries faster. As long as it's able to be hidden for users who don't want it, I don't see the harm in it. With that being said though, I can't seem to get it to even appear on my end:
image

In addition, the splash screen seems to be broken for me now, which could be related to the changes in `ts_qt.py` or could just be some Qt resource shenanigans.
image

yedpodtrzitko:
> While I largely agree with @Loran425 about the approach of having an "Open Recent Libraries" menu over the idea of having a persistent panel on screen, I can see the desire to be able to switch between libraries faster. As long as it's able to be hidden for users who don't want it, I don't see the harm in it. 

that's not configurable via the following menu item:

Screenshot 2024-05-11 at 22 56 53




> With that being said though, I can't seem to get it to even appear on my end:

That should be fixed now. It got lost when checking in new commits. Thanks for catching that.

> In addition, the splash screen seems to be broken for me now, which could be related to the changes in `ts_qt.py` or could just be some Qt resource shenanigans.

I never see the splash screen (even before), so I cant test it, but I made a speculative change which could fix it.


Let me know if it works for you now as expected, thanks.

CyanVoxel:
Currently there's an issue on Windows where the path separators aren't normalized list the library list, resulting in duplicate libraries showing up in the list:
image
Normally I would recommend wrapping the paths in `os.path.normpath()`, however #156 is trying to remove the use of `os.path` in favor of the `pathlib` library.

Also, the splash screen is still broken for me but I'll see if I can make changes to get that functional on at least my end.

tagstudio/src/qt/widgets/preview_panel.py
```diff
@@ -209,6 +222,52 @@ def __init__(self, library: Library, driver: "QtDriver"):
...
+            button = QPushButton(text=lib_path)
+            button.setObjectName(lib_key)
+
+            def open_library_button_clicked(path):
+                return lambda: self.driver.open_library(path)
+
+            button.clicked.connect(open_library_button_clicked(lib_path))
+            button.setStyleSheet(
+                "background-color: #ffffff; color: #000000; text-align: left; padding-left: 10px;"
+            )
```
CyanVoxel, line 268: Proposed style changes to the library section to better match the look and feel of the rest of the program (screenshot includes changes to the section title not featured in the changes below)
image

```suggestion
            button.setStyleSheet(
                "QPushButton{"
                "background-color:#65000000;"
                "border-radius:6px;"
                "text-align: left;"
                "padding-top: 3px;"
                "padding-left: 6px;"
                "padding-bottom: 4px;"
                "}"
                "QPushButton::hover{background-color:#65AAAAAA;}"
                "QPushButton::pressed{background-color:#65EEEEEE;}"
            )
            button.setCursor(Qt.CursorShape.PointingHandCursor)
```

tagstudio/src/qt/widgets/preview_panel.py
```diff
@@ -209,6 +222,52 @@ def __init__(self, library: Library, driver: "QtDriver"):
...
+
+    def _fill_libs_widget(self, libraries: dict[str, str], layout: QVBoxLayout):
+        # remove any potential previous items
+        for idx in reversed(range(layout.count())):
+            widget = layout.itemAt(idx).widget()
+            layout.removeWidget(widget)
+            # remove from GUI
+            widget.setParent(None)
+
+        label = QLabel("Select a library")
```
CyanVoxel, line 254: Suggestion to rename this section to "Recent Libraries". I feel this is more descriptive of the section, and is more consistent with similar "File -> Open Recent Thing" functionality found in other programs, and likely this program in the future.
```suggestion
        label = QLabel("Recent Libraries")
        label.setStyleSheet("font-weight:bold;")
```

tagstudio/src/qt/ts_qt.py
```diff
@@ -416,6 +431,20 @@ def start(self):
         macros_menu.addAction(self.sort_fields_action)
 
         folders_to_tags_action = QAction("Create Tags From Folders", menu_bar)
+        show_libs_list_action = QAction("Show Libraries List", menu_bar)
```
CyanVoxel, line 434: Similar suggestion to the main "Recent Libraries" one, just changing the menu action to match.
```suggestion
        show_libs_list_action = QAction("Show Recent Libraries", menu_bar)
```

CyanVoxel:
I'll be honest, I was skeptical at first of whether I would personally use this list vs just having it tucked away in the File menu, but as soon as I actually used it I didn't want to live without it 😁 Fantastic work on this so far!

Other than the issues and styling changes commented above, I wanted to suggest a few other changes/additions. These tie into the idea of this being more of a "Recent Libraries" list rather than a list of all libraries you've historically opened.

### Proposed Changes/Additions
1. Automatic reordering of libraries based on the order of last opening. For example if you were to click on a library in the middle of the list to open it, then it would move to the top.
2. Right-click context menu option to remove a library from the list.
3. A limit for how many libraries show up in this list. I think a hardcoded limit of 5 would work well, then once we get a proper settings panel then this can be user-configurable.

Thank you again for all your work on this!

tagstudio/src/qt/ts_qt.py
```diff
@@ -90,7 +92,7 @@
     FixDupeFilesModal,
     FoldersToTagsModal,
 )
-import src.qt.resources_rc
```
CyanVoxel, line 93: Removing this import is what causes in the splash screen to break, and is currently the pinpoint of a merge conflict

yedpodtrzitko: Ah so it's not unused import, but it has side-effect of executing code. Not a good thing to do, but looks like there's no way around it. I'll add comment so it's obvious why is it here.

yedpodtrzitko:
> Currently there's an issue on Windows where the path separators aren't normalized list the library list, resulting in duplicate libraries showing up in the list: 

I'm afraid I dont have any windows machine around here, so I cant reproduce this issue. I just hope the Pathlib PR will solve this issue.

yedpodtrzitko:
> Right-click context menu option to remove a library from the list.

I was thinking to implement this in future, but rather than using the context menu, I was considering to add there another button, the same way as Tags have it (➖ ), or Fields have it (🗑️ ), so the design language is somehow consistent.

CyanVoxel:
> > Right-click context menu option to remove a library from the list.
> 
> I was thinking to implement this in future, but rather than using the context menu, I was considering to add there another button, the same way as Tags have it (➖ ), or Fields have it (🗑️ ), so the design language is somehow consistent.

Makes sense to me, I thing the (➖) would be the better route.

CyanVoxel:
Not sure if you're running into this issue as well, but when using the refactored color variable enums:
```py
scroll_area.setStyleSheet(
            "QWidget#entryScrollContainer{"
            f"background: {Theme.COLOR_BG};"
            "border-radius:6px;"
            "}"
        )
```
Qt throws an error trying to read from the stylesheets because it's using the literal name "Theme.COLOR_BG" as the color value. If I change these instances to `Theme.COLOR_BG.value` then it works as intended.

yedpodtrzitko:
> Not sure if you're running into this issue as well, but when using the refactored color variable enums:
> 
> ```python
> scroll_area.setStyleSheet(
>             "QWidget#entryScrollContainer{"
>             f"background: {Theme.COLOR_BG};"
>             "border-radius:6px;"
>             "}"
>         )
> ```
> 
> Qt throws an error trying to read from the stylesheets because it's using the literal name "Theme.COLOR_BG" as the color value. If I change these instances to `Theme.COLOR_BG.value` then it works as intended.

No issue on my side, but adding the `.value` explicitly wont hurt.

tagstudio/src/qt/ts_qt.py
```diff
@@ -1293,6 +1329,34 @@ def filter_items(self, query=""):
...
+
+        self.settings.beginGroup(SettingItems.LIBS_LIST)
+
+        current_time = str(time.time())
+        all_libs = {current_time: path}
+
+        for item_key in self.settings.allKeys():
+            item_path = self.settings.value(item_key)
+            if item_path != path:
+                all_libs[item_key] = item_path
```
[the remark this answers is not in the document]
yedpodtrzitko: I achieved this with fewer casting, hopefully it will still work as expected.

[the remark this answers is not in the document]
CyanVoxel: Paths with non-native separators are no longer normalized when loading from the .ini, but honestly this is probably a niche use case anyway and doesn't impact the actual functionality of the program. It's just a small visual quirk at this point. As long as nobody complains, I'm okay with the reduced number of casts.

CyanVoxel (approved):
Looks great!! I attempted to find a way to make the library splitter section only take up the minimum amount of space it needs instead of letting it grow along the title label, but couldn't find anything quickly that would do it. Given that it would only be a minor visual enhancement I won't worry about it too much right now. If you or I can't find a solution to it soon then perhaps opening up an issue for it after the fact would be best. After all, I think we'd both rather move onto more important fixes and additions than spend too much time fighting with Qt ;)

yedpodtrzitko:
> Looks great!! I attempted to find a way to make the library splitter section only take up the minimum amount of space it needs instead of letting it grow along the title label, but couldn't find anything quickly that would do it. Given that it would only be a minor visual enhancement I won't worry about it too much right now. If you or I can't find a solution to it soon then perhaps opening up an issue for it after the fact would be best. After all, I think we'd both rather move onto more important fixes and additions than spend too much time fighting with Qt ;)

seems like that got lost during rebasing and I didnt check after last push.
 It should be fixed now by `button_remove.setFixedWidth(30)`

CyanVoxel:
> seems like that got lost during rebasing and I didnt check after last push. It should be fixed now by `button_remove.setFixedWidth(30)`

The `button_remove` width issue wasn't actually present until I pulled just now to see what fix you were talking about. I was referring to the excess space that can surround the "Recent Libraries" label instead of having the maximum height of the section be equal to it's minimum height at any given time.
image

yedpodtrzitko:
> The `button_remove` width issue wasn't actually present until I pulled just now to see what fix you were talking about. I was referring to the excess space that can surround the "Recent Libraries" label instead of having the maximum height of the section be equal to it's minimum height at any given time. 

ok I see what you mean now. Check the behaviour after the latest commit I pushed.

CyanVoxel:
> ok I see what you mean now. Check the behaviour after the latest commit I pushed.

Awesome!! Works exactly how I'd expect! Every now and again it won't update the size when when I remove a library, but I can't predictably reproduce it. Overall though I'm more than happy to go ahead and pull this. Thank you again for all your hard work!

Issue #75, opened 2024-04-27
[suggestion] drag and drop outside the window


Vectasus:
It would be nice if one could find a file in TagStudio and then drag it towards a new destination, such as a file explorer, a website drop area or a chatroom.

Vectasus:
You can use `event.mimeData().hasUrls()` to detect files being dragged.
Here is a mock-up of a drop-target window
```python
import sys
from PySide6.QtWidgets import QApplication, QWidget
from PySide6.QtCore import Qt

class DragDropWindow(QWidget):
    def __init__(self):
        super().__init__()
        self.setAcceptDrops(True)
        self.setWindowTitle("Drag and Drop Files")
        self.setGeometry(100, 100, 400, 300)

    def dragEnterEvent(self, event):
        if event.mimeData().hasUrls():
            event.accept()
        else:
            event.ignore()

    def dragMoveEvent(self, event):
        if event.mimeData().hasUrls():
            event.accept()
        else:
            event.ignore()

    def dropEvent(self, event):
        for url in event.mimeData().urls():
            file_path = url.toLocalFile()
            print("Dropped file:", file_path)

if __name__ == "__main__":
    app = QApplication(sys.argv)
    window = DragDropWindow()
    window.show()
    sys.exit(app.exec_())
```

CyanVoxel:
Implemented in #153

PR #153, opened 2024-05-09
Drag and drop files in and out of TagStudio

Creepler13:
Added the ability to drop files into TagStudio, which adds them to the library

A popup will ask how to resume when some/all of the files names already exist in the librarys root folder
- Skip
- Override
- Rename
- Cancel
  
I put it in the modal folder because i plan/ned to add a window that allows you to import files from discord, google, etc
since these are links and they need extra logic to either download or safe them as text.

I used the method described in  #75 

Edit: Dragging out of TagStudio was added


tagstudio/src/qt/modals/drop_import.py
```diff
@@ -0,0 +1,79 @@
...
+from PySide6.QtGui import QDropEvent, QDragEnterEvent, QDragMoveEvent
+from PySide6.QtWidgets import QMessageBox
+
+if typing.TYPE_CHECKING:
+    from src.qt.ts_qt import QtDriver
+
+
+def dropEvent(driver: "QtDriver", event: QDropEvent):
+    if event.mimeData().hasUrls():
+        if event.mimeData().urls()[0].isLocalFile():
```
yedpodtrzitko, line 14: just a styling nit: it would be cleaner if the logic here would be reversed, so the code would be less indented:

```
if not event.mimeData().hasUrls():
    return

if not if event.mimeData().urls()[0].isLocalFile():
    return

# original code, but without extra 8 spaces
```

tagstudio/src/qt/modals/drop_import.py
```diff
@@ -0,0 +1,79 @@
...
+            urls = event.mimeData().urls()
+            duplicate_filesnames = []
+            for url in urls:
+                if os.path.exists(driver.lib.library_dir + "/" + url.fileName()):
```
yedpodtrzitko, line 18: there's another PR with replacing `os.path.` with `pathlib.Path`. Would be nice to have it everywhere for consistency

yedpodtrzitko:
doesnt work for me:

```
  File "/Users/yed/dev/python/TagStudio/tagstudio/tag_studio.py", line 9, in <module>
  File "/Users/yed/dev/python/TagStudio/tagstudio/src/qt/ts_qt.py", line 85, in <module>
    from src.qt.modals import (
  File "/Users/yed/dev/python/TagStudio/tagstudio/src/qt/modals/drop_import.py", line 30
    f"The files  {", ".join(map(lambda url:url.fileName(),duplicate_filesnames))}  have filenames that already exist in the library folder."
                   ^
SyntaxError: f-string: expecting '}'
```

tagstudio/src/qt/modals/drop_import.py
```diff
@@ -0,0 +1,82 @@
...
+    if not event.mimeData().hasUrls():
+        return
+
+    if not event.mimeData().urls()[0].isLocalFile():
+        return
+
+    urls = event.mimeData().urls()
+    duplicate_filesnames = []
+    for url in urls:
+        if Path(driver.lib.library_dir + "/" + url.fileName()).exists():
```
yedpodtrzitko, line 22:
```suggestion
        if Path(driver.lib.library_dir / url.fileName()).exists():
```

yedpodtrzitko:
beside the error it works as expected, nice job.

tagstudio/src/qt/modals/drop_import.py
```diff
@@ -0,0 +1,82 @@
...
+            duplicate_filesnames.append(url)
+
+    ret = -1
+
+    if len(duplicate_filesnames) > 0:
+        msgBox = QMessageBox()
+        msgBox.setText(
+            f"The files  {", ".join(map(lambda url:url.fileName(),duplicate_filesnames))}  have filenames that already exist in the library folder."
```
yedpodtrzitko, line 30:
```suggestion
            f"The files  {', '.join(map(lambda url:url.fileName(),duplicate_filesnames))}  have filenames that already exist in the library folder."
```

tagstudio/src/qt/modals/drop_import.py
Creepler13: This will only work once the pathLib PR is done

yedpodtrzitko: good catch. So it could work like this (suggestion above updated).

Creepler13: Yes that should work

yedpodtrzitko (approved):

Creepler13:
Forgot about folders. They work now

tagstudio/src/qt/modals/drop_import.py
```diff
@@ -0,0 +1,150 @@
...
+        event.ignore()
+
+
+def import_files(driver: "QtDriver", urls):
+    files: list[Path] = []
+    dirs_in_root: list[Path] = []
+    duplicate_files: list[Path] = []
+
+    for url in urls:
+        if url.isLocalFile():
```
yedpodtrzitko, line 42: again me with the same suggestion - there's no `else:` condition, so the logic can be flipped to avoid the code indention:

```
if not url.isLocalFile():
    continue
```

Creepler13:
Added progress bars, helpfull especially for large amounts of or large files

Creepler13:
Added drag out of window support 

#75

Creepler13:
i give up, no more formatting

CyanVoxel:
> i give up, no more formatting

I'll take a look at it, I've been fighting with the MyPy stuff myself

tagstudio/src/qt/modals/drop_import.py
```diff
@@ -0,0 +1,251 @@
...
+    def import_files(self):
+        self.files: list[Path] = []
+        self.dirs_in_root: list[Path] = []
+        self.duplicate_files: list[Path] = []
+
+        create_progress_bar(
+            self.collect_files_to_import,
+            "Searching Files",
+            "Searching New Files...\nPreparing...",
+            lambda x: f'Searching New Files...\n{x[0]+1} File{'s' if x[0]+1 != 1 else ''} Found. {(f'{x[1]} Already exist in the library folders') if x[1]>0 else ''}',
```
yedpodtrzitko, line 89: it should be enough to use double quotes on the "edges":
```suggestion
            lambda x: f"Searching New Files...\n{x[0]+1} File{'s' if x[0]+1 != 1 else ''} Found. {(f'{x[1]} Already exist in the library folders') if x[1]>0 else ''}",
```

Generally I'd recommend to extract this into a standalone function because this is way too much logic in one f-string.

CyanVoxel:
I apologize for the delay on getting this pulled. Thank you for all your work on this!

PR #159, opened 2024-05-11
Change QSettings behavior to work with executables

Loran425:
see below


Loran425:
Changes Necessary to provided expected behavior for executable files when saving/loading QSettings to disk.

Defaults QSettings to default behavior as specified in [QSettings Docs](https://doc.qt.io/qtforpython-6/PySide6/QtCore/QSettings.html#locations-where-application-settings-are-stored) 

A specific config file can be specified with the -c/--config-file argument with the following behavior.
- Not Specified (default): Follow QSettings for file location
- An existing file is specified: File used as settings file (may have implications as more things use the QSettings object)
- An existing directory is specified: Add `TagStudio.ini` to the path and use that
  - If the file exists it will be used
  - If the file doesn't exist it will be created
- A path is specified where the directory exist and the filename ends with `.ini` but the file does not exist: The file will be created
- Any other case: Revert to the default location

In all cases the location of the settings file used is logged to the console.

tagstudio/src/qt/ts_qt.py
```diff
@@ -189,9 +186,42 @@ def __init__(self, core, args):
...
-            QSettings.IniFormat, QSettings.UserScope, "tagstudio", "TagStudio"
-        )
+        if self.args.config_file:
+            path = Path(self.args.config_file)
+            if path.is_dir():
+                path = path / "TagStudio.ini"
+                self.settings = QSettings(str(path), QSettings.IniFormat)
+                logging.info(
+                    f"[QT DRIVER] Directory provided defaulting to TagStudio.ini in directory, using {self.settings.fileName()}"
+                )
```
yedpodtrzitko, line 196: I wouldnt do this. The user should provide valid path to existing config file, or it should fail loudly. Imagine the user makes a typo in the path (wrong autocomplete  choice etc.) and instead of `/path/to/TagStudio.ini`, they would type `/path/to/Tarkov/`, so it would create new empty config there. Most likely not desired behaviour.

tagstudio/src/qt/ts_qt.py
```diff
@@ -189,9 +186,42 @@ def __init__(self, core, args):
...
+            elif path.is_file():
+                self.settings = QSettings(str(path), QSettings.IniFormat)
+                logging.info(
+                    f"[QT DRIVER] Config File exists, using {self.settings.fileName()}"
+                )
+            else:
+                if path.suffix == ".ini" and path.parent.is_dir():
```
yedpodtrzitko, line 203: I dont understand this condition - `path.parent.is_dir()` will be always true, no?

But I'd say the same as in comment above - just handle the case where a path to existing file is provided, fail loudly otherwise.

yedpodtrzitko (changes requested):

yedpodtrzitko:
I wouldnt be doing any of the guessing games where path is specified but the config doesnt exist there. But maybe other people have different stance on this.

tagstudio/src/qt/ts_qt.py
Loran425: Not necessarily, 
If config_file is passed a path like `/home/Loran425/projects/TagStudio/TagStudio.ini` the path will always be setup properly from the string but if any of the folders in that chain don't exist then the folder wouldn't exist.

tagstudio/src/qt/ts_qt.py
Loran425: I think the only case that needs to be handled is how to create a .ini file from the command line argument then. If I want to use a new config location then I shouldn't need to go there and manually create the config file, I should be able to point the program at where I want it to save the config and then it creates the blank config file there.

tagstudio/src/qt/ts_qt.py
yedpodtrzitko: I see. In that case `.exists()` would be probably better choice.

Loran425: If we cut out the guess work and the file is expected to already exists yes `.exists()` would 100% be the proper call.

This still cuts out the ability to specify a new config file location from the command line arguments which means a user trying to specify a non-standard location would have to go `touch` the file prior to running TagStudio with a config flag pointing at that location, this doesn't seem ideal.

I think at this point, I'd agree the guess work needs to go, but I'm not sold on the config file existence requirement.

Loran425:
New behavior would be this 

A config file can be specified with the -c/--config-file argument with the following behavior.
- Not Specified (default): Follow QSettings `IniFormat` `UserScope` for file location
- A file is specified: File used as settings file (WARN and create if no config found at path)

In all cases the location of the config file used is logged to the console.

CyanVoxel:
> New behavior would be this
> 
> A config file can be specified with the -c/--config-file argument with the following behavior.
> 
> * Not Specified (default): Follow QSettings `IniFormat` `UserScope` for file location
> * A file is specified: File used as settings file (WARN and create if no config found at path)
> 
> In all cases the location of the config file used is logged to the console.

I think as long as there's fair warning of there being no config file at the specified path then this should be alright. I think it would strike a balance between not failing silently (even if it isn't loud) as well as keeping the expected behavior of creating a config file at the specified location on a first run or not letting the program fail if the config was removed at some point, in the same way I would expect it to not fail if the config in the default userscope location was removed.

tagstudio/src/qt/ts_qt.py
```diff
@@ -186,9 +183,21 @@ def __init__(self, core, args):
...
+                    f"[QT DRIVER] Config File does not exist creating {str(path)}"
+                )
+            logging.info(f"[QT DRIVER] Using Config File {str(path)}")
+            self.settings = QSettings(str(path), QSettings.IniFormat)
+        else:
+            self.settings = QSettings(
+                QSettings.IniFormat, QSettings.UserScope, "TagStudio", "TagStudio"
+            )
+            logging.info(
+                f"[QT DRIVER] Config File not specified, defaulting to {self.settings.fileName()}"
```
Loran425, line 199: This should probably have better wording, right now it sounds like the user did something wrong if they didn't specify a config file
```suggestion
                f"[QT DRIVER] Using local config stored in {self.settings.fileName()}"
```

PR #160, opened 2024-05-11
Refactor: remove __init__.py files meant for Python versions before 3.3 

xarvex:
This does mess with a large amount of imports, as the system was being
misused to re-export submodules. This change is necessary if PyInstaller
is to work.

This has been carefully checked that no imports were "lost".


tagstudio/src/qt/ts_qt.py
```diff
@@ -74,24 +74,21 @@
...
+from src.qt.widgets.progress import ProgressWidget
+from src.qt.widgets.preview_panel import PreviewPanel
+from src.qt.widgets.item_thumb import ItemThumb
+from src.qt.modals.build_tag import BuildTagPanel
+from src.qt.modals.tag_database import TagDatabasePanel
+from src.qt.modals.file_extension import FileExtensionModal
+from src.qt.modals.fix_unlinked import FixUnlinkedEntriesModal
+from src.qt.modals.fix_dupes import FixDupeFilesModal
+from src.qt.modals.folders_to_tags import FoldersToTagsModal
+
```
CyanVoxel, line 91: 🤓☝️ 
```suggestion
```

PR #156, opened 2024-05-10
Replace use of os.path with pathlib

Icosahunter:

PR #161, opened 2024-05-11
ci: add mypy check

yedpodtrzitko:
This add another automatic check beside ruff.

For example in #156 it caught all the errors with wrong method names like `isfile()`/`is_file()`, eg. https://github.com/TagStudioDev/TagStudio/pull/156#discussion_r1596171363

This might hurt a bit occasionally, because sometime it will give false positives, but generally it should prevent various errors slipping through, and it's super useful when doing bigger refactorings. 

The configuration is now set to slightly permissive mode to keep the diff small-ish. When the pipeline will be in place, we can keep making it more strict.


tagstudio/src/core/library.py
```diff
@@ -276,18 +279,19 @@ def __repr__(self) -> str:
...
     def __eq__(self, __value: object) -> bool:
+        __value = cast(Self, __value)
         if os.name == "nt":
             return (
-                int(self.id) == int(__value.id_)
-                and self.filename.lower() == __value.filename.lower()
-                and self.path.lower() == __value.path.lower()
+                int(self.id) == int(__value.id)
+                and self.filename.lower() == __value.filename.lower()  # type: ignore
+                and self.path.lower() == __value.path.lower()  # type: ignore
```
yedpodtrzitko, line 287: looking at this, I'm not convinced this has ever worked, because this object looks like it doesnt have `.filename` nor `.path` property 🤔

CyanVoxel:
I'm ready to pull this if you're all set

yedpodtrzitko:
@CyanVoxel please go ahead, thanks.

PR #165, opened 2024-05-12
ci: try to run the app

yedpodtrzitko:
When working on some of previous changes, it caused an issue when running the app without me noticing. This should run the app in CI to see if no error happens.

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -193,6 +193,9 @@ class QtDriver(QObject):
         )
 
         max_threads = os.cpu_count()
+        if args.ci:
+            # spawn only single worker in CI environment
+            max_threads = 1
         for i in range(max_threads):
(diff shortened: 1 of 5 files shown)


tagstudio/src/qt/ts_qt.py
```diff
@@ -238,9 +241,6 @@ def start(self):
 
         # Handle OS signals
         self.setup_signals()
-        timer = QTimer()
-        timer.start(500)
-        timer.timeout.connect(lambda: None)
```
CyanVoxel, line 243: The removal of the timer here stops the Qt thread from pausing to process console input, therefore breaking the ability to handle SIGTERMs sent from the console.

yedpodtrzitko: Ah that's what this is for. I thought it's some leftover code doing nothing. I added a comment there to make it more obvious

CyanVoxel (approved):
Looks good! Seems to be running fine as far as I can tell. Thanks for this!

Issue #164, opened 2024-05-12
Large file misidentified as a decompression bomb


Equinox3141:
There's a file in my library that can't be interacted with or shown (other than a red crossed out tag). Due to it being misinterpreted as a decompression bomb DOS attack. Warning message below.

> D:\TagStudio-main\.venv\Lib\site-packages\PIL\Image.py:3218: DecompressionBombWarning: Image size (108603828 pixels) exceeds limit of 89478485 pixels, could be decompression bomb DOS attack.
  warnings.warn(

I was confused by this, so after some searching through the folder that I was looking at, I found that the file it was struggling with was a very large .png file that was about 20000x14000 px.

Thesacraft:
Hi,
This error is caused by a library and is kind of fixable:
You can suppress this error by setting `Image.MAX_IMAGE_PIXELS=None`
Or 
`IMAGE.MAX_IMAGE_PIXELS=108603829`
But i don't think it is a good idea to have safety precautions disabled or weakened in production code. But I'm not sure what others think about it.

CyanVoxel:
Fixed in #166

PR #166, opened 2024-05-12
Fix pillow decompression bomb error mentioned in #164

Thesacraft:
Prevents DecompressionBombError that is triggered by pictures with to many pixels.
If the picture is to big it prevents the thumbnail from being rendered but allows the user to tag the file etc. (Just without a thumbnail)
This should fix Issue #164


yedpodtrzitko:
This code is in the codebase in one more place, so it would be worth fix it there as well, and add the missing the `exif_transpose()` there

https://github.com/TagStudioDev/TagStudio/blob/3aa71d6f8abb67f309f6d66ef4e6c570332d82be/tagstudio/src/qt/widgets/preview_panel.py#L371-L377

Also it's a bit out of scope, but when you are touching this code already, it would be worth it to DRY it and extract it into a function instead of having exactly the same block of code in three different places.

tagstudio/src/qt/widgets/preview_panel.py
```diff
@@ -407,6 +408,10 @@ def update_widgets(self):
 
                     except (UnidentifiedImageError, FileNotFoundError, cv2.error):
                         pass
+                    except DecompressionBombError as e:
+                        logging.info(
```
yedpodtrzitko, line 412: maybe these except blocks can be merged together, what do you think?
```suggestion
                    except (DecompressionBombError, UnidentifiedImageError, FileNotFoundError, cv2.error) as e:
                        logging.exception(
```

Thesacraft:
Ohh i didn't see that one thank you!
I will look into moving this code into its own function tommorow and all other changes(im currently procastinating and have a final exam tomorrow hahaha).
Just one more question:
Do you know why in the preview_panel only the Video and image types are handled (line 368 onwards) and only in the Thumbrenderer the text types are handled?

tagstudio/src/qt/widgets/preview_panel.py
Thesacraft: Probably smart yep

yedpodtrzitko:
> Do you know why in the preview_panel only the Video and image types are handled (line 368 onwards) and only in the Thumbrenderer handles text types?

no idea, maybe person who was adding the code for text files overlooked the logic in preview panel, similarly as it happened here? :)

CyanVoxel:
> > Do you know why in the preview_panel only the Video and image types are handled (line 368 onwards) and only in the Thumbrenderer handles text types?
> 
> no idea, maybe person who was adding the code for text files overlooked the logic in preview panel, similarly as it happened here? :)

The foreboding line of 366: `# TODO: Do this somewhere else, this is just here temporarily.`

Now if *I* were to put myself in the mindset of whoever may have written this code... It looks like the image/video frames were being loaded here just to grab their dimensions to display in the preview panel. Absolutely *not* the right place to do this, as 366 warned, but still it's remained there. I bet whoever wrote this was planning on displaying specs from other filetypes like word count or line length for text files, video length for videos, and things like that - but knew that allll off that would be better off in a different class which `preview_panel` could just call from.

That's just my speculation though 😉

Thesacraft:
> This code is in the codebase in one more place, so it would be worth fix it there as well, and add the missing the `exif_transpose()` there
> 
> https://github.com/TagStudioDev/TagStudio/blob/3aa71d6f8abb67f309f6d66ef4e6c570332d82be/tagstudio/src/qt/widgets/preview_panel.py#L371-L377
> 
> Also it's a bit out of scope, but when you are touching this code already, it would be worth it to DRY it and extract it into a function instead of having exactly the same block of code in three different places.

I intentionally did not DRY it because i'm thinking about a way for it to be moved completly out of there.
I DRYed the other two occurances in PR #167 (just to seperate things properly), or would you prefer that the DecompressionBombError is also just fixed in #167 and this one is closed?

Thesacraft:
I think i handled every `DecompressionBombError` that were occuring by now

tagstudio/src/cli/ts_cli.py
```diff
@@ -1109,24 +1114,32 @@ def create_collage(self) -> str:
...
+                                                    "RGB",
+                                                    (thumb_size, thumb_size),
+                                                    color,
+                                                ),
+                                            )
+                                        collage.paste(
+                                            pic, (y * thumb_size, x * thumb_size)
+                                        )
+                                except DecompressionBombError as e:
+                                    print(f"[ERROR] One of the images was to big ({e})")
```
CyanVoxel, line 1141: Typo fix
```suggestion
                                    print(f"[ERROR] One of the images was too big ({e})")
```

tagstudio/src/qt/widgets/collage_icon.py
```diff
@@ -95,22 +96,25 @@ def render(
...
+                                pic = pic.resize(size)
+                            if data_tint_mode and color:
+                                pic = pic.convert(mode="RGB")
+                                pic = ImageChops.hard_light(
+                                    pic, Image.new("RGB", size, color)
+                                )
+                            # collage.paste(pic, (y*thumb_size, x*thumb_size))
+                            self.rendered.emit(pic)
+                    except DecompressionBombError as e:
+                        logging.info(f"[ERROR] One of the images was to big ({e})")
```
CyanVoxel, line 117: Typo fix
```suggestion
                        logging.info(f"[ERROR] One of the images was too big ({e})")
```

PR #167, opened 2024-05-13
Refactor ThumbRenderer to be less repetitive

Thesacraft:
I refactored the Thumbrenderer render methods and merged them into one because there were only small differences.
The outputed preview is the same as the one that used render/render_big, all small differences between the output of these methods is kept( for example the four color gradient background for the small thumbs and no for the big preview)
I am open for any changes :)

### DIFF
--- a/.github/workflows/apprun.yaml
+++ b/.github/workflows/apprun.yaml
@@ -1,0 +1,50 @@
+name: PySide App Test
+
+on: [ push, pull_request ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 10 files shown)


Thesacraft:
@CyanVoxel If its ok i would close this pr and make a new one based on the current main ( i will change the things that i had changed here) but this is now just chaos because pycharm triedd to resolve the conflicts

CyanVoxel:
> @CyanVoxel If its ok i would close this pr and make a new one based on the current main ( i will change the things that i had changed here) but this is now just chaos because pycharm tried to resolve the conflicts

All good!

PR #168, opened 2024-05-13
Refactor Thumbrenderer

Thesacraft:
I refactored Thumbrenderer to remove duplicate code.
I merged the render and render_big methods into one _render method
I'm open to any suggestions and changes


tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -96,65 +95,53 @@ def render(
...
+        if ThumbRenderer.font_pixel_ratio != pixel_ratio:
+            ThumbRenderer.font_pixel_ratio = pixel_ratio
             ThumbRenderer.ext_font = ImageFont.truetype(
                 os.path.normpath(
                     f"{Path(__file__).parents[3]}/resources/qt/fonts/Oxanium-Bold.ttf"
                 ),
                 math.floor(12 * ThumbRenderer.font_pixel_ratio),
             )
 
-        if isLoading or filepath:
```
Thesacraft, line 119: Removed this because it is never called without one of these options been set(but if it should not be removed let me know)

tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -96,65 +95,53 @@ def render(
...
+        """Internal renderer. Renders an entry/element thumbnail for the GUI."""
+        image: Image.Image = None
+        pixmap: QPixmap = None
+        final: Image.Image = None
         extension: str = None
-        broken_thumb = False
-        # adj_font_size = math.floor(12 * pixelRatio)
-        if ThumbRenderer.font_pixel_ratio != pixelRatio:
-            ThumbRenderer.font_pixel_ratio = pixelRatio
+        resampling_method = Image.Resampling.BILINEAR
```
Thesacraft, line 108: added this to make it easier to change the resampling method used

tagstudio/src/qt/widgets/thumb_renderer.py
CyanVoxel: Seems to work fine without this to me 🤷‍♂️

tagstudio/src/qt/widgets/thumb_renderer.py
CyanVoxel: Guessing you saw #174! Much appreciated! 😁

Thesacraft: yep haha

CyanVoxel:
Overall this looks pretty good to me! The thumb renderer is about to be sort of a hot spot in the code with lots of activity (#171, #175, #179, #183, [raw-image-previews](https://github.com/TagStudioDev/TagStudio/tree/raw-image-previews), etc.), so while I'm open to an even larger refactor things will probably need to be done in smaller chunks over time after this PR. Consolidating the `render` and `render_big` methods was the huge thing that needed done though, so I'm happy to have that improved in time for the next round of fixes and additions!

With all that being said, I'll be pulling #156 before this which refactors out all uses of `os.path` in favor of the `pathlib` library going forward. I'll also be pulling the MyPy workflow from #169 which will also shake things up here. I apologize for the additional refactoring/formatting.

Thesacraft:
Thanks for resolving the conflicts. i was just waiting for the os.paths pr to get merged and do it all in one big cleanup, haha. But i really appreciate the help.

PR #188, opened 2024-05-17
Update README.md

tadpole256:
In response to an existing issue, I updated the ReadMe to reflect the ability to click "Ctrl-T" to create a new tag. 

I also removed the Warning about having no method to view all tags, as the warning is out of date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ Click the “+” button at the end of the Tags list, and search for tags to add
 
 ### Creating Tags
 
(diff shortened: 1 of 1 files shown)


yedpodtrzitko:
it could be worth it to create standalone section `Hotkeys` with all available hotkeys and shortcuts like this, what do you think?

tadpole256:
I think it could be worth while to have a Hotkeys section, as a quick reference. That would take a bit more time to write.

yedpodtrzitko:
Just start with one. if someone is skimming the readme, they not gonna find it buried in paragraph of text anyway.

CyanVoxel:
Closing this as it hasn't had any activity since the last time feedback was given, and the changes here are subjectively minor.

PR #192, opened 2024-05-18
use reviewdog for mypy job

yedpodtrzitko:
this improves the way how mypy errors can be checked. Instead of going to the logs, the errors will be added as comments on relevant lines, see:

https://github.com/yedpodtrzitko/TagStudio/pull/2/files#file-tagstudio-src-core-library-py-L529


yedpodtrzitko:
cc @CyanVoxel

PR #197, opened 2024-05-20
A file system provider

DannyAlas:
An efficient file system provider. It uses multiple processes to scan through many files efficiently and provides a centralized place to implement file system-related tasks. Based off #187 - explanations [1](https://github.com/TagStudioDev/TagStudio/pull/187#issuecomment-2118348392), [2](https://github.com/TagStudioDev/TagStudio/pull/187#issuecomment-2119369851)

Implements a cross-platform method for acquiring system unique file IDs -  fixing/minimizing [relinking](https://github.com/TagStudioDev/TagStudio/#:~:text=Relinking%20Renamed/Moved%20Files) issues.

Implements file system watching using the Qt provided [QFileSystemWatcher](https://doc.qt.io/qt-6/qfilesystemwatcher.html). As an alternative to #125 (still requires implementation into Library class)

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -586,6 +586,7 @@ class QtDriver(DriverMixin, QObject):
 
         self.lib.close()
 
+        self.thumb_job_queue.queue.clear()
         if is_shutdown:
             # no need to do other things on shutdown
             return


BilliAlpha:
It would be interesting to provide abstraction and create a generic file provider interface (with a default implementation per OS filesystem) so that we could later add support for other file systems (be it on the network encrypted or really anything).
It would allow support for S3, FTP, SSHFS, GPG encrypted files etc ...

PR #199, opened 2024-05-21
add unit tests for Tag class

DJStowe:
Added unit tests just for the Tag class. Tests the following:
add_subtag
remove_subtag
remove_subtag if id is not in subtag_ids
debug_name
display_name with shorthand
display_name without shorthand
display_name with no subtag_ids

Also added tests for edge cases with display name when tags don't have names but I commented those out for now since we probably don't want to display them in that format. ex result for tag & subtag having no name " ()"

Added pytest-mock to requirements-dev.txt
     I'm not sure what version to add, I tested on both v3.0.0 & v3.14.0 
     
I have not been following the path.os changes so the header in test_tags.py might need to be changed


requirements-dev.txt
```diff
@@ -3,3 +3,4 @@ pre-commit==3.7.0
 pytest==8.2.0
 Pyinstaller==6.6.0
 mypy==1.10.0
+pytest-mock 3.14.0
```
yedpodtrzitko, line 6:
```suggestion
pytest-mock==3.14.0
```

tagstudio/tests/core/test_tags.py
```diff
@@ -1,18 +1,161 @@
...
+
+@pytest.fixture
+def test_tag2():
+    return Tag(id=2, name="tag2 Name", shorthand="tag2", aliases=[], subtags_ids=[3], color="")
+
+@pytest.fixture
+def test_tag3():
+    return Tag(id=3, name="tag3 Name", shorthand="tag3", aliases=[], subtags_ids=[], color="")
+
+class TestTags():
```
yedpodtrzitko, line 29: isnt this wrapping class superfluous?

tagstudio/tests/core/test_tags.py
```diff
@@ -1,18 +1,161 @@
...
-        color="",
-    )
-    assert tag
+        mock_library.get_tag.return_value = test_tag2
+        result = test_tag1.display_name(mock_library)
+        assert result == "tag1 Name (tag2 Name)"
+'''
 
+if __name__ == "__main__":
+    pytest.main()
```
yedpodtrzitko, line 160: I dont think this is needed.

yedpodtrzitko (commented):
Thanks for the PR. Ideally please add a CI job which will run the tests. Otherwise having tests no one runs is like having no tests at all.

tagstudio/tests/core/test_tags.py
yedpodtrzitko: I dont understand how is a failing test (if I understand correctly) is related to the wrapping class. 
Pytest tests are usually done as a top-class functions. If a test is failing, add decorator `@pytest.mark.skip` or comment it out 🤔

tagstudio/tests/core/test_tags.py
DJStowe: You're right it isn't needed. I've removed it in the latest commits.

DJStowe:
> Thanks for the PR. Ideally please add a CI job which will run the tests. Otherwise having tests no one runs is like having no tests at all.

I've added runtest.yaml This is the first time I've worked with github actions but I think it's working as intended.

yedpodtrzitko:
The interface of `Tag` has changed significantly, and there are some tests covering most of the Tag scenarios, so I will close this, sorry.

PR #204, opened 2024-05-21
Duplicate Entry Handling (Fixes #179)

### LINKED ISSUE #179: [BUG] "Fix Unlinked Entries"/"Refresh Directories" Creates Duplicate Entries for Moved Files
### TagStudio Version
Alpha v9.2.0
### Operating System
Windows 10 22H2
### Describe The Issue
The "Fixed Unlinked Entries" does not take into account entries for existing files when relinking entries, thus creating duplicate entries for the same files.
### Steps to Reproduce
1. Create a library full of files
2. Move those files into a subfolder (this breaks the links)
3. Refresh the library directories (this creates new entries for the files while not touching the old unlinked ones)
4. Use the "Fix Unlinked Entries" option. The old entries are now relinked, but without reconciling the duplicate entries that were created in the meantime
image

CyanVoxel:
*Initially this was intended to be a quick patch that ended up needing a bigger fix, especially when it came to conflicts with the workflows that I was unable to reproduce on my machine. Either way, I should really be making my changes via PRs in the first place...*

This PR makes the following changes:
- Running "Fix Unlinked Entries" will no longer result in duplicate entries if the directory was refreshed after the original entries became unlinked.
- A "Duplicate Entries" section is added to the "Fix Unlinked Entries" modal to help repair existing affected libraries.
    - Having duplicate entries like this is not something that's supposed to happen in the first place, so this section here is not a long-term solution.

There were a couple bodges in this involving Path casting and applying `# type: ignore` to problematic lines that set off MyPy, false positives or otherwise.


tagstudio/src/core/library.py
```diff
@@ -86,20 +86,20 @@ def __repr__(self) -> str:
         return self.__str__()
 
     def __eq__(self, __value: object) -> bool:
```
yedpodtrzitko, line 88: At this point there's no type check in this method. I'm not sure how how is that better than casting the variable, but if you want disable the type checking then you can add this decorator to the method declaration rather than adding `ignore` on each line with error.

```suggestion
    @typing.no_type_check
    def __eq__(self, __value: object) -> bool:
```

CyanVoxel: Thanks for the info on the `@typing.no_type_check` decorator, I'll definitely use that if needed going forward 👍 

The casting itself was causing me the following error when invoking the \_\_eq\_\_ method:
```py
  File "F:\GitHub\TagStudio/tagstudio\src\qt\helpers\custom_runnable.py", line 18, in run
    self.function()
  File "F:\GitHub\TagStudio/tagstudio\src\qt\modals\merge_dupe_entries.py", line 44, in <lambda>
    r = CustomRunnable(lambda: iterator.run())
                               ^^^^^^^^^^^^^^
  File "F:\GitHub\TagStudio/tagstudio\src\qt\helpers\function_iterator.py", line 20, in run
    for i in self.iterable():
  File "F:\GitHub\TagStudio/tagstudio\src\core\library.py", line 1043, in merge_dupe_entries
    self.entries.remove(id_to_entry_map[id])
  File "F:\GitHub\TagStudio/tagstudio\src\core\library.py", line 92, in __eq__
    int(self.id) == int(__value.id)
                        ^^^^^^^^^^
AttributeError: type object 'object' has no attribute 'id'
```
I can reproduce the error by comparing any two Entry objects:
```py
entry_1 = Entry(1, "file_2", "path_2", [])
        entry_2 = Entry(2, "file_2", "path_2", [])
        logging.info(entry_1 == entry_2)
```

It looks like the object cast is just stripping the Entry's attributes down and turning it back into a bare object, right? I'm not sure how the cast is intended to function here.

yedpodtrzitko: right, my bad. It should be this instead:
```
__value = cast(Self, __value)
```

tagstudio/src/core/library.py
```diff
@@ -86,20 +86,20 @@ def __repr__(self) -> str:
...
-        __value = cast(Self, object)
+        # __value = cast(Self, object)
```
yedpodtrzitko, line 89:
```suggestion
        __value = cast(Self, __value)
```

PR #205, opened 2024-05-21
feat: toggle sidebar items

yedpodtrzitko:
Add ability to toggle sidebar items (thumbnail etc.). See screenshots below.

I can imagine the window for adding fields to Entry could fit in the sidebar as well instead of appearing as a popup window. This is a laying ground for it. Let me know if this somehow fits in your vision. If not, it's fine I'll keep it just in my fork.

One thing I wanted to do is to hide the whole Splitter in case all sidebar items are hidden, but I couldnt figure out how to do it, so maybe sometime in future.

For example when I want to work with fields, I want that widget occupy as much space as possible instead of let's say the thumbnail. So I can hide the thumbnail. Super useful on a small screen.

Screenshot 2024-05-22 at 10 38 54

### DIFF
--- a/tagstudio/src/qt/widgets/preview_panel.py
+++ b/tagstudio/src/qt/widgets/preview_panel.py
@@ -77,8 +77,6 @@ class PreviewPanel(QWidget):
         self.image_ratio: float = 1.0
 
         self.image_container = QWidget()
-        image_layout = QHBoxLayout(self.image_container)
-        image_layout.setContentsMargins(0, 0, 0, 0)
 
         self.open_file_action = QAction("Open file", self)
         self.open_explorer_action = QAction("Open file in explorer", self)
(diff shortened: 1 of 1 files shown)


yedpodtrzitko:
together with the widget for adding fields in sidebar

Screenshot 2024-05-22 at 12 21 20

CyanVoxel:
Sorry for the late comment on this, I've been kicking this one around in my head while prioritizing some other fixes.

I think a sidebar like this has a lot of potential, and the ability to toggle items could become even more useful as more items are potentially added (see 8359ae58ad5d482bbd3daf9721e9952301f208e8). I can see this turning into an even more customizable bar with the ability to set which toggles and options appear in a (much needed) settings panel, somewhat in the vein of VSCode.

Not to create too many branches, but I'm thinking about creating a branch for new UI ideas such as this to test out and tweak before eventually pulling into the main branch. That way it can see some use instead of sitting in a PR while I ponder the long term direction of the features.

yedpodtrzitko:
this doesnt look like it will get merged any time soon, so I will close it.

Issue #92, opened 2024-04-29
[BUG] Adding 2 Tags field to multiple selected files breaks Tag assignation


SirPhilip3:
### Steps to reproduce
1. Select two or more files
2. Add the Tags field 
3. Add another Tags field
4. You can now add Tags only to the first Tags field , the second will always remain empty and the tags added to this field will be added to the first field instead 

https://github.com/TagStudioDev/TagStudio/assets/118560349/b01e3cc3-2216-46df-8636-6c150a673fde

Loran425:
This is currently a known bug, that was caused by some of the changes allowing multiselection to work. 
I believe the plan is to fix this with the JSON->DB swap that is in progress but for the moment limiting to a single tag group of each type (tags/content tags/meta tags) is probably the easiest way to prevent this from occurring.

A library loading bug was causing TagStudio to not recognize existing tag groups and add copies, this unintended multiple group creation should be resolved if #95 is merged but this particular issue having multiple tag groups remains. 

From the UI you can remove the second tag group without issue if you decide to work with one tag group.

PR #220, opened 2024-05-24
Fix bug 213

### LINKED ISSUE #213: The Date fields are not working.
Sw version: 9.2.1

When user want to add the Date fields (Date, Date Created, Date_x...) then the filed is added without date or the filed have old data from another filed.

Actions:
1. Click on the picture
2. Click on the "Add Field" button
3. Choose filed with the date
4. Add the filed

Expected:
- new data filed is added with the right data format.
Observed:
- Every date type filed is empty or have values form other fields.

image
image

bigjohn98pl:
I added "container.set_inner_widget(inner_container)" in the exception to fix #213 
I don't know if this method is temporary or not, so I don't changed anything, but I think this need some rework.  The default value shoud be the dates form the file metadata. 

In library.py I see that  there is a add_generic_data_to_entry() for  this, but the add_field_to_entry() just set my date to ''

### DIFF
--- a/tagstudio/src/core/library.py
+++ b/tagstudio/src/core/library.py
@@ -837,7 +837,7 @@ class Library:
 
         logging.info(f"[LIBRARY] Saving Library Backup to Disk...")
         start_time = time.time()
-        filename = f'ts_library_backup_{datetime.datetime.utcnow().strftime("%F_%T").replace(":", "")}.json'
+        filename = f'ts_library_backup_{datetime.datetime.now(datetime.UTC).strftime("%F_%T").replace(":", "")}.json'
 
         self.verify_ts_folders()
         with open(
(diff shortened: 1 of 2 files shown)


CyanVoxel:
I apologize for the late comment on this.

The underlying issue of #213 is that the date fields aren't completely implemented yet, with the biggest missing piece being a UI for the user to edit them with. In reality this really shouldn't even be allowed to show as an option for the user to add, same with Collations. #213 is also encountering separate known issues stemming from #92.

So really there are two things going on here: Date Fields aren't finished yet, and it looks like there's a desire/expectation to see the modified and created dates of files. 

The first issue can be solved by either removing the ability to let users add date fields altogether, or by fully implementing the date field type complete with a UI and default values. 

The second issue can be solved by adding the file creation and modification dates to the file metadata section above the fields panel. This is data that's inherit to the file, while Date Fields would be intended to be populated by the user's data. This is a change I'm already working on separately.
image
Would the file/modification dates displayed like this be suitable for what you have in mind?

bigjohn98pl:
Yes, I think this is the best approach. I like to use the date of files to get what I want. It will be nice if there will be a feature that create tags bases on date. This can be similar to the tags that can be created based on the folder. With the information about date we can also manually add a date tag to a picture, or to a group of folders.

bigjohn98pl:
So if the Date Fields aren't finished yet then this merge is meaningless .

PR #223, opened 2024-05-26
Update to pyside6 version 6.7.1

Thesacraft:
I updatet the project to use pyside6 6.7.1 / Qt 6.7.1.
I had to change the way signals where disconnected because the RuntimeError was replaced by a RuntimeWarning which is not catched by a try-except block. I fixed one bug that was introduced but i did not find more so feel free to test for bugs.
I am open to changes :)


yedpodtrzitko:
please remove also this line pinning PySide version when running mypy in CI, thanks. It was added there as the types in Pyside 6.6.3 are broken.
https://github.com/TagStudioDev/TagStudio/blob/868b553670ddfd4bfd5c3822c6721a4c9b84f140/.github/workflows/mypy.yaml#L23

Thesacraft:
I don't know why the PySide App Test Error occurs here because in my main branch of Tagstudio(which is the same as the official main branch) it also occurs. The only idea i have is that a new numpy version is beeing used now and it fails there but i don't really know.

Thesacraft:
I just looked into it and the main branch here passed with this following version:
numpy-1.26.4-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (61 kB)
and mine uses:
numpy-2.0.0-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (60 kB)

Thesacraft:
I fixed it but is this ok in one pr or should i remove it from here and make another pr for that?

CyanVoxel:
> I fixed it but is this ok in one pr or should i remove it from here and make another pr for that @CyanVoxel ?

Yeah it's all good in this one

CyanVoxel:
Things seem to be working at a glance on my machine with the new PySide and numpy versions. I'm not exactly sure how the numpy issue has materialized but it's suddenly causing previously passed runs to fail everywhere.

My only immediate suggestion for this one would probably be to rename `CustomQPushButton` to something a bit more descriptive, maybe `QPushButtonWrapper`? Eh, it's hard to think of something good for it though...

Thesacraft:
> Things seem to be working at a glance on my machine with the new PySide and numpy versions. I'm not exactly sure how the numpy issue has materialized but it's suddenly causing previously passed runs to fail everywhere.

This is actually because a new numpy version was released about 9h ago and rawpy was built on the 1.x version of numpy and numpy 2.0 isn't supported. They seem to not have pinned the numpy version that is installed when installing rawpy.

CyanVoxel (approved):

PR #229, opened 2024-05-30
Create and activate venv via Nix Flake

seakrueger:
When running `nix develop` to generate a development shell, the python virtual environment will now automatically be created and dependencies from requirements.txt will be installed (and the venv will be activated). Effectively, this moves the setup portion of the TagStudio.sh to the flake and means the setup script for Linux/Mac is no longer needed when using nix. The README has been updated to reflect the new process of `nix develop` then `python3 tagstudio/tag_studio.py`. 

The python virtualenvironment package has been replaced with venvShellHook, which creates a virtual environment in the directory specified by $venvDir (in this case ./.venv). This also gives the postVenvCreation hook, where pip install is then executed. The previous shellHook needed to be moved to postShellHook to execute after the venv is created. `unset SOURCE_DATE_EPOCH` is probably not necessary, but is to solve a potential issue between pip and nix and appears in a lot of python flake examples.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,10 +64,10 @@ _Learn more about setting up a virtual environment [here](https://docs.python.or
 
   - Run the "TagStudio.sh" script and the program should launch! (Make sure that the script is marked as executable if on Linux). Note that launching from the script from outside of a terminal will not launch a terminal window with any debug or crash information. If you wish to see this information, just launch the shell script directly from your terminal with `./TagStudio.sh`.
 
-  - **NixOS** (TagStudio.sh)
+  - **NixOS** (Nix Flake)
     > [!WARNING]
     > Support for NixOS is still a work in progress.
-    - Use the provided `flake.nix` file to create and enter a working environment by running `nix develop`. Then, run the `TagStudio.sh` script.
+    - Use the provided [Flake](https://nixos.wiki/wiki/Flakes) to create and enter a working environment by running `nix develop`. Then, run the program via `python3 tagstudio/tag_studio.py` from the root directory.
 
 - **Any** (No Scripts)
 
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1717602782,
-        "narHash": "sha256-pL9jeus5QpX5R+9rsp3hhZ+uplVHscNJh8n8VpqscM0=",
+        "lastModified": 1718318537,
+        "narHash": "sha256-4Zu0RYRcAY/VWuu6awwq4opuiD//ahpc2aFHg2CWqFY=",
         "owner": "nixos",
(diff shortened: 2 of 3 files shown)


seakrueger:
This was finished prior to #149 being merged, which appears to have added a dep for libpulse and currently prevents the program from starting successfully. I'll create pr to fix that (and some other dep issues) soon, but this was bad timing on my part. Converting to draft

seakrueger:
Now based on #244

xarvex:
Same as [my comment](https://github.com/TagStudioDev/TagStudio/pull/244#issuecomment-2153282947) on #224, I've will follow up as soon as I've done a proper test and review, thank you!

CyanVoxel:
Just a heads up, the instructions for running/building from source have been moved from the README to CONTRIBUTING.md

seakrueger:
Currently, everything related to the actual creation of the venv works. The make blocker I'm on right now is mypy always fails with `[import-not-found]` for all third-party libraries, I don't know if that's something fixable or a fundamental broken-on-nix problem.

xarvex:
Thanks for the hard work on this. I'll see if I can figure something out myself for mypy.

xarvex:
Really couldn't figure out a way to get mypy to function without using from nixpkgs repos. For now the versions are the same as in the `requirements-dev.txt` but will have to see what this might mean when they drift apart.

yedpodtrzitko:
> Really couldn't figure out a way to get mypy to function without using from nixpkgs repos. For now the versions are the same as in the `requirements-dev.txt` but will have to see what this might mean when they drift apart.

I have no experience with NixOS, so sorry if I'm missing something obvious, but why is installation of `requirements-dev.txt` needed in first place? The dependencies are meant for developers, and in such case you will be most likely cloning the git repo instead of installing it via package manager, no?

xarvex:
> > Really couldn't figure out a way to get mypy to function without using from nixpkgs repos. For now the versions are the same as in the `requirements-dev.txt` but will have to see what this might mean when they drift apart.
> 
> 
> 
> I have no experience with NixOS, so sorry if I'm missing something obvious, but why is installation of `requirements-dev.txt` needed in first place? The dependencies are meant for developers, and in such case you will be most likely cloning the git repo instead of installing it via package manager, no?

This is a dev shell for developers to use, on NixOS you cannot just clone a repo and expect that to work and libraries don't exist the same as on a traditional distribution. This dev shell provides the utilities and libraries needed for a working development environment to use on TagStudio.
Building and running is something I will cook up at some point for NixOS, which as well needs to be handled in this flake.

xarvex:
This totally got lost with other responsibilities and life events, for which I am at fault for for not giving this the attention it needed sooner. I am going to pull this as-is right now to give me a starting place for a dev shell using [devenv](https://devenv.sh) which it appears is what the big thing in the Nix development ecosystem is. Be on the lookout for that being merged tonight or sometime tomorrow.
I have the template pulled up as a write this, so no excuses from me if that isn't done by then.

PR #232, opened 2024-05-31
Multi mode search system

YoyloNerd:
Adding a dropdown to select between 2 searching modes

- And (item contains all tags)
- Or (item contains a tag)


start_win.bat
```diff
@@ -1,2 +1,2 @@
 @echo off
-.venv\Scripts\python.exe .\TagStudio\tag_studio.py --ui qt %*
\ No newline at end of file
+.\venv\Scripts\python.exe .\TagStudio\tag_studio.py --ui qt %*
```
CyanVoxel, line 2: Was this change was accidentally included?
```suggestion
.venv\Scripts\python.exe .\TagStudio\tag_studio.py --ui qt %*
```

tagstudio/src/qt/ui/home_ui.py
```diff
@@ -1,185 +1,129 @@
...
-    QMainWindow,
-    QMenuBar,
-    QPushButton,
-    QScrollArea,
-    QSizePolicy,
-    QStatusBar,
-    QWidget,
-)
+
+from PyQt5 import QtCore, QtGui, QtWidgets
```
CyanVoxel, line 11: TagStudio uses PySide6 and not PyQt5

YoyloNerd: For some reason I didn't get the error when I ran MyPy locally but I've changed it

tagstudio/src/qt/main_window.py
```diff
@@ -181,15 +212,24 @@ def setupUi(self, MainWindow):
...
+  
+        # Search field
 		self.searchField.setPlaceholderText(
 			QCoreApplication.translate("MainWindow", u"Search Entries", None))
 		self.searchButton.setText(
 			QCoreApplication.translate("MainWindow", u"Search", None))
+  
+        # Search type selector
+		self.comboBox_2.setItemText(0, QCoreApplication.translate("MainWindow", "And (includes all tags)"))
+		self.comboBox_2.setItemText(1, QCoreApplication.translate("MainWindow", "Or (includes any tag)"))
```
CyanVoxel, line 229: Use "Title Case" for the combo box options
```suggestion
		self.comboBox_2.setItemText(0, QCoreApplication.translate("MainWindow", "And (Includes All Tags)"))
		self.comboBox_2.setItemText(1, QCoreApplication.translate("MainWindow", "Or (Includes Any Tag)"))
```

CyanVoxel (changes requested):
Looks pretty good! Just a few comments and nitpicks and this should be good to go:
1. `SEARCH_MODES` would be better off inside of `enums.py` for better readability instead of having situations such as `search_mode: int = 0` where it isn't clear to other people what's happening:
```py
class SearchMode(int, enum.Enum):
    """Operational modes for item searching."""
    
    AND = 0
    OR = 1
```
2. Remove leftover `print` statements from `library.py`
3. Use "Title Case" for the combo box options

Thank you again for all your work on this!

CyanVoxel (approved):
Looks awesome! Thank you again for all your work on this!

PR #233, opened 2024-06-01
Fix Raw Image Handling and Improve Text File Encoding Compatibility

CyanVoxel:
Some small bugfixes and improvements involving thumbnail and preview rendering:

- Fix RAW images not being loaded correctly in the preview panel
- Fix trying to read size data from null images
- Refactor `os.stat` to `<Path object>.stat()`
- Remove unnecessary upper/lower conversions
- Improve encoding compatibility beyond UTF-8 when reading text files
- Misc. code cleanup

### DIFF
--- a/tagstudio/src/core/utils/encoding.py
+++ b/tagstudio/src/core/utils/encoding.py
@@ -2,27 +2,26 @@
 # Licensed under the GPL-3.0 License.
 # Created for TagStudio: https://github.com/CyanVoxel/TagStudio
 
+from chardet.universaldetector import UniversalDetector
 from pathlib import Path
-from src.core.constants import ENCODINGS
 
 
-def get_text_encoding(filepath: Path) -> str:
+def detect_char_encoding(filepath: Path) -> str | None:
     """
-    Attempts to determine the encoding of a text file.
+    Attempts to detect the character encoding of a text file.
 
     Args:
-    file (TextIOWrapper): The text file to analyze.
+    filepath (Path): The path of the text file to analyze.
 
     Returns:
-    str: The assumed encoding. Defaults to utf-8.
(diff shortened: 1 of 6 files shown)


tagstudio/src/core/utils/encoding.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Args:
+    file (TextIOWrapper): The text file to analyze.
+
+    Returns:
+    str: The assumed encoding. Defaults to utf-8.
+    """
+
+    for encoding in ENCODINGS:
+        with open(filepath, "r", encoding=encoding) as text_file:
```
yedpodtrzitko, line 21: I dont think opening the file (up to) 5 times it's a good approach, even though `utf-8` might be the right choice in majority of cases.

I'd probably use existing library like [chardet](https://pypi.org/project/chardet/).

If you want to do it without introducing another dependency, opening the file once in a binary mode and then trying to the content decode might also work (code not tested):

    with open(file_path, 'rb') as text_file: # `rb` opens the file in binary mode 
        text_file = file.read(1024)  # Read the first 1024 bytes

    for encoding in ENCODINGS:
        try:
            decoded_data = raw_data.decode(encoding, errors='replace')
            if '�' not in decoded_data:
                return encoding
        except Exception:
            continue


and then it could be worth saving the file encoding somewhere in the library, so it doesnt need to be done every time 🤔

CyanVoxel: I was on the fence about adding another dependency, but I think I needed to hear it from someone else to realize that it's the better approach. I'll go and get a chardet implementation going 👍 

Also, saving the encoding somewhere would definitely be the way to go. Probably out of the scope of this PR, but that would be very useful to have on hand along with cached file stats.

PR #238, opened 2024-06-03
Bugfix for recent library re-creating a library at the last library location.

### LINKED ISSUE #212: TagStudio opens recent library even if it has been deleted
Simply, when the most recent library opened has its .TagStudio folder deleted, the program assumes that it still exists when re-opening it on launch (due to the Window > Open library on start toggle being enabled).

Probably all that would be needed is a check to confirm it is still on disk.

Loran425:
If the latest opened library no longer contains a `TS_FOLDER_NAME` (currently `.TagStudio`) folder clear the recent library and revert to default behavior for opening TagStudio. 

NOTE: if the library moves to a different architecture (e.g. living alongside preference instead of alongside the files this check will need to be removed)

Fixes #212

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -525,6 +525,14 @@ class QtDriver(QObject):
         elif self.settings.value(SettingItems.START_LOAD_LAST, True, type=bool):
             lib = self.settings.value(SettingItems.LAST_LIBRARY)
 
+            # TODO: Remove this check if the library is no longer saved with files
+            if not (Path(lib) / TS_FOLDER_NAME).exists():
+                logging.error(
+                    f"[QT DRIVER] {TS_FOLDER_NAME} folder in {lib} does not exist."
+                )
+                self.settings.setValue(SettingItems.LAST_LIBRARY, "")
+                lib = None
+
         if lib:
             self.splash.showMessage(
                 f'Opening Library "{lib}"...',
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -526,7 +526,7 @@ class QtDriver(QObject):
             lib = self.settings.value(SettingItems.LAST_LIBRARY)
 
             # TODO: Remove this check if the library is no longer saved with files
-            if not (Path(lib) / TS_FOLDER_NAME).exists():
+            if lib and not (Path(lib) / TS_FOLDER_NAME).exists():
                 logging.error(
                     f"[QT DRIVER] {TS_FOLDER_NAME} folder in {lib} does not exist."
                 )
--- a/tagstudio/src/core/enums.py
+++ b/tagstudio/src/core/enums.py
@@ -16,3 +16,4 @@ class Theme(str, enum.Enum):
     COLOR_HOVER = "#65AAAAAA"
     COLOR_PRESSED = "#65EEEEEE"
     COLOR_DISABLED = "#65F39CAA"
+    COLOR_DISABLED_BG = "#65440D12"
--- a/tagstudio/src/core/enums.py
+++ b/tagstudio/src/core/enums.py
@@ -15,3 +15,4 @@ class Theme(str, enum.Enum):
     COLOR_BG = "#65000000"
     COLOR_HOVER = "#65AAAAAA"
     COLOR_PRESSED = "#65EEEEEE"
+    COLOR_DISABLED = "#65F39CAA"
--- a/tagstudio/src/qt/widgets/preview_panel.py
+++ b/tagstudio/src/qt/widgets/preview_panel.py
@@ -31,7 +31,7 @@ from humanfriendly import format_size
 
 from src.core.enums import SettingItems, Theme
 from src.core.library import Entry, ItemType, Library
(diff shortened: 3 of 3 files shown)


Loran425:
Another check should probably be done when populating the recent libraries in the preview panel so that broken libraries either don't appear or appear in a different color (an unavailable network share doesn't get deleted but gets crossed out and becomes unclickable or clicking does a check for availability)

CyanVoxel:
> Another check should probably be done when populating the recent libraries in the preview panel so that broken libraries either don't appear or appear in a different color (an unavailable network share doesn't get deleted but gets crossed out and becomes unclickable or clicking does a check for availability)

The following changes to `def _fill_libs_widget()` on line 264 `preview_panel.py` will check if the locations are valid, then disable the buttons if they aren't:
```py
def _fill_libs_widget(
        self, libraries: list[tuple[str, tuple[str, str]]], layout: QVBoxLayout
    ):
        def clear_layout(layout_item: QVBoxLayout):
            for i in reversed(range(layout_item.count())):
                child = layout_item.itemAt(i)
                if child.widget() is not None:
                    child.widget().deleteLater()
                elif child.layout() is not None:
                    clear_layout(child.layout())  # type: ignore

        # remove any potential previous items
        clear_layout(layout)

        label = QLabel("Recent Libraries")
        label.setAlignment(Qt.AlignCenter)  # type: ignore

        row_layout = QHBoxLayout()
        row_layout.addWidget(label)
        layout.addLayout(row_layout)

        def set_button_style(
            btn: QPushButton, extras: list[str] | None = None, enabled: bool = True
        ):
            base_style = [
                f"background-color:{Theme.COLOR_BG.value};",
                "border-radius:6px;",
                "text-align: left;",
                "padding-top: 3px;",
                "padding-left: 6px;",
                "padding-bottom: 4px;",
            ]

            full_style_rows = base_style + (extras or [])

            btn.setStyleSheet(
                (
                    "QPushButton{"
                    f"{''.join(full_style_rows)}"
                    "}"
                    f"QPushButton::hover{{background-color:{Theme.COLOR_HOVER.value};}}"
                    f"QPushButton::pressed{{background-color:{Theme.COLOR_PRESSED.value};}}"
                )
            )
            btn.setCursor(Qt.CursorShape.PointingHandCursor)

        for item_key, (full_val, cut_val) in libraries:
            button = QPushButton(text=cut_val)
            button.setObjectName(f"path{item_key}")
            disabled_style: list[str] = [
                f"color:{get_tag_color(ColorType.LIGHT_ACCENT, "Red")};",
                f"background-color:{get_tag_color(ColorType.DARK_ACCENT, "Red")};",
            ]

            if Path(full_val).exists():

                def open_library_button_clicked(path):
                    return lambda: self.driver.open_library(Path(path))

                button.clicked.connect(open_library_button_clicked(full_val))
                set_button_style(button)
            else:
                button.setEnabled(False)
                button.setToolTip("Location no longer exists.")
                set_button_style(button, disabled_style)

            button_remove = QPushButton("➖")
            button_remove.setCursor(Qt.CursorShape.PointingHandCursor)
            button_remove.setFixedWidth(30)
            set_button_style(button_remove)

            def remove_recent_library_clicked(key: str):
                return lambda: (
                    self.driver.remove_recent_library(key),
                    self.fill_libs_widget(self.libs_layout),
                )

            button_remove.clicked.connect(remove_recent_library_clicked(item_key))

            row_layout = QHBoxLayout()
            row_layout.addWidget(button)
            row_layout.addWidget(button_remove)

            layout.addLayout(row_layout)
```
image

tagstudio/src/qt/ts_qt.py
```diff
@@ -525,6 +525,14 @@ def start(self) -> None:
...
             lib = self.settings.value(SettingItems.LAST_LIBRARY)
 
+            # TODO: Remove this check if the library is no longer saved with files
+            if not (Path(lib) / TS_FOLDER_NAME).exists():
+                logging.error(
+                    f"[QT DRIVER] {TS_FOLDER_NAME} folder in {lib} does not exist."
+                )
+                self.settings.setValue(SettingItems.LAST_LIBRARY, "")
+                lib = None
+
```
CyanVoxel, line 535: If `LAST_LIBRARY` is absent from the config while `START_LOAD_LAST` remains (a situation I encountered while testing) then this block soft crashes if it doesn't verify that `lib` received a value from the config.
```suggestion
            if lib:
                # TODO: Remove this check if the library is no longer saved with files
                if not (Path(lib) / TS_FOLDER_NAME).exists():
                    logging.error(
                        f"[QT DRIVER] {TS_FOLDER_NAME} folder in {lib} does not exist."
                    )
                    self.settings.setValue(SettingItems.LAST_LIBRARY, "")
                    lib = None
```

Loran425: Ahh yeah I didn't think about that case. change implemented.

Loran425:
Went a slightly different direction on the stylesheet just giving those buttons a disabled style and adding the color to the theme instead of importing the get_tag_color function.

Loran425:
but now I've got an issue where Path("") is evaluating to Path(".") which is obviously not intended behavior.

PR #244, opened 2024-06-03
Separately pin QT nixpkg version

seakrueger:
By pinning QT separately, the rest of dependent nixpkgs can be updated without being affected by the current QT version in nixpkgs unstable. The nixpkg revision is also bumped, fixing an issue where vim would immediately segfault on launch in the nix dev shell. This is for QT version 6.6.3, for when #223 is merged, the nixpkg commit tied to QT will need to be bumped to commit 47da0aee5616a063015f10ea593688646f2377e4. 

Also adds missing dependecies from the new video player.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1712473363,
-        "narHash": "sha256-TIScFAVdI2yuybMxxNjC4YZ/j++c64wwuKbpnZnGiyU=",
+        "lastModified": 1716948383,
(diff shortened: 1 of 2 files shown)


xarvex:
Meant to respond to this earlier but I have been busy on vacation. I'm currently almost up to a stable NixOS setup so I can actually test this, on a glance though this seems great!
Will follow up as soon as I have more.

xarvex:
Confirmed working on NixOS, able to manipulate and create libraries. This is currently hardcoded to x86_64 on Linux, but so was the original flake, and is out of the scope of just making this work. I will make my own PR to address that (and I also don't want to mess up #229).
On the topic of #229, I see it is based on this PR. Is it just in a draft until this gets used?

CC @CyanVoxel - ready to be merged

PR #245, opened 2024-06-04
Fix Default Ignored File Extension

Loran425:
Adds leading `.` to default extensions
Adds item delegate for Ignored File Extension to add leading `.` if left off extension


Loran425:
Worth noting that this does allow for multi suffix extension (e.g. <file>.tar.gz) and saves them properly but multi suffixes are not parsed in other locations at this time. the example file shows up as a `.gz` file and will still appear in the library.

PR #247, opened 2024-06-04
fix(QtDriver): redundant search override

career-yashaswee:
fix(QtDriver): redundant search override

**Description:**

This pull request addresses an issue where the `filter_items` function was being called unnecessarily in the QtDriver class. This function filters assets based on a search term.

**Expected Behavior:**

When the user enters a search term and clicks the search button or presses Enter:
If the search term hasn't changed (i.e., it's the same as the previous search), `filter_items` shouldn't be called again.
This check should be bypassed during window initialisation to ensure all assets are initially visible.

**Previous Behavior:**
`filter_items` was called regardless of whether the search term had changed, leading to redundant filtering.

**Changes Introduced:**
A new variable, previous_search_string, is introduced.
When the user performs a search:previous_search_string is set to the current search term.

Before calling `filter_items`:
The code checks if previous_search_string is the same as the current search term.
If they are the same and the window is not initializing, `filter_items`is skipped.
An override flag is added for functions that call `filter_items` during window initialisation to bypass the check and ensure all assets are displayed.

**Benefits:**
- Improves performance by avoiding unnecessary filtering.
- Enhances user experience by preventing redundant filtering actions.

**Testing:**
Manual testing has been conducted to ensure the fix functions as intended.
- Additional unit tests can be added to further solidify the fix.

Please review the code changes and provide feedback.

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -162,7 +162,7 @@ class Consumer(QThread):
 
 class QtDriver(QObject):
     """A Qt GUI frontend driver for TagStudio."""
-
+    last_search_string: None = None | str,
     SIGTERM = Signal()
 
     preview_panel: PreviewPanel
@@ -558,11 +558,11 @@ class QtDriver(QObject):
 
         search_button: QPushButton = self.main_window.searchButton
         search_button.clicked.connect(
-            lambda: self.filter_items(self.main_window.searchField.text())
+            lambda: self.filter_items(self.main_window.searchField.text(),False) # False as a parameter cancels default 'redundant search' override in filter_items()
         )
         search_field: QLineEdit = self.main_window.searchField
         search_field.returnPressed.connect(
-            lambda: self.filter_items(self.main_window.searchField.text())
+            lambda: self.filter_items(self.main_window.searchField.text(),False) # False as a parameter cancels default 'redundant search' override in filter_items()
         )
 
         back_button: QPushButton = self.main_window.backButton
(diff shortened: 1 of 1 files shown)


tagstudio/src/qt/ts_qt.py
```diff
@@ -1323,57 +1323,60 @@ def get_frame_contents(self, index=0, query: str = ""):
...
-                # self.nav_forward(self.lib.search_library(query))
-                self.nav_forward(*self.get_frame_contents(0, query))
-            self.cur_query = query
-
-            end_time = time.time()
-            if query:
+    
+    def filter_items(self, query: str = "", override: bool = True):
+        if self.lib and (override or (not (self.last_search_string == query))):
+            if(override or (not (self.last_search_string == query))):
```
YoyloNerd, line 1329: This second if statement can be removed since its already checked above

YoyloNerd: And self.last_search_string should be replaced with self.cur_query since it already keeps track of the same thing and is used below

career-yashaswee: Sure, done that!

tagstudio/src/qt/ts_qt.py
```diff
@@ -1331,58 +1331,60 @@ def get_frame_contents(self, index=0, query: str = ""):
             index,
             len(self.frame_dict[query]),
         )
-
-    def filter_items(self, query: str = ""):
+    
+    def filter_items(self, query: str = "", override: bool = True):
         if self.lib:
```
yedpodtrzitko, line 1336: While you're touching the code already, it could be worth to flip this condition as well. There's no `else:` block, so instead of this:
```
if self.lib:
    # all the code
```

it could be done as this:

```
if not self.lib:
    return

# the same code except it's less indented now and you dont have to scroll 50 lines to see what will happen (spoiler: nothing) if the condition `if self.lib:` is not fulfilled 
```

YoyloNerd:
You also fully broke some stuff from the main branch when merging that affects search

CyanVoxel:
Closing due to inactivity

Issue #230, opened 2024-05-31
[Suggestion] Add option to invert "Ignored File Extensions" into a whitelist


enzomtpYT:
Hello, while using the app i noticed there was an option for Excluding some File Extensions, but i didn't find anything to "invert" that list into a white list that only Allows some File Extensions. That's why im suggesting it as a kind of tickbox.

CyanVoxel:
Implemented in #251

PR #251, opened 2024-06-05
Add option to use a allowed extensions instead of ignored extensions

Thesacraft:
I added the option to invert the ignored extension list like its mentioned in #230.

### DIFF
--- a/tagstudio/src/core/json_typing.py
+++ b/tagstudio/src/core/json_typing.py
@@ -9,6 +9,7 @@ class JsonLibary(TypedDict("", {"ts-version": str})):
     macros: "list[JsonMacro]"
     entries: "list[JsonEntry]"
     ignored_extensions: list[str]
+    ignore_extensions: bool
 
 
 class JsonBase(TypedDict):
--- a/tagstudio/src/core/json_typing.py
+++ b/tagstudio/src/core/json_typing.py
@@ -8,8 +8,8 @@ class JsonLibary(TypedDict("", {"ts-version": str})):
     fields: list  # TODO
     macros: "list[JsonMacro]"
     entries: "list[JsonEntry]"
-    ignored_extensions: list[str]
-    ignore_extensions: bool
+    ext_list: list[str]
+    is_exclude_list: bool
 
 
 class JsonBase(TypedDict):
--- a/tagstudio/src/core/json_typing.py
+++ b/tagstudio/src/core/json_typing.py
@@ -1,4 +1,4 @@
-from typing import TypedDict
+from typing import TypedDict, NotRequired
 
 
 class JsonLibary(TypedDict("", {"ts-version": str})):
@@ -10,7 +10,7 @@ class JsonLibary(TypedDict("", {"ts-version": str})):
     entries: "list[JsonEntry]"
     ext_list: list[str]
     is_exclude_list: bool
-    ignored_extensions: NotRequired[list[str]] # deprecated
+    ignored_extensions: NotRequired[list[str]]  # deprecated
 
 
 class JsonBase(TypedDict):
--- a/tagstudio/src/core/json_typing.py
+++ b/tagstudio/src/core/json_typing.py
@@ -1,4 +1,5 @@
-from typing import TypedDict, NotRequired
+from typing import TypedDict
+from typing_extensions import NotRequired
 
 
 class JsonLibary(TypedDict("", {"ts-version": str})):
--- a/tagstudio/src/core/json_typing.py
+++ b/tagstudio/src/core/json_typing.py
@@ -10,7 +10,7 @@ class JsonLibary(TypedDict("", {"ts-version": str})):
     entries: "list[JsonEntry]"
     ext_list: list[str]
     is_exclude_list: bool
-    ignored_extensions: list[str]
+    ignored_extensions: NotRequired[list[str]] # deprecated
 
 
 class JsonBase(TypedDict):
--- a/tagstudio/src/core/json_typing.py
+++ b/tagstudio/src/core/json_typing.py
@@ -10,6 +10,7 @@ class JsonLibary(TypedDict("", {"ts-version": str})):
     entries: "list[JsonEntry]"
     ext_list: list[str]
     is_exclude_list: bool
+    ignored_extensions: list[str]
 
 
 class JsonBase(TypedDict):
--- a/tagstudio/src/core/library.py
+++ b/tagstudio/src/core/library.py
@@ -1523,7 +1523,7 @@ class Library:
         else:
             for entry in self.entries:
                 added = False
-                allowed_ext: bool = entry.filename.suffix not in self.ext_list
+                allowed_ext = entry.filename.suffix not in self.ext_list
                 if (allowed_ext and self.ignore_extensions) or (
                     not allowed_ext and not self.ignore_extensions
                 ):
--- a/tagstudio/src/core/library.py
+++ b/tagstudio/src/core/library.py
@@ -871,10 +871,7 @@ class Library:
                     if f.suffix not in self.ext_list and self.is_exclude_list:
                         self.dir_file_count += 1
                         file = f.relative_to(self.library_dir)
-                        try:
-                            _ = self.filename_to_entry_id_map[file]
-                        except KeyError:
-                            # print(file)
+                        if file not in self.filename_to_entry_id_map:
                             self.files_not_in_library.append(file)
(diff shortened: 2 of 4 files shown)


tagstudio/src/core/library.py
```diff
@@ -343,6 +343,7 @@ def __init__(self) -> None:
         # A list of file extensions to be ignored by TagStudio.
         self.default_ext_blacklist: list = [".json", ".xmp", ".aae"]
         self.ignored_extensions: list = self.default_ext_blacklist
```
Thesacraft, line 345: I don't know if i should rename this variable and if i should i don't know what the new name should be.

CyanVoxel: `default_ext_blacklist` should be fine since it's still true: it *is* the *default* extension *blacklist*.

`ignored_extensions` could be changed to something like "`ext_list`" since its purpose is now more generic, plus it brings the naming in line with `default_ext_blacklist`.

Thesacraft: Yep agree with that and have added it

Thesacraft:
I don't understand why mypy complain about the line 1526 in my PR because i didn't change it ( or more accurate removed the change i did and it still complains).

CyanVoxel:
> I don't understand why mypy complain about the line 1526 in my PR because i didn't change it ( or more accurate removed the change i did and it still complains).

I ran into a similar issue recently. You could create the definition of `allowed_ext` outside of the `for` loop with the type hint and a default value, then change the two existing definitions/"redefinitions" to just be assignments without the type hints.

tagstudio/src/core/library.py
```diff
@@ -501,8 +502,9 @@ def open_library(self, path: str | Path) -> int:
 
                     # Load Extension Blacklist ---------------------------------
                     if "ignored_extensions" in json_dump.keys():
-                        self.ignored_extensions = json_dump["ignored_extensions"]
-
+                        self.ext_list = json_dump["ignored_extensions"]
+                    if "ignore_extensions" in json_dump.keys():
```
yedpodtrzitko, line 506: This variable name is very similar to the variable `ignored_extensions`, it would be better to use a different name which can be easily told apart.

btw checking for item's presence in dictionary in ...uh ...Pythonic (I hate that word) way is done this way:
`if "ignore_extensions" in json_dump:` instead of:
`if "ignore_extensions" in json_dump.keys():`

Thesacraft: Hi, thanks for the review. I agree that ignore_extensions is not the best name for the variable do you think use_as_blacklist would be better? I also think i am gonna rename the ignored_extensions in the json file to ext_list. 

The if statment i will change. I will also change the other ones that use the same syntax ( they are in the next lines )

Thesacraft: > There is some recent movement to avoid terminologies like master/slave or whitelist/blacklist, so I would avoid this term.
Oh makes sense sorry didn't think about that.

tagstudio/src/core/library.py
```diff
@@ -1352,12 +1364,14 @@ def search_library(
...
-                allowed_ext: bool = entry.filename.suffix not in self.ignored_extensions
+                allowed_ext: bool = entry.filename.suffix not in self.ext_list
                 # try:
                 # entry: Entry = self.entries[self.file_to_library_index_map[self._source_filenames[i]]]
                 # print(f'{entry}')
 
-                if allowed_ext:
+                if (allowed_ext and self.ignore_extensions) or (
+                    not allowed_ext and not self.ignore_extensions
+                ):
```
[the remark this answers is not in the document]
Thesacraft: Yep works fine (changed it here and the other occurence). Thanks!

tagstudio/src/core/library.py
Thesacraft: what do you think of `use_as_denylist` ?

CyanVoxel: I think "include list" and "exclude list" would work well here

Thesacraft: yep sound even better.

tagstudio/src/core/library.py
```diff
@@ -499,10 +501,14 @@ def open_library(self, path: str | Path) -> int:
...
                     major, minor, patch = json_dump["ts-version"].split(".")
 
-                    # Load Extension Blacklist ---------------------------------
-                    if "ignored_extensions" in json_dump.keys():
-                        self.ignored_extensions = json_dump["ignored_extensions"]
-
+                    # Load Extension Exclude List ---------------------------------
+                    # The following if statment is only used when a old library is loaded
+                    if "ignored_extensions" in json_dump:
+                        self.ext_list = json_dump["ignored_extensions"]
```
Thesacraft, line 507: This line is why mypy complains but it keeps the ignored_extensions and saves them as the ext_list to update old librarys.
Should i remove this line and break it or does someone know a way arround it?

yedpodtrzitko: keep `ignored_extensions` in `JsonLibary` (sic!) for backward compatibility and change the loading here to:
```
self.ext_list = json_dump.get("ext_list") or json_dump.get("ignored_extensions")
```
so it will first try to use `ext_list` and in case it doesnt exist yet (old library), it will try to use `ignored_extensions`

tagstudio/src/core/library.py
```diff
@@ -735,17 +740,20 @@ def to_json(self):
...
             "ts-version": VERSION,
-            "ignored_extensions": [],
+            "ext_list": [],
+            "is_exclude_list": True,
             "tags": [],
             "collations": [],
             "fields": [],
             "macros": [],
             "entries": [],
+            "ignored_extensions": [],
```
Thesacraft, line 750: This fixes the mypy error.
But it also saves an empty list in the json file, which I wanted to avoid.

tagstudio/src/core/library.py
```diff
@@ -735,17 +740,20 @@ def to_json(self):
...
         }
 
         print("[LIBRARY] Formatting Tags to JSON...")
 
-        file_to_save["ignored_extensions"] = [i for i in self.ignored_extensions if i]
+        file_to_save["ext_list"] = [i for i in self.ext_list if i]
+        file_to_save["is_exclude_list"] = self.is_exclude_list
```
yedpodtrzitko, line 756: you can do these assignment above on lines 743-744 instead of having there the dummy values

tagstudio/src/core/json_typing.py
```diff
@@ -8,6 +8,8 @@ class JsonLibary(TypedDict("", {"ts-version": str})):
     fields: list  # TODO
     macros: "list[JsonMacro]"
     entries: "list[JsonEntry]"
+    ext_list: list[str]
+    is_exclude_list: bool
     ignored_extensions: list[str]
```
yedpodtrzitko, line 13:
```suggestion
    ignored_extensions: NotRequired[list[str]] # deprecated
```
\+ import:
`from typing import NotRequired`

tagstudio/src/core/library.py
yedpodtrzitko: if you will [mark it ](https://github.com/TagStudioDev/TagStudio/pull/251/files#r1629367462)via `NotRequired`, you dont have to set it here
```suggestion

```

tagstudio/src/core/json_typing.py
Thesacraft: Thanks!

tagstudio/src/core/library.py
```diff
@@ -864,7 +868,15 @@ def refresh_dir(self) -> Generator:
...
                 ):
-                    if f.suffix not in self.ignored_extensions:
+                    if f.suffix not in self.ext_list and self.is_exclude_list:
+                        self.dir_file_count += 1
+                        file = f.relative_to(self.library_dir)
+                        try:
+                            _ = self.filename_to_entry_id_map[file]
+                        except KeyError:
+                            # print(file)
+                            self.files_not_in_library.append(file)
```
yedpodtrzitko, line 878: last nitpick, sorry. I know this is copy of the existing code, but let's make a bit better:
```suggestion
                        if file not in self.filename_to_entry_id_map:
                            self.files_not_in_library.append(file)
```

Thesacraft: All good.

CyanVoxel:
Pushed a few changes, notably switching the exclude/include mode selection from a checkbox to a combobox for better intuition of its function as well as some changes to how the extension lists are loaded. 
image

Thesacraft:
That is so much better haha

PR #259, opened 2024-06-07
Add "Clear Selection" button

PencilVoid:
Simply lets you clear the current entry selection through the Edit -> Clear Selection button, or by pressing Ctrl+D. A minor feature I think would be very convenient.

I am quite unfamiliar with both Python and Qt, so please let me know if there is anything I should change or improve.

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -403,13 +403,8 @@ class QtDriver(QObject):
 
         clear_select_action = QAction("Clear Selection", menu_bar)
         clear_select_action.triggered.connect(self.clear_select_action_callback)
-        clear_select_action.setShortcut(
-            QtCore.QKeyCombination(
-                QtCore.Qt.KeyboardModifier(QtCore.Qt.KeyboardModifier.ControlModifier),
-                QtCore.Qt.Key.Key_D,
-            )
-        )
-        clear_select_action.setToolTip("Ctrl+D")
+        clear_select_action.setShortcut(QtCore.Qt.Key.Key_Escape)
+        clear_select_action.setToolTip("Esc")
         edit_menu.addAction(clear_select_action)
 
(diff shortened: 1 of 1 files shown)


tagstudio/src/qt/ts_qt.py
```diff
@@ -401,6 +401,17 @@ def start(self) -> None:
...
 
+        clear_select_action = QAction("Clear Selection", menu_bar)
+        clear_select_action.triggered.connect(self.clear_select_action_callback)
+        clear_select_action.setShortcut(
+            QtCore.QKeyCombination(
+                QtCore.Qt.KeyboardModifier(QtCore.Qt.KeyboardModifier.ControlModifier),
+                QtCore.Qt.Key.Key_D,
+            )
+        )
+        clear_select_action.setToolTip("Ctrl+D")
```
CyanVoxel, line 412: It would make more sense to me if the key shortcut was <kbd>Esc</kbd> instead of <kbd>Ctrl</kbd> + <kbd>D</kbd>
```suggestion
        clear_select_action.setShortcut(QtCore.Qt.Key.Key_Escape)
        clear_select_action.setToolTip("Esc")
```

CyanVoxel (changes requested):
Looks pretty good! The ability to clear the grid selection is a much-needed QoL feature. My only real gripe is the key command choice, which I've suggested a change for.

Besides that, it *would* be nice if both this and the "Select All" options were in their own "Select" menu, however the menu bar is in need of it's own reorganizing and cleanup anyway so I'd say that's out of the scope of this PR.
image

tagstudio/src/qt/ts_qt.py
PencilVoid: <kbd>Esc</kbd> does seem more intuitive. I've changed the shortcut accordingly.

CyanVoxel (approved):
Awesome, thanks for your work on this!

PR #260, opened 2024-06-07
Allow Tags to be deselected in search 

seakrueger:
Make buttons in TagSearch panel togglable to allow the selected tags to be deselected without having to close the panel. The standard look of the panel has had to been modified to support this behavior, now beginning with a light grey background and only matching the tag color when selected. When a tag is selected, the "+" text is only hidden because the ability to modify the buttons properties after the fact is lost, though I may have missed something obvious. The currently selected subtags for an entry are now passed into TagSearchPanel to allow the previously selected tags to be highlighted and deselect-able.

Workflow featuring new buttons:
output

### DIFF
--- a/tagstudio/src/qt/modals/tag_search.py
+++ b/tagstudio/src/qt/modals/tag_search.py
@@ -31,7 +31,7 @@ logging.basicConfig(format="%(message)s", level=logging.INFO)
 
 
 class TagSearchPanel(PanelWidget):
-    tag_chosen = Signal(int)
+    tag_chosen = Signal(int, bool)
 
     def __init__(self, library):
         super().__init__()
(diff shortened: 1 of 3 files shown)


tagstudio/src/qt/modals/tag_search.py
```diff
@@ -139,15 +144,27 @@ def update_tags(self, query: str = ""):
                 # f'padding-left: 4px;'
                 f"font-size: 20px;"
                 f"}}"
-                f"QPushButton::hover"
-                f"{{"
+                f"QPushButton::checked{{"
+                f"border-color:{get_tag_color(ColorType.BORDER, self.lib.get_tag(tag_id).color)};"
+                f"color: {get_tag_color(ColorType.PRIMARY, self.lib.get_tag(tag_id).color)};"
+                f"background: {get_tag_color(ColorType.PRIMARY, self.lib.get_tag(tag_id).color)};"
```
yedpodtrzitko, line 150: nitpick: 
`tag_color = self.lib.get_tag(tag_id).color` on the first line of this `for` loop, and then reuse the value here instead of calling `get_tag()` repeatedly?
```suggestion
                f"border-color:{get_tag_color(ColorType.BORDER, tag_color)};"
                f"color: {get_tag_color(ColorType.PRIMARY, tag_color)};"
                f"background: {get_tag_color(ColorType.PRIMARY, tag_color)};"
```

plus the other occurrences around here.

tagstudio/src/qt/modals/tag_search.py
```diff
@@ -115,19 +119,23 @@ def update_tags(self, query: str = ""):
...
+            tag_colors = self.lib.get_tag(tag_id).color
             ab = QPushButton()
+            ab.setCheckable(True)
             ab.setMinimumSize(23, 23)
             ab.setMaximumSize(23, 23)
             ab.setText("+")
             ab.setStyleSheet(
                 f"QPushButton{{"
-                f"background: {get_tag_color(ColorType.PRIMARY, self.lib.get_tag(tag_id).color)};"
+                f"background: #d2d2d2;"
```
eivl, line 133: maybe use a sensible name instead of the magic value?
`LIGHT_GRAY` as an example.

eivl (approved):
Approved when comment is addressed.

seakrueger:
Out of date

PR #262, opened 2024-06-08
Add the ability to create tags when adding a tag to a file

### LINKED ISSUE #1: [SUGGESTION] Add Tag button in "Search Tag" window when Tag cannot be found by the app
example image

Untitled-1

bjorn-out:
Closes #1 

- Adds a button to the "add tag" modal, below the search results, to create a new tag with the search query as its name, and immediately adds it to the file.
  - It will be a simple tag with no subtags or aliases, and a random color will be assigned.
  - If there are no search results, hitting enter will also create a tag.
- Search box is now cleared when the modal opens.

Ability to change further details for the newly created tag (like subtags) should be added. My suggestion would be to allow editing any tag in the search results by right-clicking. I tried to implement this, but did not get it to work yet. 

I do believe the full tag creation modal should not pop-up by default, as this is frankly annoying when trying to tag many files manually. You can already use Ctrl + T or Cmd + T if you want to create a tag with more details, or use the aforementioned edit button once implemented.


bjorn-out:
Implements #1

tagstudio/src/qt/modals/tag_search.py
```diff
@@ -153,6 +157,73 @@ def update_tags(self, query: str = ""):
...
+        create_button.clicked.connect(lambda x=query: self.create_tag(query))
+        l.addWidget(create_button)
+
+        return c
+
+    def create_tag(self, query: str):
+        # Fuction called when tag should be created based on query
+        # Creates the tag and applies it to the document
+        new_tag: Tag = Tag(
+            id=-2,
```
yedpodtrzitko, line 212: wont this cause a collision among multiple tags created this way? Or is the ID somehow reassigned to different value elsewhere in the code?

tagstudio/src/qt/modals/tag_search.py
```diff
@@ -153,6 +157,73 @@ def update_tags(self, query: str = ""):
...
+
+        create_button.setStyleSheet(
+            f"QPushButton{{"
+            f"background: {get_tag_color(ColorType.PRIMARY, "dark gray")};"
+            f"color: {get_tag_color(ColorType.TEXT, "dark gray")};"
+            f"font-weight: 600;"
+            f"border-color:{get_tag_color(ColorType.BORDER, "dark gray")};"
+            f"border-radius: 6px;"
+            f"border-style:solid;"
+            f"border-width: {math.ceil(1*self.devicePixelRatio())}px;"
```
yedpodtrzitko, line 192: what's the `1*` multiplying for?

CyanVoxel:
I've made some changes and implemented the feedback given here as well as in #1.
Thank you for your existing work on this!

PR #263, opened 2024-06-09
feat: add all html color names

BlueFalconHD:


coolesding:
interesting pull request haha, this would definitely require a better way to select colours though, since scrolling through that endless list would be a nightmare

CyanVoxel:
While more color options is definitely something we want, I'm afraid this isn't the way to go about it. Not only does this change not function properly, but increasing the number of available colors in the dropdown from ~36 to ~746 would render it extremely difficult to use. Custom user colors is also a feature on the [Planned Features](https://github.com/TagStudioDev/TagStudio/blob/main/doc/updates/planned_features.md) list that would supersede the need for a comprehensive list of defaults. Also for the future, providing a description on the pull request of your intentions and changes would be very helpful.

BlueFalconHD:
Thank you, I apologize.

CyanVoxel:
I'll close out this particular PR but I'll be keeping issue #264 open as place to discuss the custom color feature in more detail. I appreciate the time you've put in creating this, and apologize for not being able to help guide the feature given the circumstances.

PR #273, opened 2024-06-10
Blender thumbnail support

050011-code:
Used code from Blender's source code for extracting thumbnail, modified it and added special entry for .blend in the thumbnail renderer, to then use those functions


tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -204,6 +204,40 @@ def render(
                 # 	img_buf = io.BytesIO()
                 # 	plt.savefig(img_buf, format='png')
                 # 	image = Image.open(img_buf)
+
+                # Blender ===========================================================
+                elif _filepath.suffix.lower() == '.blend':
+                    from src.qt.helpers import blender_thumbnailer
```
Thesacraft, line 210: I think this should be moved to the top of the file to the other imports.

Thesacraft: And i think the import could just import the main out of the blender_thumbnailer so it would be 
from src.qt.helpers.blender_thumbnailer import main

050011-code: Done!

CyanVoxel (changes requested):
Great work on this, it's super exciting to have blend support! 😁 

I've got some suggested changes commented above - mostly optimizations and cleanup. Also, I noticed you were running into issues with the image being flipped when it was exported from the `blend_extract_thumb` function. If you wanted, you could import PIL into that file and do the image flipping there, letting you export a PIL Image object instead of the `image_buffer, x, y` list. This would also clean up the image creation process on the `thumb_renderer` side of things.

CyanVoxel (approved):
Awesome work on this, thank you so much!! 😁

I'm marking this as ready to merge and will target the 9.4 branch. If any conflicts come up I'll fix those real quick.

PR #274, opened 2024-06-10
Refactor `video_player.py` (Fix #270)

### LINKED ISSUE #270: [BUG]: Video Playback Icons Missing in Executables
### TagStudio Version
9.3.0
### OS
Windows 10 22H2

### Description of the Issue
When hovering over a video preview, none of the on-screen control icons are displayed. I've confirmed this to be broken in both the portable and non-portable Windows builds at the very least.
image

### Additional Info
When using a custom build with a console window, I'm able to see that the icons aren't being found for some reason.
```
qt.svg: Cannot open file './tagstudio/resources/pause.svg', because: The system cannot find the path specified.
qt.svg: Cannot open file './tagstudio/resources/volume_muted.svg', because: The system cannot find the path specified.
```

CyanVoxel:
This PR makes various changes to `video_player.py`, with many involving where the SVG icons are stored and how they're loaded. In the process this fixes #270.
### Changes
- Move icons files to resources/qt/images folder, some being renamed
- Reduce icon loading to single initial import
- Tweak icon dimensions and animation timings
- Remove unnecessary commented code
- Remove unused/duplicate imports
- Add license info to file


Loran425:
Hmmmm, not seeing icons in the exe build but that might be something on my end.

Loran425:
I think its because its a local path so tagstudio is no longer valid after packaging, swapping to `_internal` breaks it when running from python but works when running from the build. 

I think we need to figure out how to incorporate that into the resources_rc file and then change the path to be similar to the way the splash screen is loaded.

CyanVoxel:
> I think its because its a local path so tagstudio is no longer valid after packaging, swapping to `_internal` breaks it when running from python but works when running from the build.
> 
> I think we need to figure out how to incorporate that into the resources_rc file and then change the path to be similar to the way the splash screen is loaded.

I wonder why these are the only assets breaking when all of the other assets that aren't in the resources_rc are still working... And I tested out building it on my end beforehand, at least the portable version, and it worked with this change

Loran425:
Hmmm does it work for you as is with `pyinstaller.exe .\tagstudio.spec` called from the root of your local repo?

If I had to guess it has something to do with that `Path(__file__).parents[3]` bit. I think the "right" way is the resources file, but because those files are in the same structure it works for them, but for this one it didn't reference the `__file__` so it didn't get updated from `tagstudio` to `_internal`

Loran425:
@CyanVoxel I somehow pulled an old version of this PR sorry for the confusion
This is working now.

CyanVoxel:
> @CyanVoxel I somehow pulled an old version of this PR sorry for the confusion This is working now.

Great news haha 😁

tagstudio/src/qt/widgets/video_player.py
```diff
@@ -51,6 +49,25 @@ class VideoPlayer(QGraphicsView):
...
+        pause_icon = bytes(icon.read())
+
+    with open(Path(_parents, "resources/qt/images/play.svg"), "rb") as icon:
+        play_icon = bytes(icon.read())
+
+    with open(Path(_parents, "resources/qt/images/volume.svg"), "rb") as icon:
+        volume_icon = bytes(icon.read())
+
+    with open(Path(_parents, "resources/qt/images/volume_mute.svg"), "rb") as icon:
+        volume_mute_icon = bytes(icon.read())
```
yedpodtrzitko, line 69: I'd keep these resources outside of the `VideoPlayer` class, and rather have another class for handling this kind of assets, `ResourceManager` or something like that. Then it can be used across the whole codebase for all the resources, which would be nicer than doing `open(Path(_parents, "resources/qt/something/something`))` everywhere. 
This will provide some level of abstraction, so if/when the underlying implementation will change, it wont affect any code using the resources.

Something like this:

```python
class ResourceManager:
    MAP = {
        "play_icon": "qt/images/play.svg",
        # ...
    }

    def _load(self, id: str):
        # eventually implement cache like
        # if id in self.CACHE: return self.CACHE[id]
        with open(self.MAP[id], 'rb') as f:
            return f.read()

    def get(self, id: str):
        """Allow to use the manager as:
        r = ResourceManager()
        r.get('play_icon')
        """
        return self._load(id)

    def __getattr__(self, id: str):
        """Allow to use the manager as:
        r = ResourceManager()
        r.play_icon
        """
        if id in self.MAP:
            return self._load(id)
        raise AttributeError(f'attribute {id} not found')
```

Loran425:
For the Qt side I'd be leaning toward utilizing the built-in [resource manager](https://doc.qt.io/qtforpython-6/overviews/resources.html)

This would let the resources be managed outside this file and hopefully avoid future path issues for them. though it does add a step to remember whenever we want to add an image asset, and these are technically hidden imports as they aren't exposed to python so much as a resource made available to Qt in the background by that `import src.qt.resources_rc` line in `ts_qt.py`. 

Adding these images to the resources.qrc file
```xml
<!-- TO COMPILE: pyside6-rcc resources.qrc -o resources_rc.py (in \tagstudio\src\qt)-->
<RCC>
    <qresource prefix="/">
        <file alias = "images/star_icon_empty_128.png">../../resources/qt/images/star_icon_empty_128.png</file>
		<file alias = "images/star_icon_filled_128.png">../../resources/qt/images/star_icon_filled_128.png</file>
        <file alias = "images/box_icon_empty_128.png">../../resources/qt/images/box_icon_empty_128.png</file>
		<file alias = "images/box_icon_filled_128.png">../../resources/qt/images/box_icon_filled_128.png</file>
        <!-- <file alias = "images/edit_icon_128.png">../../resources/qt/images/edit_icon_128.png</file> -->
        <!-- <file alias = "images/trash_icon_128.png">../../resources/qt/images/trash_icon_128.png</file> -->
        <!-- <file alias = "images/clipboard_icon_128.png">../../resources/qt/images/clipboard_icon_128.png</file> -->
        <file alias = "images/splash.png">../../resources/qt/images/splash.png</file>
        <file alias = "images/pause.svg">../../resources/qt/images/pause.svg</file>
        <file alias = "images/play.svg">../../resources/qt/images/play.svg</file>
        <file alias = "images/volume_muted.svg">../../resources/qt/images/volume_muted.svg</file>
        <file alias = "images/volume_unmuted.svg">../../resources/qt/images/volume_unmuted.svg</file>
    </qresource>
</RCC>
```

would let you only reference these images with a colon and their alias so the update controls function on L182 ends up like this. 

```python
    def updateControls(self) -> None:
        if self.player.audioOutput().isMuted():
            self.mute_button.load(":images/volume_muted.svg")
        else:
            self.mute_button.load(":images/volume_unmuted.svg")

        if self.player.isPlaying():
            self.play_pause.load(":images/pause.svg")
        else:
            self.play_pause.load(":images/play.svg")
```

CyanVoxel:
I've added a basic `ResourceManager` class that shoulders the work that the `video_player.py` was doing on it's own. For this PR I'm continuing to use a system separate from the QResource system since I've run into reliability inconsistinces with it in the past. With that being said this ResourceManger class provides a layer between the resources and the rest of the app, meaning that it can abstract the QResource system for future drop-in replacements. This abstraction can also aide in returning theme-specific resources without unnecessary intervention from the classes asking for them.

I've tested the resource changes in the executable build as well and have found no regressions with the resource fix.

tagstudio/src/qt/resource_manager.py
```diff
@@ -0,0 +1,66 @@
...
+            id (str): The name of the resource.
+
+        Returns:
+            Any: The resource if found, else None.
+        """
+        cached_res = ResourceManager._cache.get(id, None)
+        if cached_res:
+            return cached_res
+        else:
+            path, mode = ResourceManager._map.get(id, None)
```
yedpodtrzitko, line 48: this will raise exception in case the resource is not present in `._map`, because at that point the value you will get is `None`, and that cant be assigned into two variables `path, mode`:

```
>>> path, mode = None
  File "<stdin>", line 1, in <module>
TypeError: cannot unpack non-iterable NoneType object
```

storing the data as dictionary inside `_map` instead of extracted properties would prevent this issue.

btw. the default fallback value of `.get()` is `None`, so no need to define that explicitly as the second parameter of `.get()`

`assert dict().get('foo') is None`

tagstudio/src/qt/widgets/video_player.py
```diff
@@ -29,38 +30,37 @@
...
     from src.qt.ts_qt import QtDriver
 
 
 class VideoPlayer(QGraphicsView):
-    """A simple video player for the TagStudio application."""
+    """A basic video player."""
 
-    resolution = QSize(1280, 720)
-    hover_fix_timer = QTimer()
+    rm: ResourceManager = ResourceManager()
```
yedpodtrzitko, line 45: I'd initialize this class inside `QtDriver` instead of here, as this will be needed in other places in the codebase as well, and that way you'll have the same instance available everywhere. Since `driver` is passed here, then it would be used as `self.driver.rm.play_icon` instead of `VideoPlayer.rm.play_icon`

tagstudio/src/qt/resource_manager.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # Initialize _map
+    with open(
+        Path(__file__).parent / "resources.json", mode="r", encoding="utf-8"
+    ) as f:
+        json_map: dict = ujson.load(f)
+        for item in json_map.items():
+            _map[item[0]] = (item[1]["path"], item[1]["mode"])
+
+    logging.info(f"[ResourceManager] Resources Loaded: {_map}")
```
yedpodtrzitko, line 28: - this would be better to do in constructor, otherwise doing `import` of this class will have "side-effect" (ie. opening and loading from json file), similar way as doing `import src.qt.resources_rc` has side-effect of prefilling some QT resources

- when using `.items()`, you can work with it as two variables (key, value), so you dont have to unpack it via numeric index (`item[0]` `item[1]`) later

```diff
-for item in json_map.items():
-    _map[item[0]] = (item[1]["path"], item[1]["mode"])
+ for key, value in json_map.items():
+    _map[key] = (value["path"], value["mode"])
```

- to make this future-proof, I'd assign just the whole `value` instead of extracting specific keys in case there will be some other property in future. This will also mitigate a potential issue further on (will comment below)

```diff
-_map[key] = (value["path"], value["mode"])
+_map[key] = value
```

...and at that point you can do just this:
```diff
-json_map: dict = ujson.load(f)
-for item in json_map.items():
-   _map[item[0]] = (item[1]["path"], item[1]["mode"])
+_map = ujson.load(f)
```

yedpodtrzitko (commented):
this would sure work, but maybe there's some room for improvement

tagstudio/src/qt/resource_manager.py
```diff
@@ -0,0 +1,67 @@
...
+            return cached_res
+        else:
+            res: dict = ResourceManager._map.get(id)
+            if res.get("mode") in ["r", "rb"]:
+                with open(
+                    (Path(__file__).parents[2] / "resources" / res.get("path")),
+                    res.get("mode"),
+                ) as f:
+                    data = f.read()
+                    if res.get("mode") == ["rb"]:
```
yedpodtrzitko, line 55:
```suggestion
                    if res.get("mode") == "rb":
```

yedpodtrzitko (approved):

CyanVoxel:
Thanks for the review, @yedpodtrzitko !

Issue #269, opened 2024-06-09
[Feature Request] Delete files from within the program


ZephyrSnoww:
Just started using the software, and a major thing I use software like this for is for deleting unwanted files from my PC - adding an option in the right click menu (or otherwise) to delete the selected files, would be immensely helpful.

CyanVoxel:
Added in #282 and expanded upon in #409

PR #282, opened 2024-06-12
Delete file in TagStudio

PeterBouSaada:
Added the option to delete files in the right click context menu

closes https://github.com/TagStudioDev/TagStudio/issues/269


tagstudio/src/qt/helpers/file_deleter.py
```diff
@@ -0,0 +1,40 @@
+import logging
+import os.path
```
yedpodtrzitko, line 2: please try to use `pathlib` instead of `os.path`, see [CONTRIBUTING.md#implementations](https://github.com/TagStudioDev/TagStudio/blob/main/CONTRIBUTING.md#implementations)

tagstudio/src/qt/helpers/file_deleter.py
```diff
@@ -0,0 +1,41 @@
...
+
+ERROR = f"[ERROR]"
+WARNING = f"[WARNING]"
+INFO = f"[INFO]"
+
+logging.basicConfig(format="%(message)s", level=logging.INFO)
+
+
+def delete_file(path: str | Path, callback: Callable):
+    _path = str(path)
```
yedpodtrzitko, line 17: isnt this conversion pointless when you convert it to `Path` on the next line?

tagstudio/src/qt/helpers/file_deleter.py
```diff
@@ -0,0 +1,41 @@
...
+    _file = Path(_path)
+    logging.info(f"Deleting file: {_path}")
+    if not _file.exists():
+        logging.error(f"File not found: {_path}")
+        return
+    try:
+        _file.unlink()
+        callback()
+    except:
+        traceback.print_exc()
```
yedpodtrzitko, line 27: this should be probably `logging.exception` instead

tagstudio/src/qt/helpers/file_deleter.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class FileDeleterHelper:
+    def __init__(self, filepath: str | Path):
+        self.filepath = str(filepath)
+
+    def set_filepath(self, filepath: str | Path):
+        self.filepath = str(filepath)
```
yedpodtrzitko, line 35: no reason to convert these to `str` when it's later passed to `delete_file()` which expects `Path`

tagstudio/src/qt/helpers/file_deleter.py
```diff
@@ -0,0 +1,41 @@
+import logging
+import pathlib
+import subprocess
+import sys
+import traceback
+from pathlib import Path
+from typing import Callable
```
yedpodtrzitko, line 7: `typing.Callable` is deprecated

```suggestion
from collections.abc import Callable
```

tagstudio/src/qt/helpers/file_deleter.py
```diff
@@ -0,0 +1,41 @@
+import logging
+import pathlib
+import subprocess
```
yedpodtrzitko, line 3: unused import?

tagstudio/src/qt/helpers/file_deleter.py
PeterBouSaada: I got this from https://github.com/TagStudioDev/TagStudio/blob/main/tagstudio/src/qt/helpers/file_opener.py#L90 for consistency. Should I switch both or this `file_deleter.py` only?

yedpodtrzitko: I see. That looks like some old code which would be good to eventually migrate to `logging` as well. 
Anyway, I dont have strong opinion about it / not a blocker. It just caught my attention as this is not the way how I'd expect exception should be logged.

PeterBouSaada: I'll swap `file_deleter.py` to use `logging.exception` only for now.

yedpodtrzitko (approved):

yedpodtrzitko (changes requested):
- pipeline not passing
- git conflict with `main` branch

yedpodtrzitko (approved):

CyanVoxel:
I apologize for the delay on getting to this PR. I was originally going to inform you of a one or two tweaks I'd like to have made to this before pulling and asking if it would be okay if I committed them directly myself, however I ended up realizing there were a few more extensions to this I'd like as well as discovering some edge cases that are requiring much more work than this PR probably bargained for.

My intention now is to accept and pull this into a separate branch where I'll finish working on these changes and additions instead of hijacking your PR. Thank you so much for the work you've already done here!

Issue #272, opened 2024-06-10
Add search by metadata


lunaro-4:
### Similar issues
This feature is considered to be one of "Priority Features" by readme, yet I did not find any opened issues on that topic. Although I found a pull request #190, which could make use of parsing field names from search query. 

### Suggestions
I did not check contributing rules beforehand, so I accidentally started working on issue beforehand. Here is snippet from my [fork](https://github.com/lunaro-4/TagStudio/tree/search):
``` python
    def parse_metadata(self, query: str | None = None) \
            -> list[dict[str, str | list[str]]]:
        """
        Splits query into several maps 'meta_key -> value'\n
        Values without specified key parsed as tags and put in 'unbound' key \n
        example query1:
            "meta_first: value; meta_second: value; tag1;\
 | meta_first: value; meta_second: value; notag;"
        example query2:
            "tag1 | notag | tag2"
        """
        if query is None:
            return {}
        meta_list: list = []
        meta_conditions = query.strip().split("|")
        for meta_condition in meta_conditions:
            meta_to_value: dict = {}
            field_data_list = meta_condition.strip().split(";")
            print(field_data_list)
            for field_data in field_data_list:
                field_parsed = field_data.strip().split(":")
                if len(field_parsed) < 2:
                    unbound_values = field_parsed[0].strip().split(' ')
                    if meta_to_value.get('unbound') is None:
                        meta_to_value['unbound'] = unbound_values
                    else:
                        meta_to_value['unbound'].append(unbound_values)
                    continue
                if len(field_parsed) != 2:
                    logging.warning("""[ERROR] Attempt to parse mutiple fields\
                                    as one! Do not forget to specify ';'\
                                    between different meta fields""")
                meta_to_value[field_parsed[0].lower()] = field_parsed[1].lower()
            meta_list.append(meta_to_value)

        logging.info("Parsed values: ",meta_list)
        return meta_list
``` 

### Current state
I have tried to not touch much of code in `Library.search_library`, only add few new functions to allow additional entries to be filtered, but all other features should have stayed the same.

I planned to open a pull request straight up, but it is explicitly specified new issue to be opened beforehand, so let me know if is ok to go forward with this

CyanVoxel:
Thank you for opening up an issue first, I really appreciate it!

This is definitely a much-needed feature. So if I'm understanding correctly, your approach is proposing syntax similar to the following?:
`title: bookname` Single search for "bookname" in Title field
[...]

PR #284, opened 2024-06-13
Add search by metadata

lunaro-4:
### Context
Related to issue #272 

### Features / Changes

 - Major refactoring over core.library.Library.search_library method
 - Add method for parsing keys for file metadata from search query
 - Add support for negative query
 
### Usage example

In a new version, specifying all tags with whitespace as usual will return result
from AND search type, regardless of one selected in UI:

`tag1 tag2` - return all entries, tagged with 'tag1' AND 'tag2'

If you want to use OR, you need to specify OR search type in UI, and 
provide tags, separated by '|' symbol

`tag1 | tag2` - (with OR search type) return all entries, tagged with 'tag1' OR 'tag2'

`tag1 tag2 | tag3` - (with OR search type) return all entries, tagged with 'tag1' and 'tag2' OR with 'tag3'

Search by non-tag metadata is also available by 'key: value' syntax:

`description: desc` - return all entries, which description contains 'desc'

`tag1 ; description: desc ` - return all entries, tagged with 'tag1' AND which description contains 'desc'

`description: desc | description: smth`:
 - (with OR search type) return all entries, which description contains either 'desc' or 'smth'
 - (with AND search type) return all entries, which description contains both 'desc' and 'smth'

Negative query can be set by specifying lookup value as 'null':

`tag1; description: null` - return all entries, tagged with 'tag1' and not having a description (or having a null description)

Original negative query tags like `empty` or `no author` also supported:

`no author tag1; description: desc` - return all entries without author, tagged with 'tag1' and description containing 'desc'

### Refactoring note

For ease of maintenance, many functions of original code were extracted to new methods in library.Library class. 
To avoid clutter, I extracted them even further to a new class 'library.Filter'. This should help maintain the codebase
in future, but due to lack of tests, I decided to extract to 'Filter' only new methods, and import any required variables 
and methods, to aviod breaking other functionality of 'library.Library' class:
``` python
class Filter:
    def __init__(self, lib: Library) -> None:
        # self.lib = lib
        self.ext_list = lib.ext_list
        self.entries = lib.entries
        self.is_exclude_list = lib.is_exclude_list
        self.get_field_attr = lib.get_field_attr
        self._tag_strings_to_id_map = lib._tag_strings_to_id_map
        self.get_tag_cluster = lib.get_tag_cluster
        self.library_dir = lib.library_dir
        self._field_id_to_name_map = lib._field_id_to_name_map
        self.get_field_obj = lib.get_field_obj
        self.missing_files = lib.missing_files
```

This class was added in last commit, so this can be reverted with ease.

### Other possible side effects

Changed code works fine at a first glance, but it is not clear for me how exactly
original code worked on some edge cases, especially with coalitions. 

I can write tests for it, but I would like to have test cases, for deeper understanding on how should it work


lunaro-4:
cherry-piked commit with tests
All github actions passed except for ruff, possibly because of files I didn't touch.
On my machine, I get no errors on files I edited:
``` bash
# at TagStudio/tagstudio
> ruff check --config ../pyproject.toml src/core/library.py
All checks passed!
>  ruff check --config ../pyproject.toml tests/core/
All checks passed!
```

PR is ready for review

yedpodtrzitko:
@lunaro-4 can you please rebase your branch? There are some changes which are already merged in `main` branch, eg `.github/workflows/pytest.yaml`

lunaro-4:
@yedpodtrzitko 
- [x] Rebase done
- [x] All checks passed

tagstudio/src/core/library.py
```diff
@@ -2324,3 +2172,287 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+            return list(result_set)
+        # Entries should match any of parts between '|'
+        elif search_mode == SearchMode.OR:
+            for result in pre_results:  # type: ignore [assignment]
+                for entry in result:  # type: ignore [attr-defined]
+                    result_set.add(entry)
+            return list(result_set)
+
+    def handle_tag_id(self, query: str, entry_tags: list[int]) -> bool:
+        id_query: int = 0
```
yedpodtrzitko, line 2299: this line looks unnecessary

tagstudio/src/core/library.py
```diff
@@ -2324,3 +2172,287 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+        try:
+            id_query = int(query)
+        except Exception:
+            return False
+        if id_query in entry_tags:
+            return True
+        return False
```
yedpodtrzitko, line 2306: just a nitpick
```suggestion
        return id_query in entry_tags
```

tagstudio/src/core/library.py
```diff
@@ -2324,3 +2172,287 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+    def handle_unbound(
+        self,
+        entry: Entry,
+        all_tag_terms: list[str],
+        entry_tags: list[int],
+        entry_authors: list[list[str]],
+        unbound: list[str],
+    ) -> tuple[ItemType, int] | None:
+        collations_added: list = []
+        orig_query = "".join(unbound)
```
yedpodtrzitko, line 2317: why is this converted to string? the conditions below would work even with list no? Also if you will add here this assert, you will see that some tests are passing here `unbound` as string instead of list. Which doesnt look correct according to the type annotation:

```suggestion
        assert isinstance(unbound, list), f'unbound is not list: {type(unbound)}'
        orig_query = "".join(unbound)
```

tagstudio/src/core/library.py
```diff
@@ -2324,3 +2172,287 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+            return None
+
+    def add_entries_from_special_flags(
+        self,
+        special_flags: tuple[bool, bool, bool, bool],
+        entry_tags: list[int],
+        entry_authors: list[list[str]],
+        entry: Entry,
+    ):
+        (only_untagged, only_no_author, only_empty, only_missing) = special_flags
```
yedpodtrzitko, line 2389: how about creating a class for encapsulating these flags instead of packing and unpacking them from tuple back and forth? Something like this:

```
class SearchFlags:
    only_untagged: bool = False
    only_no_author: bool = False
    only_empty: bool = False
    only_missing: bool = False

    def __init__(self, orig_query):
        # do the query evaluation here
        self.only_untagged = "no author" in orig_query or "no artist" in orig_query
        ...

flags = SearchFlags(orig_query)
```

tagstudio/src/core/library.py
```diff
@@ -2324,3 +2172,287 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+        if only_untagged and not entry_tags:
+            return True
+        elif only_no_author and not entry_authors:
+            return True
+        elif only_empty and (
+            len(entry.fields) <= 0 or len(entry.fields[0].keys()) <= 0
```
yedpodtrzitko, line 2395: I think the length can never be lower than 0
```suggestion
            not entry.fields or not entry.fields[0]
```

tagstudio/src/core/library.py
```diff
@@ -2324,3 +2172,287 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+            return True
+        return False
+
+    def remap_fields(self, entry: Entry) -> dict[str | int, str]:
+        entry_fields: dict = {}
+        for field in entry.fields:
+            if len(field.keys()) > 0:
+                field_keys = list(field.keys())
+                field_key = self._field_id_to_name_map.get(int(field_keys[0]))
+                entry_fields[field_key] = list(field.values())[0]
```
yedpodtrzitko, line 2412: This is breaking change. Currently each item can have the same field multiple times. I'm not sure if the design is intentional or accidental, @CyanVoxel? 

With this change you take only the first occurance.


identical code for this could be a bit simplifed
```suggestion
            for key, val in field.items():
                key_mapped = self._field_id_to_name_map.get(int(key))
                entry_fields[key_mapped] = val
```

tagstudio/src/core/library.py
```diff
@@ -2324,3 +2172,287 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+        return entry_fields
+
+    def populate_tags(self, entry: Entry) -> tuple[list[int], list[list[str]]]:
+        """Parse tags from entry.fields\n
+        returns tuple (entry_tags, entry_authors)"""
+        entry_tags: list[int] = []
+        entry_authors: list[list[str]] = []
+        if entry.fields:
+            for field in entry.fields:
+                if len(field.keys()) <= 0:
```
yedpodtrzitko, line 2422: nit
```suggestion
                if not field:
```

tagstudio/src/core/library.py
```diff
@@ -2324,3 +2172,287 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+
+            filtered_entries: list = []
+
+            # Iterate over all Entries =============================================================
+            for entry in self.entries:
+                entry_tuple: tuple | None = None
+                allowed_ext: bool = entry.filename.suffix not in self.ext_list
+                if allowed_ext != self.is_exclude_list:
+                    continue
+                (entry_tags, entry_authors) = self.populate_tags(entry)
```
yedpodtrzitko, line 2222: the tags and authors (among others) are extracted into `entry_fields` a few lines below via `self.remap_fields(entry)`, so this feels redundant.

tagstudio/src/core/library.py
```diff
@@ -2324,3 +2172,287 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+                field_id = list(field.keys())[0]
+                if self.get_field_obj(field_id)["type"] == "tag_box":
+                    entry_tags.extend(field[field_id])
+                if self.get_field_obj(field_id)["name"] == "Author":
+                    entry_authors.extend(field[field_id])
+                if self.get_field_obj(field_id)["name"] == "Artist":
+                    entry_authors.extend(field[field_id])
+        return (entry_tags, entry_authors)
+
+    def check_filename(self, entry: Entry, query: str | None) -> bool:
```
yedpodtrzitko, line 2433: I dont understand what does this method do

yedpodtrzitko (changes requested):
Nice work overall, however I dont think it works as expected. Or it works different way than I'd expect. Have this scenario:

- Entry 1:
  - author: foo
  - tag: bar

- Entry 2:
  - author: foo


now I do this query:
`author:foo | bar`

I would expect it will return me all items which have `author:foo` OR tag `bar`. But it returns only the Entry 1.


Also I it would be nice if I could be explicit that I search for `tag`, ie. query `tag:bar` would behave the same as query `bar`

yedpodtrzitko:
for the negative syntax it would be nice if it work the same way as with google, ie. if I want to exclude something, I could write `-author:foo` instead of `not author:foo`. But that's just wishful thinking, the decision is not up to me.

tagstudio/src/core/library.py
```diff
@@ -1320,6 +1319,47 @@ def get_entry_id_from_filepath(self, filename: Path):
...
+        if query is None:
+            return []
+        meta_list: list = []
+        meta_conditions = query.strip().split("|")
+        for meta_condition in meta_conditions:
+            query_part: dict = {}
+            field_data_list = meta_condition.strip().split(";")
+            for field_data in field_data_list:
+                field_parsed = field_data.strip().split(":")
+                if len(field_parsed) < 2:
```
yedpodtrzitko, line 1342: here you expect length exactly 1, not lower than 2

tagstudio/src/core/library.py
```diff
@@ -1320,6 +1319,47 @@ def get_entry_id_from_filepath(self, filename: Path):
...
+                    unbound_values = (
+                        field_parsed[0].strip().lower().casefold().split(" ")
```
yedpodtrzitko, line 1344: `"two  spaces  inbetween".split(" ")` will produce `['two', '', 'spaces', '', 'inbetween']`
`.split()` would be better, as it will produce the expected `['two', 'spaces', 'inbetween']`
```suggestion
                        field_parsed[0].strip().lower().casefold().split()
```

tagstudio/src/core/library.py
```diff
@@ -1332,229 +1372,37 @@ def search_library(
...
-            # 	print("[INFO][FILTER]: Filter returned no results.")
         else:
             for entry in self.entries:
                 added = False
                 allowed_ext = entry.filename.suffix not in self.ext_list
                 if allowed_ext == self.is_exclude_list:
-                    for f in entry.fields:
-                        if self.get_field_attr(f, "type") == "collation":
+                    for f in entry.fields:  # type: ignore [assignment]
+                        if self.get_field_attr(f, "type") == "collation":  # type: ignore [arg-type]
```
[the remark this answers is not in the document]
lunaro-4: This is actually part of the original code. Thank you for pointing it out, I did not notice it

tagstudio/src/core/library.py
lunaro-4: They actually won't
``` python
>>> l: list = ['no', 'empty', 'artist']
>>> 'no artist' in l
False
>>> 'no empty' in l
False
```
The problem is that some special flags in the original version contained spaces, and we should be combing separate 'no' and 'artist' into 'no artist' this way. It could be done, but converting it to a monolithic string is done much easily and more readable in my opinion

tagstudio/src/core/library.py
lunaro-4: This is needed for special queries like 'no authors' to work, since there are multiple fields with types `tag_box` and names related to `Author`. Although I could create loops to check every field if it is a `tag_box` or not, I think current solution is preffered, let me know what you think

lunaro-4:
Thank you for the feedback, I will comment again with a mention in a few days, when I apply your suggestions

tagstudio/src/core/library.py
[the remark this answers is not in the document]
yedpodtrzitko: this is original code, but `f = Filter(...)` is not

tagstudio/src/core/library.py
yedpodtrzitko: I'd assume the negative syntax needs to work for every field, otherwise that would be a bit non-intuitive that it works only for tags and authors 🤔

yedpodtrzitko: I see. And I dont know. This, together with the difficult way how the negative queries like 'no artist' are parsed, makes me really think that the syntax `-artist` would be more straightforward and easier to work with. I'm here just to review code for fun, but I dont make the product decision.

tagstudio/src/core/library.py
yedpodtrzitko: I see. And I dont know. I'm here just to review code for fun, but I dont make the product decision. But this, together with the difficult way how the negative queries like 'no artist' are parsed, makes me really think that the syntax -artist would be more straightforward and easier to work with.

tagstudio/src/core/library.py
CyanVoxel: Right, it's intentional design that there can be more than one field of a single type applied to an entry. For example, two author fields with two different authors.

CyanVoxel:
Would getting the field refactor in #157 tested and pulled first help with this PR, or at this point would it be like pouring water over a grease fire? *(Not to compare this PR to a fire, it's just a big feature on top of some of my worst code in the pre-existing codebase).* Since the database will be migrating anyway at some point, it wouldn't matter if the existing field types were locked in as constants that would be easier to work with while we're still using JSON.

yedpodtrzitko:
> Would getting the field refactor in #157 tested and pulled first help with this PR, or at this point would it be like pouring water over a grease fire? _(Not to compare this PR to a fire, it's just a big feature on top of some of my worst code in the pre-existing codebase)._ Since the database will be migrating anyway at some point, it wouldn't matter if the existing field types were locked in as constants that would be easier to work with while we're still using JSON.

any substantial PR will be most likely working with `library.py`, so there will never be better time to merge it than "now", as the number of potential conflicts in code will be always teh same, but working with the constants instead of magic numbers will make it easier to work with the existing code. I just rebased the PR to fix the conflicts, so from my point of view it's ready for merge.

lunaro-4:
Thank you for your patience, I had some IRL stuff last week and only now am able to return to the project.

Here is overview of last changes:
- **Proper negative prompt:** you can specify field contents, you want to exclude, by specifying field name with `-` in front of it:
   -  `-description` (in part without key specified) will exclude any entry with `description` field
   - `-description: foo` will exclude all entries, containing `foo` in description.
   - `-description: foo; description: bar` will select all entries, that contain `bar` but do not contain `foo` in description field
- **Minor code refactoring:** applied feedback from code review:
  - entity metafields' data is not lost, if entity has multiple fields with same ID
  - some original code rewritten for better readability
  - new `SpecialFlag` class
- **Prepared for merge**
  - [x] New code covered with tests
  - [x] Merged with one of recent commits
  - [x] All GitHub workflows green

lunaro-4:
@yedpodtrzitko @CyanVoxel 
Thank you for your patience, I had some IRL stuff last week and only now am able to return to the project.

Here is overview of last changes:
- **Proper negative prompt:** you can specify field contents, you want to exclude, by specifying field name with `-` in front of it:
   -  `-description` (in part without key specified) will exclude any entry with `description` field
   - `-description: foo` will exclude all entries, containing `foo` in description.
   - `-description: foo; description: bar` will select all entries, that contain `bar` but do not contain `foo` in description field
- **Minor code refactoring:** applied feedback from code review:
  - entity metafields' data is not lost, if entity has multiple fields with same ID
  - some original code rewritten for better readability
  - new `SpecialFlag` class
- **Prepared for merge**
  - [x] New code covered with tests
  - [x] Merged with one of recent commits
  - [x] All GitHub workflows green 
 
I can also write update for `README`, to clarify search syntax, if you approve my changes

tagstudio/src/core/library.py
```diff
@@ -1320,6 +1320,58 @@ def get_entry_id_from_filepath(self, filename: Path):
...
+                if len(field_parsed) > 2:
+                    logging.warning("""[ERROR] Attempt to parse mutiple fields\
+                                    as one! Do not forget to specify ';'\
+                                    between different meta fields""")
+                query_part[field_parsed[0].strip().casefold()] = (
+                    field_parsed[1].strip().casefold()
+                )
+            negative_flags: list = []
+            # selecting null fields are done with '-field' word in unbound part
+            if "unbound" in query_part.keys():
```
yedpodtrzitko, line 1360: nit
```suggestion
            if "unbound" in query_part:
```

tagstudio/src/core/library.py
```diff
@@ -2315,3 +2175,340 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+        self.get_tag_cluster = lib.get_tag_cluster
+        self.library_dir = lib.library_dir
+        self.get_field_obj = lib.get_field_obj
+        self.missing_files = lib.missing_files
+        self.default_fields = lib.default_fields
+
+        def generate_field_id_to_name_map(self) -> dict[int, str]:
+            map: dict = {}
+            for field in self.default_fields:
+                map[field.get("id")] = field.get("name").lower()
```
yedpodtrzitko, line 2210: `.get()` may return `None`, and it's most likely not desired to assign value into `map[None]` in case `id` will be missing. So it would be better to make it fail loudly here 
```suggestion
                map[field["id"]] = field.get("name").lower()
```

tagstudio/src/core/library.py
```diff
@@ -2315,3 +2175,340 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+            return map
+
+        self._field_id_to_name_map: dict = generate_field_id_to_name_map(self)
+
+    def filter_results(
+        self, query: str, split_query: list[dict], search_mode: SearchMode
+    ) -> list[tuple[ItemType, int]]:
+        # start_time = time.time()
+        query = query.strip().lower()
```
yedpodtrzitko, line 2219: other places are using `.casefold()` instead of `.lower()`, it would be good to unify the approach since both of these methods are behaving differently in some edge cases (eg. German's scharfes S).

tagstudio/src/core/library.py
```diff
@@ -2315,3 +2175,340 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+                                entry,
+                                all_tag_terms,
+                                entry_tags,
+                                entry_authors,
+                                query_words,
+                            )
+                        if entry_tuple is None:
+                            is_selected = False
+                            break
+                    elif key == "EMPTY":
```
yedpodtrzitko, line 2271: Could you define this value as a constant instead with a comment what it is, so it's obvious straight away? Thanks.

tagstudio/src/core/library.py
```diff
@@ -2315,3 +2175,340 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+                    filtered_entries.append((ItemType.ENTRY, entry.id))
+                elif is_selected and entry_tuple is not None:
+                    filtered_entries.append(entry_tuple)
+            pre_results.append(filtered_entries)
+
+        # Entries should match all parts separated by '|'
+        result_set: set = set()
+        if search_mode == SearchMode.AND:
+            if len(pre_results) == 1:
+                return list(pre_results[0])
```
yedpodtrzitko, line 2297: since we know the `pre_results` has 1 item, and it's a list, it could be enough to return the value as it is, no?
```suggestion
                return pre_results
```

tagstudio/src/core/library.py
```diff
@@ -2315,3 +2175,340 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+
+    def handle_common_field(
+        self, entry_value: list[str] | str | None, value: str, query_is_negative: bool
+    ) -> bool:
+        if isinstance(entry_value, list):
+            entry_value = "".join(entry_value)
+
+        if entry_value is None:
+            return query_is_negative
+        else:
```
yedpodtrzitko, line 2328: nit: this `else:` looks unnecessary, the code can be unintended

tagstudio/src/core/library.py
yedpodtrzitko: this line still looks unnecessary

tagstudio/src/core/library.py
```diff
@@ -2315,3 +2175,340 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+            added = False
+            for f in entry.fields:
+                if self.get_field_attr(f, "type") == "collation":
+                    if self.get_field_attr(f, "content") not in collations_added:
+                        collations_added.append(self.get_field_attr(f, "content"))
+                        return (
+                            ItemType.COLLATION,
+                            self.get_field_attr(f, "content"),
+                        )
+                    added = True
```
yedpodtrzitko, line 2372: afaik `collation` isnt currently implemented, so this doesnt do anything at this moment, and `added` will be always `False`, right?

tagstudio/src/core/library.py
yedpodtrzitko: can you add a docstring here what this method do?

tagstudio/src/core/library.py
```diff
@@ -2315,3 +2175,340 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+            return True
+        elif file_filename_fuzzy:
+            return True
+        return False
+
+    def required_fields_empty(
+        self, entry_fields: dict, empty_fields: list[str]
+    ) -> bool:
+        for field in empty_fields:
+            if field in entry_fields.keys() and entry_fields.get(field) is not None:
```
yedpodtrzitko, line 2494:
```suggestion
            if entry_fields.get(field) is not None:
```

yedpodtrzitko (changes requested):
this query still doesnt work as I'd expect. Am I doing something wrong?

https://github.com/TagStudioDev/TagStudio/pull/284#pullrequestreview-2120281121

tagstudio/src/core/library.py
```diff
@@ -2315,3 +2175,340 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+                ):
+                    if key_mapped in entry_fields.keys():
+                        entry_fields[key_mapped].append(value)
+                    else:
+                        entry_fields[key_mapped] = [value]
+                else:
+                    if key_mapped in entry_fields.keys():
+                        entry_fields[key_mapped].extend(value)
+                    else:
+                        entry_fields[key_mapped] = value
```
yedpodtrzitko, line 2451: this is perfect use-case for `collections.defaultdict`:
`entry_dict = defaultdict(list)`
then you can do:
```suggestion
                    entry_fields[key_mapped].append(value)
                else:
                    entry_fields[key_mapped].extend(value)
```

tagstudio/src/core/library.py
```diff
@@ -2315,3 +2175,340 @@ def sort_fields(self, entry_id: int, order: list[int]) -> None:
...
+
+    def preprocess_tag_terms(self, query_words: list[str]):
+        all_tag_terms: list[str] = []
+        for i, term in enumerate(query_words):
+            for j, term in enumerate(query_words):
+                if (
+                    query_words[i : j + 1]
+                    and " ".join(query_words[i : j + 1]) in self._tag_strings_to_id_map
+                ):
+                    all_tag_terms.append(" ".join(query_words[i : j + 1]))
```
yedpodtrzitko, line 2506: `" ".join(query_words[i : j + 1])` happens twice there, can it be DRYed?

samuellieberman:
Words with capital letters in text fields cannot be searched.

If the description is ```Banker goes to the circus``` then searching ```description: goes to the circus``` returns the entry, but ```description: Banker``` does not. Only the capitalization in the original field matters, capitalization in the search still does not matter. Searching ```description: GOES to THE CiRcUs``` still returns the entry.

I assume this bug applies to the Title, Author, Artist, URL, Description, and Notes fields, though I did not check them all.

samuellieberman:
A ```; no author``` search does not return an entry unless the entry has at least one tag, even though ```author: null``` and ```artist: null``` return the entry.

```-description: foo``` does not return any entries no matter what I've tried with my entries. Note the minus to indicate exclusion. ```-description: foo; description: bar``` also does not appear to work.

File: tagstudio/src/core/library.py (hunk 19 above)
Comment (lunaro-4) [reply tagstudio/src/core/library.py:2387]: @yedpodtrzitko 
This part is basically copy-paste from original code. I left it for backward compatability, but it is safe to remove, if feature was never implemented

**Comment by lunaro-4** [comment]
> this query still doesnt work as I'd expect. Am I doing something wrong?
> 
> [#284 (review)](https://github.com/TagStudioDev/TagStudio/pull/284#pullrequestreview-2120281121)

Sorry for not adressing this before.
You actually need to choose `OR` search type in GUI for it to work as you expect. Otherwise, if you leave it as `AND`, which is default value, you should recive about the same results as if you were using only `;`

**Comment by lunaro-4** [comment]
@samuellieberman 

Thank you for pointing things out, will fix and add additional tests

**Comment by lunaro-4** [comment]
Changes overview:

- Multiple minor changes from code review by @yedpodtrzitko
- New class `KeyNameConstants` for constant names, replaces hardcoded field names like "UNBOUND" and "EMPTY"
- Fix incorrect behavior with multi-word special flags like `no author`
- Fix negative prompt parsing

I also had to change search test from `test_lib.py` file, as `--nomatch--` case worked incorrectly with negative prompt implementation.

Thank you for patience, will be waiting for new review.

File: tagstudio/src/core/library.py [hunk 23]
```diff
@@ -43,6 +43,15 @@ class ItemType(Enum):
 logging.basicConfig(format="%(message)s", level=logging.INFO)
 
 
+class KeyNameConstants:
+    """Class, containing constants for special keys in search query, that are not supposed to be visible to user."""
+
+    # FILENAME_KEYNAME = "filename"
+    # TAG_ID_KEYNAME = "tag_id"
```
eivl, line 50: Unused attributes can be deleted.

eivl (commented):
It is to complex and nested for me to understand what it does. Im happy to approve it once I have had time to understand this PR. @lunaro-4 if you want to explain it to me I can give it another shot later.

lunaro-4:
@eivl
The intention behind this PR is to give the user the ability to search entries by multiple custom fields. Current version of TagStudio only supports search by tags and filename, with this PR it is possible to filter entries by other fields, such as `Description`.

At first, there was only one `search_libraty` method, that did all the work. What I did was creating new method `parse_metadata` that parses arguments from search query, and extracting functionality from `search_library` into a separate class, called `Filter`. On initialization, this class copies all necessary information from `Library`. After that, `Filter.filter_results` could be called to get all entries, that fit the parsed search query. 

# Reading the query

`Library.parse_metadata` method splits the query into multiple parts, returning a list of dictionaries:

- list element: query parts, which are separated by `|` symbol in query
- dictionary element: query keys, provided inside query parts and separated by `;` symbol, and it's value (if `:` symbol is provided)

Query keys that do not have values added to the 'unbound' dictionary key (name set in `KeyNameConstants.UNBOUND_QUERY_ARGUMENTS_KEYNAME` constant), and later treated as a whole query in original code. 

`negative prompt` values are extracted from `unbound` values and put in dictionary key, specified in `KeyNameConstants.EMPTY_FIELD_QUERY_KEYNAME`. These values filter out entries, containing non-null values of specified query key.


# Filtering the entries

`Filter.filter_results` method contains a nested loop, that iterates over the output from `parse_metadata`.

It iterates over list elements (`for query_part in split_query`), then each entry (`for entry in self.entries`), and then for each entry it iterates over contents of dictionary (`for key, value in query_part.items()`)

If the entry does not break loop, it is considerate valid for query and is added to list `filtered_entries`.

The `filtered_entries` list is created for each part of the query, and then added to `pre_results` list. 

The `pre_results` is a list of lists, and each element `pre_results[n]` is a representation of `parse_metadata` element, as if the whole query was just `parse_metadata()[n]'.

`pre_results` elements are then combined into a single set, depending on `SearchMode`, selected in GUI.

lunaro-4:
I see a merge conflict, and will try to resolve it this weekend

I also would like to hear some tips to make code more readable. For me, this level of nesting is more or less comfortable and I do not know what to do with `if` statements in a loop at 2263 (`for key, value in query_part.items():`)

PR #286, opened 2024-06-13
add tests into CI

yedpodtrzitko:
partially overlaps with #199, but that one has several issues, so it feels faster to do it from scratch and add some real tests at that opportunity.


CyanVoxel (approved):
A much-needed kickstart for tests, thanks for getting this going! I suppose either #199 can be revisited after this or it can be closed in favor of a new PR with additional tests.

PR #293, opened 2024-06-14
fix: pip cache failures in workflows

xarvex:
From #255:
> Ocassionally, a pip install will fail because of mismatched hashes, which from my research I believe this to be related to a cache error. This cannot easily be tested due to its seemingly random nature, but my hope is that my fix in using the action/setup-python workflow will avoid it.

Unfortunately the fix in [82037e5](https://github.com/TagStudioDev/TagStudio/pull/255/commits/82037e570d8376ebf4007430bc45c5cd0f312666) seems to have just caused registered bad cache to be reused over and over. So, this serves as a sort of "opposite" to that and removes using the cache altogether.

Again, due to random nature, this cannot easily be tested and will show results over time.

### DIFF
--- a/.github/workflows/apprun.yaml
+++ b/.github/workflows/apprun.yaml
@@ -14,7 +14,6 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: '3.12'
-          cache: 'pip'
 
       - name: Install system dependencies
         run: |
@@ -37,7 +36,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install -Ur requirements.txt
+          pip install --no-cache-dir -r requirements.txt
 
       - name: Run TagStudio app and check exit code
         run: |
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -19,14 +19,13 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: '3.12'
-          cache: 'pip'
 
       - name: Install dependencies
         run: |
           # pyside 6.6.3 has some issue in their .pyi files
-          pip install PySide6==6.6.2
-          pip install -r requirements.txt
-          pip install mypy==1.10.0
+          pip install --no-cache-dir PySide6==6.6.2
+          pip install --no-cache-dir -r requirements.txt
+          pip install --no-cache-dir mypy==1.10.0
           mkdir tagstudio/.mypy_cache
 
       - uses: tsuyoshicho/action-mypy@v4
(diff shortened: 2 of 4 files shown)


yedpodtrzitko:
This is most likely caused by `requirements-dev.txt` which is not taken into account when creating the cache key, only `requirements.txt` is considered by default.

This can be fixed by specifying [cache-dependency-path](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages)

I'm not sure if this wouldn't exclude `requirements.txt` for a change:
```
cache-dependency-path: '**/requirements-dev.txt'
``` 

so using this should be safe:
```
cache-dependency-path: '**/requirements*.txt'
```

xarvex:
> This is most likely caused by `requirements-dev.txt` which is not taken into account when creating the cache key, only `requirements.txt` is considered by default.
> 
> This can be fixed by specifying [cache-dependency-path](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages)
> 
> I'm not sure if this wouldn't exclude `requirements.txt` for a change:
> 
> ```
> cache-dependency-path: '**/requirements-dev.txt'
> ```
> 
> so using this should be safe:
> 
> ```
> cache-dependency-path: '**/requirements*.txt'
> ```

I would think this if it weren't for the fact that packages in `requirements.txt` are affected by this too, and actually happens with only those packages as far as I have been able to tell. For example, the release workflow which only uses `requirements.txt` experienced this error recently.

yedpodtrzitko:
So this means the error happens during download, which means the problem isnt in the cached files. I tried to search for the error message, and it seems `pip` doesnt know the connection was interrupted, considering the download is complete which causes the hash mismatch.

In that case not using cache wont help, it might make the problem worse as we'll be downloading each dependency each time, increasing the chance for the mismatch.

Is there any occurrence of this problem for the `pytest` workflow? In that one there's upgrade to latest version of `pip` (`python -m pip install --upgrade pip`) which we can hope has the problem fixed.

Other than that I found that `uv` has also dealt with this problem, but they actually fixed it ( https://github.com/astral-sh/uv/issues/1912 ), so using `uv` instead of `pip` could be worth trying:

```
pip install uv
uv pip install -r requirements.txt
```

xarvex:
Going back to the drawing board for this.

yedpodtrzitko:
this was most likely fixed within #332 by using `uv` and I havent experienced any errors, so we can close this:

https://github.com/TagStudioDev/TagStudio/blob/fb949b82b7e3ed76a70446bd5cf7d2f18c940266/.github/workflows/pytest.yaml#L39-L43

PR #297, opened 2024-06-16
Detach Library loading from Render thread

seakrueger:
In scenarios where the library takes a long amount of time to load, the application would become locked up and unresponsive, as library loading was blocking. By detaching library loading to its own thread, the program remains responsive even while the library is loading.

Eventually, a ProgressWidget should probably be added, but that would require a signifigant change to the logic of lib.open_library.

### DIFF
--- a/tagstudio/src/qt/helpers/custom_runnable.py
+++ b/tagstudio/src/qt/helpers/custom_runnable.py
@@ -7,7 +7,7 @@ from PySide6.QtCore import Signal, QRunnable, QObject
 
 
 class CustomRunnable(QRunnable, QObject):
-    done = Signal()
+    done = Signal(object)
 
     def __init__(self, function) -> None:
         QRunnable.__init__(self)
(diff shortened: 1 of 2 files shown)


Thesacraft:
Don't worry about the Pyside app test failing for now (its a bug that is fixed now). But if you want to you can fix it by pulling the changes from the main branch and merging it into yours.

tagstudio/src/qt/helpers/custom_runnable.py
```diff
@@ -7,13 +7,13 @@
 
 
 class CustomRunnable(QRunnable, QObject):
-    done = Signal()
+    done = Signal(object)
```
eivl, line 10: why object?

eivl (approved):
I dont see anything incorrect. approved when my comment is addressed.

tagstudio/src/qt/helpers/custom_runnable.py
seakrueger: > CustomRunnable needs the
ability to return the int value. Since CustomRunnable is designed to be
generic, the signal returns of a value of any `object` type.

seakrueger:
Closing as this seems to have been solved more eloquently in SQL migration

PR #302, opened 2024-06-28
feat(qt): Add hotkey to add tags to selection

dtomvan:
This is used so that you don't have to go through the mouse-heavy
nitty-gritty of:

1. Select one or more files
2. Add field
3. Tags
4. OK
5. Add Tag
6. Click on the plus sign
7. Done

To get to a tag selection modal. This reduces a lot of friction in my
opinion. I like the way that the fields can have multiple types, but in
the majority of cases I just want to quickly tag one or more files in my
selection and move on. The workflow after this change is as follows:

1. Select one or more files
2. Just press `Ctrl+Shift+T`
3. Click on the plus sign
4. Done

This way the tagbox gets added automatically (through the powerful core
library system) and you can immediately start tagging!

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -55,7 +55,7 @@ from PySide6.QtWidgets import (
 )
 from humanfriendly import format_timespan
 
-from src.core.enums import SettingItems, SearchMode
+from src.core.enums import FieldID, SettingItems, SearchMode
 from src.core.library import ItemType
 from src.core.ts_core import TagStudioCore
 from src.core.constants import (
@@ -85,6 +85,7 @@ from src.qt.modals.file_extension import FileExtensionModal
 from src.qt.modals.fix_unlinked import FixUnlinkedEntriesModal
 from src.qt.modals.fix_dupes import FixDupeFilesModal
 from src.qt.modals.folders_to_tags import FoldersToTagsModal
+from src.qt.modals.tag_search import TagSearchPanel
 
 # this import has side-effect of import PySide resources
 import src.qt.resources_rc  # pylint: disable=unused-import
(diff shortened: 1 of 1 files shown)


Thesacraft (commented):
Its a bit nitpicky but i think it should make it more readable.

dtomvan:
I agree with your proposed changes, however, I think that those issues should be fixed in one fell swoop as per a refactor changeset. I don't think that should be included in a feature PR. Do as you want though; maybe I'll look at the refactoring I could do later on.

dtomvan:
To add to that: I've written my PR consciously in this way, as to stay consistent with the existing code. I know that constitutes "copy-paste code", but I thought it was appropriate here.

Thesacraft:
Yeah i thougt that and you are probably right that the refactoring part is better of in another pr.

eivl (approved):
I agree with the other comments, but it is still fair to use existing code conventions. I approve.

CyanVoxel:
Unfortunately this is going to need some refactoring to work with the new SQL backend (#332). I understand this PR has been open for a while now, and I apologize for taking a while to get around to it.

The merge conflict here is pretty minor, and from what I can tell the `FieldID` has changed places and is imported as `FieldID_` now. The `self.selected` list is now consists solely of Entry ID ints so it simplifies some of the logic there. Finally the logic for adding Tags to Entries themselves has changed, and this is something I need to become more familiar with myself before giving instructions on how it currently operates.

I love the idea of this feature, and it brings things one step closer to being fully keyboard controlled. My only gripe is that there's no way to choose which type of Tag Field the tag gets added to. Personally I use "Meta Tags" and "Content Tags" in my library, so I would effectively be locked out of using such a great shortcut. *I'm also obligated to mention that the entire "Tag Fields" concept will eventually be replaced by automatic tag categories in a future update, so issues like this will no longer be a concern.*

CyanVoxel:
Made obsolete by #655 which removed tag fields in favor of applying tags directly to entries and then automatically sorting them by category

dtomvan:
Oops, haven't looked back at this PR for a while now. Good to hear things are better overall for the software. Apologies accepted 😂

PR #315, opened 2024-07-12
Fix file path intercompatibility for Windows and POSIX systems

### LINKED ISSUE #298: [Bug]: File Paths not Intercompatible between Windows and Linux
### Checklist

### TagStudio Version

Alpha 9.3.1

Windows 11 and Arch Linux

### Description

When copying the Tagstudio folder from Windows to Linux the Linux version can not open the Tagstudio Library properly due to the Windows version saving the path as `Subfolder\\File` using two backslashes, which the Linux version can't read properly.

Here is a screenshot of what the Linux device read.
image

Manually replacing `\\` with `/` fixes the issue on Linux and still works properly on Windows

We're in love with Bocchi

### Expected Behavior

Moving a Tagstudio Library between two different operating systems should open properly and work.

### Steps to Reproduce

1. Create a Tagstudio Library with subfolders on Windows and Tag files
2. Move the Tagstudio Library to a Linux system
3. Try opening the Tagstudio Library

### Logs

_No response_

searinminecraft:
Fixes #298 

This fixes a bug where if you move a library saved on Windows to Linux or any POSIX operating system, it will cause the files to not be found due to backslashes. This is a hack I done to prevent this from happening.

Needs extensive testing though, even though that I have confirmed this fix works and does not crash Tag Studio.

cc @coolesding

### DIFF
--- a/tagstudio/src/core/library.py
+++ b/tagstudio/src/core/library.py
@@ -13,7 +13,7 @@ import xml.etree.ElementTree as ET
 import ujson
 
 from enum import Enum
-from pathlib import Path
+from pathlib import Path, PureWindowsPath
 from typing import cast, Generator
 from typing_extensions import Self
 
@@ -52,7 +52,7 @@ class Entry:
         # Required Fields ======================================================
         self.id = int(id)
         self.filename = Path(filename)
-        self.path = Path(path)
+        self.path = Path(Path(path).as_posix())
         self.fields: list[dict] = fields
         self.type = None
 
--- a/tagstudio/src/core/library.py
+++ b/tagstudio/src/core/library.py
@@ -52,7 +52,7 @@ class Entry:
         # Required Fields ======================================================
         self.id = int(id)
         self.filename = Path(filename)
-        self.path = Path(Path(path).as_posix())
+        self.path = Path(path).as_posix()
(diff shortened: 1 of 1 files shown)


searinminecraft:
Well thats a concern.. putting Path() on it would just make the bug return. (it becomes \\ again, making the as_posix() useless). Is there any other thing i can do?

CyanVoxel:
Since Windows-style backslashes are valid characters in Posix paths, it makes things a bit tricky.

I would say the best course of action to fix #298 would be to always save and load file paths inside of `ts_library.json` as Posix paths moving forward, regardless of the OS. In addition there would need to be consideration for libraries saved with versions before this fix (currently 9.3.2 and earlier) - I propose replacing any "\" characters with "/" in older library saves. This would, however, break data inside of anyone's libraries on Posix systems that currently contain file entries that use backslashes as legitimate file characters. I believe this can be avoided by performing a "file exists" check on both the old and new file paths, and only performing the conversion in cases where the swapped path comes back valid. If both cases come back false, then there's an existing unlinked entry that can't be broken any further from us messing with it.

searinminecraft:
I have no motivation in trying to fix this. Sorry...

PR #318, opened 2024-07-13
[chore]Make get_field_attr safe and improve control flow in ts_core.py

dardevelin:
Refactor get_field_attr in library.py:
- Ensure safety using list slicing to avoid IndexError.
- Use .get() for dictionary access to handle default cases and prevent KeyError.

Improve control flow in ts_core.py:
- Remove nesting by exiting early in match_conditions.
- Eliminate the need for try-catch blocks for better readability and control flow.

Note: Kept existing_fields case unchanged due to add_field_to_entry's procedural nature.

this commit introduces changes to library.py get_field_attr

### DIFF
--- a/tagstudio/src/core/library.py
+++ b/tagstudio/src/core/library.py
@@ -2164,7 +2164,7 @@ class Library:
     # 	entry = self.entries[entry_index]
     # 	pass
     # 	# TODO: Implement.
-    
+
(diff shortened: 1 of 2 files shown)


dardevelin:
There is a bug on the assumption for list slicing. Closing and opening a new one with the change fixed

PR #322, opened 2024-07-17
Add Tag Merging

### LINKED ISSUE #12: Tag Merging
Let's say you have tag "atsuko kagari" and "akko", both are the same character, but one is a nickname for the other. You've tagged some files with Akko and some others with the full name, but finally want to merge them together under a single tag.

akko can become an alias for the full name, but then all existing files with that tag need to be processed with the following:
- add the new tag
- remove the old tag
- add all parent-child relationships from the old tag to the new one
- remove all parent-child relationships from the old one

doing this with a single modal would be helpful :)

blatessa:
Added the ability to merge tags in the tags context menu (should this be in the manage tags modal only?). Merge tag adds the source tag name and aliases to the target tag, appends all subtags to the target tag, reassigns all entries to the target tag, and removes the source tag. 

Let me know if I'm missing anything in the merge process. Also, when the modal closes the ui state doesn't immediately show the merge. Any ways to make it more reactive?

Closes #12

### DIFF
--- a/tagstudio/src/core/library.py
+++ b/tagstudio/src/core/library.py
@@ -1899,12 +1899,9 @@ class Library:
         source_tag_id: int = source_tag.id
         target_tag_id: int = target_tag.id
 
-        if not source_tag or not target_tag:
-            return 0
-        
         if source_tag.name not in target_tag.aliases:
             target_tag.aliases.append(source_tag.name)
-            
+
         for alias in source_tag.aliases:
             if alias not in target_tag.aliases:
                 target_tag.aliases.append(alias)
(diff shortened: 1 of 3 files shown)


CyanVoxel:
Thank you for your work on this! I apologize for not getting around to this sooner. The tag merging logic seems to be sound, but there's a few UI improvements I'd like to add myself in addition to the issue you mentioned about the tag search modal not updating. Rather than work off your PR, I'll be pulling this as-is into a feature branch before touching up the UI and eventually merging into the v9.4 branch.

PR #331, opened 2024-07-24
refactor: move type constants to new media classes

CyanVoxel:
This refactor is meant to provide an improved interface for determining media categories from file extensions. Instead of relying on searching though lists of extension constants, this refactor provides a new `MediaCategory` class, which in turn is composed of a `MediaType` enum, a set of extensions defined by one or more internal sets, and additional data such as IANA registration (more on that later).

The `MediaCategory` class is then used by the `MediaCategories` class, which contains every pre-built category used by the program along with the `get_types` method, which is the new intended way to check for media category membership. Instead of checking if extension `.foo` is inside a `BAR_TYPES` list constant, `foo` is passed to the `get_types` method which returns any `MediaType` matches from the extension sets inside of `MediaCategory` objects added inside the `ALL_CATEGORIES` list.

While I originally intended to rely heavily on MIME types rather than curated extension lists, it became apparent to me that this wouldn't be feasible for the use cases of TagStudio. There are cases where multiple media categories are desired, such as `.csv` in both `PLAINTEXT` and `SPREADSHEET` categories. In addition, MIME type categories on their own are too limited in scope for some of categorization I'd like to achieve, such as with Office files. With that being said I still wanted the ability to leverage MIME types where possible. The `get_types` method has an optional flag that when true will leverage the `mimetypes` module and attempt to guess the media category of an extension if a match was not found in one of the built-in sets. The IANA registration flag within the `MediaCategory` class is then used to prevent the `get_types` method from doing unnecessary membership checks inside otherwise unregistered media categories.

**TL;DR:** New `MediaType` enums and `get_types` method to replace the extension list constants with a more flexible system.


tagstudio/src/core/media_types.py
```diff
@@ -0,0 +1,408 @@
...
+        ".pptx",
+        ".key",
+        ".odp",
+    }
+    _PROGRAM_SET: set[str] = {".app", ".exe"}
+    _SHORTCUT_SET: set[str] = {".desktop", ".lnk", ".url"}
+    _SPREADSHEET_SET: set[str] = {
+        ".csv",
+        ".numbers",
+        ".ods" ".xls",
```
Thesacraft, line 231: I think there's a comma missing here

PR #336, opened 2024-07-30
Tag ordering backend

samuellieberman:
This is intended to provide more and more complicated tag sorting options than #329, as well as bringing tag sorting functionality into its own file to be used in multiple places. Currently the code works and sorts all results in the add tag panels, but the code itself is a work in progress. See commit b9316fa1f488205dc793117d7dc6df05fd2c9709 for a little of the work I have planned.

The most notable tag sorting option is sorting tags based on an automatically generated lineage, a chain of parent tags present in the search. This feature is set up so that parent tags and their children and grandchildren will all appear next to each-other in the list of tags, in a tree-like fashion. It is set up to automatically respect any sort properties that come before or after it. Here is an example with ascending lineage depth:
image

To mess around with the sort order, modify `_default_sort`.
https://github.com/TagStudioDev/TagStudio/blob/8444b8d4700aa2af3a92fef79d83f1354b3a522f/tagstudio/src/core/tag_sort.py#L25-L32
Try changing the order of the properties and swapping between `TagSortDirection.ASC` and `TagSortDirection.DESC`

Currently these are the only tag properties that are supported. In the future, I plan to add other options, such as sorting tags based on how many times they appear in the library's entries. I am taking suggestions for other tag properties to calculate and sort by.

### DIFF
--- a/tagstudio/src/core/tag_sort.py
+++ b/tagstudio/src/core/tag_sort.py
@@ -1,0 +1,243 @@
+from enum import Enum, Flag, auto
+from typing import Any, Callable
+
+from src.core.constants import TAG_COLORS
+from src.core.library import Library
+
+
+class TagSortProperty(int, Enum):
+    MATCHES_PREFIX = auto()
+    CATEGORY = auto()
+    LINEAGE_DEPTH = auto()
+    DESCENDENT_COUNT = auto()
+    NAME = auto()
+    COLOR = auto()
+    ID = auto()
+
+
+class TagSortDirection(Flag):
+    ASC = False
+    DESC = True
+
+
+Sort = list[tuple[TagSortProperty, TagSortDirection]]
+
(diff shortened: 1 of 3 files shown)


eivl:
Any news on this work @samuellieberman ? i would love to do a review once you are ready and all tests are passing.

CyanVoxel:
> Any news on this work samuellieberman ? i would love to do a review once you are ready and all tests are passing.

This one will likely need to get reworked once #332 is merged

seakrueger:
Just wanted to bump this and check if this is still being worked on. Unfortunately, since the SQL migration, this might require a significant rework. I think this could a useful feature especially with newer, similar requests (I.e; #542). Thanks!

CyanVoxel:
Closing due to inactivity - If this is a mistake and you're still interested in working on this, please let me know

PR #353, opened 2024-08-21
Update Numpy to version 2.1.0

Thesacraft:
I Updated the Numpy version Tagstudio is running on, to the newest version compatible with rawpy and opencv. I also updated the version they are pinned to.
I tested it a bit but feel free to test yourself

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,12 +1,12 @@
 humanfriendly==10.0
-opencv_python>=4.8.0.74,<=4.9.0.80
+opencv_python==4.10.0.84
 Pillow==10.3.0
 PySide6==6.7.1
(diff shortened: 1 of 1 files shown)


eivl:
I think that a pyproject file would be the place to add dependencies and the requirements files should be pinned based on the resolver. I do approve on this change as it is.

eivl (approved):

CyanVoxel (approved):
Seems to work just fine on Windows 10 and macOS, thanks for this!

CyanVoxel:
Shoot just realized that this was pointing to the 9.4 branch - shouldn't break anything on there, but I'll add these changes to main as well

PR #344, opened 2024-08-09
add JPEG XL support via pillow-jpegxl-plugin

BPplays:
this PR adds JPEG XL support via a pillow plugin [pillow-jpegxl-plugin](https://github.com/Isotr0py/pillow-jpegxl-plugin).
I've tested on JPEG XL images made with [libjxl](https://github.com/libjxl/libjxl) `cjxl` and it seems to work fine
​
​
​
this closes #333

PR #357, opened 2024-08-21
improved handling of animated images

### LINKED ISSUE #333: [Feature Request]: jpeg xl support
### Checklist

### Description

supporting [jpeg xl](https://en.wikipedia.org/wiki/JPEG_XL) would be nice.
in my testing it seems to have better lossless compression then webp, and way way better lossloss compression then avif

### Solution

_No response_

### Alternatives

_No response_

BPplays:
I've changed the handling of animated images to include support for more types of animated images.
PR features:
- includes transcoding from an animated image type to webp or gif using pillow.
- only counts an image as animated if it has more then 1 frame
- checks what image types are support by qt and by pillow, as well as includes a list of known good types what pillow can encode with animation, and sorts available image types by quality

this PR also includes the changes from #344 so feel free to close that one if you prefer this one, though sadly that plugin doesn't seem to support decoding animated jxls

Closes #333


BPplays:
i added some code to test converting into animated webp before showing the image but it's sometimes causing crashing and i have no idea why, when it works for a small amount of time it seems to work.

BPplays:
i think i might have fixed the crashing

CyanVoxel:
Wanted to chime in here with the state of this PR and other relevant developments around the project.

#344 was at a sticking point regarding licenses, but is probably going to move forward with [pillow-jpegxl-plugin](https://github.com/Isotr0py/pillow-jpegxl-plugin) assuming it functions correctly (I need to get my hands on some .jxl files).

In the meantime, I've unknowingly implemented some of changes here in #409 - mostly the animated image media type category and some code for loading GIFs into memory rather than streaming them from disk.

I would also suggest rebasing this to `Alpha-v9.4`, however there's likely due to be more than a few merge conflicts... Some of this is due to the changes mentioned above, and others are due to some git mishaps that happened in a recent update of `Alpha-v9.4`. But with the rest of the Alpha v9.4 milestone features out of the way, this should have a clear path on that branch now (outside of #344 presumably getting merged).

BPplays:
@CyanVoxel i manually merged the work from the only version of this branch on to `Alpha-v9.4`

eivl:
Are no tests affected by this PR after the gif changes? And is there a way to add missing tests?
I don't see anything that makes me think this would not work; I would also like a short description of how to test this live on my machine.

BPplays:
@eivl are there any tests for animated image handling?
i couldn't find any after a quick look in `./tagstudio/tests` in the `Alpha-v9.4` branch.

BPplays:
@CyanVoxel i think the PR is good enough for you to have another look if you want.
i also moved setting what preview media type to display to it's own function.

tagstudio/src/qt/widgets/preview_panel.py
```diff
@@ -601,28 +712,15 @@ def update_widgets(self) -> bool:
...
-                        self.preview_gif.show()
+                    if MediaCategories.is_ext_in_category(
+                        ext, MediaCategories.IMAGE_ANIMATED_TYPES, mime_fallback=True
+                    ):
+                        with open(filepath, mode="rb") as f:
+                            file_bytes: bytes = f.read()
+                            image: Image.Image = Image.open(io.BytesIO(file_bytes))
+                            if hasattr(image, "n_frames"):
+                                if image.n_frames > 1:
+                                    self.set_anim_img(filepath, ext, file_bytes, image)
```
CyanVoxel, line 723: Just an approval comment: I like the check to see if an image has multiple frames - for some reason Qt displays QMovies a bit strangely compared to static images, so this check is nice for improving the quality of non-animated images that share extensions with animated counterparts (i.e. gif).

tagstudio/src/qt/widgets/preview_panel.py
```diff
@@ -130,21 +139,48 @@ def __init__(self, library: Library, driver: "QtDriver"):
...
+
+        self.preview_anim_img_fmts = [fmt.data().decode("utf-8") for fmt in qmovie_formats]
+
+        ani_img_priority_order = ["jxl", "apng", "png", "webp", "gif"]
+
+        self.preview_anim_img_pil_map = {"apng": "png"}
+
+        self.preview_anim_img_pil_map_args = {"gif": {"disposal": 2}}
+
+        self.preview_anim_img_pil_known_good = {"webp", "gif"}
```
CyanVoxel, line 172: There was a lot of this code that I opted to leave out when porting the functionality over due to it being difficult to decipher the purpose of and to trace back what was occurring inside the animated image creation block.

If it's necessary for important additional functionality, I would appreciate some documentation on what exactly all of these maps are being used for and why they need to be used over just loading in the image as a gif/webp.

BPplays: @BPplays this is for choosing image format priority
- ani_img_priority_order: is an order of image extensions that are sorted first if they're supported by both pillow and qmovie
- preview_anim_img_pil_map: this maps what extension to use in pillow, pillow encodes animated pngs with the format 'png' instead of 'apng' so if apng comes up anywhere pillow should use png
- preview_anim_img_pil_map_args: this is special args for pillow should use for specific encoding formats, here gif has a trail on transparency if this is not used, but webp does not need it.
- preview_anim_img_pil_known_good: it's a list of known image types pillow can encode with animation, i couldn't find a way to test this in the program, ~it also probably needs expanding~ i made a commit

BPplays:
@CyanVoxel it seems like it blocks the gui thread while it's saving an image causing quite a bit of lag for large-ish images, 
this should be fixed by saving the image, waiting for it to be done then loading it without blocking the gui but im not sure how to do that.

xarvex:
The overall functionality present here has been achieved in the program at this point, with the significant hole of JXL support (#333) being resolved with #549. As for the part in the animated image handling, there doesn't seem to be any specific reason not to just load an image as a GIF or WebP, as brought up previously. There is a lot of logic and burden to do all this switching and checks, that might even need to be changed and adjusted as underlying API and support changes. There also seems to be the issues with large files and threading, as such I'm afraid the implementation here cannot be accepted.

Issue #365, opened 2024-08-22
[Feature Request]: Set Up/Switch To Poetry for dependency management. 


Joshua-Beatty:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

Switching to poetry would simplify dependency and venv management, allowing new developers to get started with just two commands regardless of platform. Ideally be able to run `poetry install` then run `poetry run python tagstudio\tag_studio.py` and have the project just work.

### Solution

Modify the pyproject.toml to support Poetry Dependencies.

### Alternatives

Alternatively set up support for PDM.

Issue #200, opened 2024-05-21
Support `nix run`


Arya-Elfren:
Allow people to run `nix run github:TagStudioDev/TagStudio` to build and run the app.

Familex:
Can be done with [poetry2nix](https://github.com/nix-community/poetry2nix) if #368 gets merged

zierf:
I have a nearly complete _flake.nix_ for this in my [poetry2nix Branch](https://github.com/zierf/TagStudio/tree/poetry2nix).
The flake has a _devShell_ providing `poetry` (and also `qtcreator`, `mypy`, `ruff` as before).

It would also solve #415 and #422.

<details>
<summary>415: Application Icon, Name and global menu</summary>

![Screenshot_20240901_155513](https://github.com/user-attachments/assets/635f292f-126f-4121-ada3-a68ca14e9436)
</details>
<details>
<summary>422 Pin to panel</summary>

![Screenshot_20240901_165559](https://github.com/user-attachments/assets/665ce5cb-3629-4090-ae95-d6002f367f39)
</details>

Please note that the icon and application name only appear if the application has been installed in the system. A _.desktop_ file is then available, which will also beused when starting from the _devShell_.
The same applies to pinning.


I think `nix run` only works with `apps.${system}.default` when you are in the flake itself. Otherwise, after installation, I get an error message like:
```SH
       error: A definition for option `environment.systemPackages."[definition 1-entry 3]"' is not of type `package'. Definition values:
       - In `/nix/store/ajvaxx5hvdhcdr3r9f0wpww61b05x9i2-source/configuration.nix':
           {
             program = "/nix/store/6srq67rl2fnawd9q4ikm6npqgx8nbg17-python3.12-tagstudio-9.3.2/bin/tagstudio";
             type = "app";
           }
```
When providing a derivation with `packages.${system}.default`, it can also simply be started directly from the system with `tagstudio`.

---

The only remaining problem is that I have to add `python312.pkgs.pyside6` as an extra dependency. Due to this additional dependency, the execution at least works in the _devShell_ itself with `nix run` etc.

However, if you install the Flake into the system, you get this error:
```SH
Traceback (most recent call last):
  File "/nix/store/lkih98vq66kb3cfw3j961bln7wd30843-python3.12-tagstudio-9.3.2/bin/..tagstudio-wrapped-wrapped", line 6, in <module>
    from tagstudio.tag_studio import main
[...]

PR #368, opened 2024-08-22
Add poetry dependency manager and package builder

tintin10q:
# What is this?

This pull request adds poetry support. Poetry provides an easy way to install dependencies and to publish.
This change was first suggested by #365 but I actually started working on it just before that issue was created.

As was stated in that issue:

> Switching to poetry would simplify dependency and venv management, allowing new developers to get started with just two commands regardless of platform. 
> Ideally be able to run `poetry install` then run `poetry run python tagstudio\tag_studio.py` and have the project just work.

I improved it to the point where you can get started with just this:

```shell
poetry install
poetry run tagstudio
```

That tagstudio run script also sets up a [console_script](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html). That basically allows you to start tagstudio by just calling `tagstudio` in a shell after installing with pip.  

Using poetry you can also easily publish the package to pypi so that one can actually install using pip with:

 ```poetry publish --username __token__ --password <api_token_generated_on_pypi>```

[link to poetry website](https://python-poetry.org/)

# Setting up publishing infrastructure 

On this discord it was suggested to make the publishing infrastructure part of the same pr.  Currently if you want to publish just adjust the version number and issue the comand above.

However, I think its nice to make a github action so that it publishes a new release when you make a github release.
I found a pretty good looking github workflow [here](https://github.com/marketplace/actions/pypi-poetry-publish)

I saw that on pypi that you can add github as a Trusted Publisher [here](https://pypi.org/manage/project/tagstudio/settings/publishing/). I assume this works by adding the workflow and then you can simply point pypi to that yml file. 

But honestly I have not set that up before. For this reason I think this should just be a separate pr. Also because do we even want to use poetry for this? I think its a good idea because then dependency management and publishing is just one tool. 

I suggest to just merge this and setup the automatic release to pypi (and apt) upload when you actually want to do a release. 

Also do not forget that pypi has a test version where you can test out uploads without wasting a version number on the official one. 

# Things to know 

- The CliDriver currently has many errors in it. For this reason I decided to not include it in the published package. That means the code is not there when you install it. To make that work I made each driver only be imported when you actually use it using the --ui option. So that means that --ui cli will not work because the code is not there. But it was not going to work anyways because the cli code currently does not work in the first place.  
- To make the publishing work I had to make `tagstudio/resources` an importable package otherwise I could not include it in the wheel.   
- The maintainers field in pyproject.toml is currently empty. But at some point that should probably be filled with the people that contributed code here

Let me know if you want me to change anything :)


pyproject.toml
```diff
@@ -7,3 +7,51 @@ disable_error_code = ["union-attr", "annotation-unchecked", "import-untyped"]
 explicit_package_bases = true
 warn_unused_ignores = true
 exclude = ['tests']
+
+[tool.poetry]
+name = "tagstudio"
+version = "9.1.1"
+packages = [{include = "tagstudio/tag_studio.py"}, {include = "tagstudio/src/core"}, {include = "tagstudio/src/qt"}, {include = "tagstudio/resources"}, {include = "tagstudio/__init__.py"}]
```
Sigmanificient, line 14: can this be put in multiline for better readability?
```suggestion
packages = [
    {include = "tagstudio/tag_studio.py"},
    {include = "tagstudio/src/core"},
    {include = "tagstudio/src/qt"},
    {include = "tagstudio/resources"},
    {include = "tagstudio/__init__.py"}
]
```

pyproject.toml
```diff
@@ -7,3 +7,51 @@ disable_error_code = ["union-attr", "annotation-unchecked", "import-untyped"]
...
+
+description = "TagStudio (Alpha): A User-Focused Document Management System TagStudio is a photo & file organization application with an underlying system that focuses on giving freedom and flexibility to the user. No proprietary programs or formats, no sea of sidecar files, and no complete upheaval of your filesystem structure."
```
Sigmanificient, line 16: The description should be kept short as it apears on the small banner section.
```suggestion
description = "User-Focused Photo & File Management System"
readme = "README.md"
```

tagstudio/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = '9.1.1'
```
Sigmanificient, line 1:
```suggestion
__version__ = '9.1.1'

__all__ = ("__version__",)
```

Sigmanificient (changes requested):

pyproject.toml
```diff
@@ -7,3 +7,51 @@ disable_error_code = ["union-attr", "annotation-unchecked", "import-untyped"]
 explicit_package_bases = true
 warn_unused_ignores = true
 exclude = ['tests']
```
Sigmanificient, line 9: why excluding tests from the PyPI tarball?

Sigmanificient (changes requested):

pyproject.toml
tintin10q: Sure

pyproject.toml
tintin10q: I thought we would only need the tests when developing.

Sigmanificient (approved):

Sigmanificient:
```patch
diff --git a/tagstudio/__init__.py b/tagstudio/__init__.py
index df6a834..feefebb 100644
--- a/tagstudio/__init__.py
+++ b/tagstudio/__init__.py
@@ -1,3 +1,3 @@
-__version__ = '9.1.1'
+__version__ = "9.1.1"
 
 __all__ = ("__version__",)
diff --git a/tagstudio/tag_studio.py b/tagstudio/tag_studio.py
index 7213914..d557d79 100644
--- a/tagstudio/tag_studio.py
+++ b/tagstudio/tag_studio.py
@@ -5,7 +5,10 @@
 """TagStudio launcher."""
 
 import os, sys
-sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__)))) # add this so that `poetry run tagstudio` works
+
+sys.path.insert(
+    0, os.path.abspath(os.path.join(os.path.dirname(__file__)))
+)  # add this so that `poetry run tagstudio` works
 
 from src.core.ts_core import TagStudioCore
 import argparse
@@ -68,14 +71,17 @@ def main():
     # Driver selection based on parameters.
     if args.ui and args.ui == "qt":
+
         driver = QtDriver(core, args)
         ui_name = "Qt"
     elif args.ui and args.ui == "cli":
         from src.cli.ts_cli import CliDriver
+
         driver = CliDriver(core, args)
         ui_name = "CLI"
     else:
+
         driver = QtDriver(core, args)
         ui_name = "Qt"
```

CyanVoxel:
The version should match the current version of the program: 9.3.2 instead of 9.1.0/9.1.1. This incorrect version is also listed on the PyPI page.

I would also suggest updating the CONTRIBUTING.md with the new instructions for installing Poetry and launching TagStudio via it. I've also been meaning to update the CONTRIBUTING.md to list Python 3.11 as the minimum supported version, which would also need to be set here in `[tool.poetry.dependencies]`.

tintin10q:
Ok I will add instructions to CONTRIBUTING.md.

At the moment it says in pyproject.toml that we only support 3.12. Do you mean you actually (want to) support 3.11 as well?

Do you want me to also publish a version to pypi with this version number?

CyanVoxel:
> At the moment it says in pyproject.toml that we only support 3.12. Do you mean you actually (want to) support 3.11 as well?

Yes, the intention is to support 3.11 as well. 

> Do you want me to also publish a version to pypi with this version number?

I'm honestly not sure about publishing a new version to PyPI since it wasn't my intention to have the program on there in the first place at this time. I suppose if it's going to be there though then it might as well have the correct version number.

tintin10q:
@CyanVoxel ok I made the changes you requested.

pyproject.toml
```diff
@@ -7,3 +7,57 @@ disable_error_code = ["union-attr", "annotation-unchecked", "import-untyped"]
...
+authors = ["Travis Abendshien"]
+license = "GPL-3.0-only"
+repository = "https://github.com/TagStudioDev/TagStudio"
+maintainers = [] # Maybe someone can find a way to export this from GitHub?
+
+[tool.poetry.urls]
+Discord = "https://discord.gg/hRNnVKhF2G"
+
+[tool.poetry.dependencies]
+python = ">=3.11 <3.13"
```
Sigmanificient, line 33: Why not supported python 3.13 out of the box?
```suggestion
python = ">=3.11"
```

pyproject.toml
```diff
@@ -7,3 +7,57 @@ disable_error_code = ["union-attr", "annotation-unchecked", "import-untyped"]
...
+humanfriendly = "10.0"
+opencv_python=">=4.8.0.74,<=4.9.0.80"
+Pillow="10.3.0"
+PySide6="6.7.1"
+PySide6_Addons="6.7.1"
+PySide6_Essentials="6.7.1"
+typing_extensions=">=3.10.0.0,<=4.11.0"
+ujson=">=5.8.0,<=5.9.0"
```
Sigmanificient, line 41:
```suggestion
ujson=">=5.8.0,<=6.0.0"
```

Sigmanificient (changes requested):
Some version constraints are quite restrictive, I think we can let them be a bit more permissive

Will have to update `poetry.lock` accordingly i suppose

pyproject.toml
```diff
@@ -7,3 +7,57 @@ disable_error_code = ["union-attr", "annotation-unchecked", "import-untyped"]
...
+readme = "README.md"
+authors = ["Travis Abendshien"]
+license = "GPL-3.0-only"
+repository = "https://github.com/TagStudioDev/TagStudio"
+maintainers = [] # Maybe someone can find a way to export this from GitHub?
+
+[tool.poetry.urls]
+Discord = "https://discord.gg/hRNnVKhF2G"
+
+[tool.poetry.dependencies]
```
Sigmanificient, line 32: Dependencies should use `-` instead of `_`
eg: `opencv_python`  -> `opencv-python`

Sigmanificient (changes requested):

pyproject.toml
CyanVoxel: We can't guarantee that the program will work on 3.13 until it's fully out and all of the dependencies have been updated. 

One of the first mistakes I made when opening this project was releasing it on 3.9 and assuming it would work on anything later and labeling it as >=3.9, only to get a bunch of reports from people saying it didn't work on 3.12. Then by updating to 3.12 I unknowingly locked myself out of several modules such as [Open3D](https://www.open3d.org/docs/release/getting_started.html) which are still only supported on up to 3.11.

Basically I'd rather play it safe and wait for the dependencies, both the ones currently being used and the ones I'm still waiting to be updated, before supporting 3.13.

pyproject.toml
```diff
@@ -7,3 +7,57 @@ disable_error_code = ["union-attr", "annotation-unchecked", "import-untyped"]
...
+python = ">=3.11 <3.13"
+humanfriendly = "10.0"
+opencv_python=">=4.8.0.74,<=4.9.0.80"
+Pillow="10.3.0"
+PySide6="6.7.1"
+PySide6_Addons="6.7.1"
+PySide6_Essentials="6.7.1"
```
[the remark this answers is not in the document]
CyanVoxel: PySide has given us some trouble with bugs across different versions, I'd rather only keep it on version(s) we know to be stable, such as 6.7.1.

pyproject.toml
tintin10q: No but how I actually wrote it says: Support python version equal to 3.11 and less than 3.13. So that means only 3.11 and 3.12. This is what you asked right?

In fact poetry doesn't even let you install if you say >=3.11 because you get this error:

```
he current project's supported Python range (>=3.11) is not compatible with some of the required packages Python requirement:
  - syrupy requires Python >=3.8.1,<4, so it will not be satisfied for Python >=4

Because tagstudio depends on syrupy (4.6.1) which requires Python >=3.8.1,<4, version solving failed.
```

And I know there is another dependency that actually doesn't support 3.13.

So it was good how it was supporting only 3.11 and 3.12 and your suggestion will not work.

tintin10q: The program actually doesn't work with 3.13. I know there are some dependencies that do not support 3.13 yet.

Sigmanificient (approved):
Considering previous statements, we are going to let deps as is, so this is LGTM

CONTRIBUTING.md
```diff
@@ -24,58 +24,46 @@ Thank you so much for showing interest in contributing to TagStudio! Here are a
...
+### Installing dependencies 
 
-If you wish to launch the source version of TagStudio outside of your IDE:
+After installing poetry and python you can install the dependencies with the following command:
 
-> [!IMPORTANT]
-> Depending on your system, Python may be called `python`, `py`, `python3`, or `py3`. These instructions use the alias `python3` for consistency. You can check to see which alias your system uses and if it's for the correct Python version by typing `python3 --version` (or whichever alias) into your terminal.
+```shell
+poetry install
+```
```
CyanVoxel, line 36: This section should specify where to run this command

CyanVoxel:
While this is also a high priority to get this squared away, I'll be prioritizing #332 over this as that adds additional dependencies and is a preexisting high priority PR

tintin10q:
@CyanVoxel It is up to you but I think it is probably easier to first merge this. This pull request is for the main branch and not sql-bump like #332. After merging this you could merge main into sql bump when ready and just do `poetry add <dependency>` in sql-bump. So merging this does not  affect #332. Additionally, it is not like the virtual environments you already have and use now are broken by merging this pull request you can just keep using them and switch over when you are ready. 

The point of this pull request is not to update the dependencies it is to add poetry, a system to make adding managing dependencies simpler and easier.

CyanVoxel:
> @CyanVoxel It is up to you but I think it is probably easier to just merge this and not connect it to #332. This pull request is for the main branch and not sql-bump like #332. After merging this you could merge main into sql bump when ready and just do `poetry add <dependency>` in sql-bump. So merging this does not affect #332. 


#332 is also targeting `main`; `sql-bump` is the name of yed's branch. #332 has already needed to continuously update for other changes going on in `main` and has been my priority to get shipped before this PR was opened.

> Additionally, it is not like the virtual environments you already have and use now are broken by merging this pull request you can just keep using them and switch over when you are ready. But switching over is not hard. If you want to try this out before merging it just [install poetry](https://python-poetry.org/docs/) and run `poetry install --with dev` and `poetry run tagstudio`. If it works I think this can just be merged. You can actually change your workflow when ready.

While it won't break the existing venvs we're using, it would still put these new Poetry responsibilities on #332 since it's updating the `requirements.txt` for `main`.
Besides the pre-existing priority for #332, merging this first would involve needing @yedpodtrzitko to switch over to the Poetry system mid-PR to update the changes and additions made to the `requirements.txt` and redo them the "Poetry way" for the sake of Poetry. If he chimed in here and said he doesn't mind then I'd be fine pulling this PR first, but for now I'd rather put it on this PR to adapt to the changed reqs over there once #332 gets pulled.  

> The point of this pull request is not to update the dependencies it is to add poetry, a system to make adding managing dependencies simpler and easier.

I personally don't find Poetry necessarily easier easier to work with compared to raw venvs, especially when I've had to install new package managers for a new package just to do what the venv and `requirements.txt` was already doing. I understand the simplicity if you're already used to the system and already have Poetry, but coming from it fresh it's just changed the way that things were being done without necessarily making them easier. That's my two cents, at least. Since this is here though and there's interest in it, plus it's not like creating venvs manually is something *anyone* likes, I still intend on pulling this in time.

#332 and it's predecessors, #190 and #187, have been constantly thwarted by the rest of the codebase updating quicker than they could keep up. The last thing I want to do is give more work to yed right now with a different dependency management system when it can be avoided by just waiting to pull this. Again if he's fine with it then it's all good, but that's my view on it as it stands.

tintin10q:
Alright I understand. Thanks for expanding more on this for me. 

I checked out the latest commit of #332 and if @yedpodtrzitko is fine with merging this pr earlier then here are the commands you would need to issue if #332 were to be merged with the dependencies it has right now:

```shell
poetry add structlog==24.2.0 SQLAlchemy==2.0.32
poetry add ruff==0.6.2 mypy==1.11.1 pytest-qt==4.4.0 --group dev
```

This will update poetry.lock and pyproject.toml correctly. Besides this [and installing poetry](https://python-poetry.org/docs/) you can just ignore it. 
Of course, wait if you want to rather wait!

yedpodtrzitko:
tl;dr, I have just a few questions 

- why Poetry specifically? I'm not that much involved in things, but I'd assume there will be some discussion upfront to make a consensus what is the best/right tool to use. I'm not gonna just randomly bring up other packaging like "you should use pdm/hatch instead!", because they're moreless equal with Poetry, but right now [uv](https://github.com/astral-sh/uv) feels like a way better choice for multiple reasons (speed, fix for #293, easy of installation, virtualenv handling etc. ) 

- if we're doing Poetry (or any tool in general), could you please open parallel PR to my branch #332, so I can merge it there instead of having to deal with the conflicts and "you just need to run this command and this command"? Thanks.

tintin10q:
I looked at uv and it looks quite nice I have not heard of it before. I do not see the publishing feature that poetry has however. The rest of the things you mention, easy of installation, virtualenv handling etc I think are all the same. But uv claims to be a bit faster indeed. As for ease of install Poetry just recommends to install with pipx but there are also separate install scripts or you can install it with pip so I think that is the same as well. 
Given the similarities I think it matters more that there is a pull request to add poetry and not one to add uv. 

I will make a pr to your branch.

tintin10q:
https://github.com/yedpodtrzitko/TagStudio/pull/4

yedpodtrzitko:
> I looked at uv and it looks quite nice I have not heard of it before. I do not see the publishing feature that poetry has however.


I dont want to disregard the effort you put into this, but why does PyPI publishing matter? Normal users will download the executable package. I dont see the use-case for installing python, setting up virtualenv just so I can install it from PyPI and run in awkward way from shell... afaik nobody runs (nor publish) desktop apps this way. 

So if the reason to choose Poetry (over uv) is because it has a feature which is not necessary / relevant for the project (as @CyanVoxel mentioned his stance above), then is that still the best option?

tintin10q:
I don't think PyPi publishing is particularly important in this case I just mentioned it as something uv doesn't have. I understand that most normal people will just download an executable. 

I did a rough test using the time command.

Running `poetry install` on a fresh project took:

```
real	0m18,636s
user	0m18,127s
sys	0m1,521s
```

Running `uv pip sync requirements.txt` took:

```
real	0m18,684s
user	0m5,166s
sys	0m2,593s
```

Its basically the same and I gave uv leeway because creating the venv is a seperate command in uv (not included in the time) but with poetry that was also included in the time. 

When running poetry install again uv is faster to realize nothing needs to happen:

```
```

tintin10q:
I don't think PyPi publishing is particularly important in this case I just mentioned it as something uv doesn't have. I understand that most normal people will just download an executable. 

I did a rough test using the time command.

Running `poetry install` on a fresh project took:

```
real	0m18,636s
user	0m18,127s
sys	0m1,521s
```

Running `uv pip sync requirements.txt` took:

```
real	0m18,684s
user	0m5,166s
sys	0m2,593s
```

Its basically the same and I gave uv leeway here because creating the venv is a seperate command in uv (not included in the time) but with poetry that was also included in the time. 

Running poetry install again uv when nothing needs to happen:

```
$ time uv pip sync requirements.txt

real	0m0,039s
user	0m0,032s
sys	0m0,046s
``` 

```
$ time poetry instal

real	0m0,995s
user	0m0,917s
sys	0m0,078s
```

Adding a dependency:

```
🯆 time poetry add tqdm

real	0m2,273s
user	0m1,929s
sys	0m0,114s
```

```
time uv add tqdm 

real	0m0,215s
user	0m0,101s
sys	0m0,036s
```

The initial install which is the most important has no difference.  Besides that uv is unsurprisingly a bit faster.
Fast enough to disregard the work here and redo it? That is for @CyanVoxel to decide. 

I personally think poetry is good and I also like the cli interface more. But I am of course biased. Also do not forget that other people also mentioned poetry independently before I made this. But of course they might not have known about uv either.

CONTRIBUTING.md
```diff
@@ -24,58 +24,48 @@ Thank you so much for showing interest in contributing to TagStudio! Here are a
...
-- [Mypy](https://github.com/python/mypy) (Included in `requirements-dev.txt`)
-- [PyTest](https://docs.pytest.org) (Included in `requirements-dev.txt`)
+- [Python](https://www.python.org/downloads/) 3.11 or 3.12
+- [Poetry](https://python-poetry.org/docs/)
 
-### Creating a Python Virtual Environment
+### Installing dependencies 
 
-If you wish to launch the source version of TagStudio outside of your IDE:
+After installing poetry and python you can install the dependencies with the following command:
```
eivl, line 32: Im in favor of sharing one way to install a tool if there is consensus to use it. I use poetry daily, there are many things I don't like about it, but it makes the interface we use the same for all developers, no matter what platform you are on.

Currently the recommended way to install poetry is by using `pipx` this is a tool that is recommended for python devs to have, but it is yet another install process to document. 

Im happy write the changes needed if that is something you would like me to do.

CONTRIBUTING.md
```diff
@@ -155,9 +145,30 @@ Most of the style guidelines can be checked, fixed, and enforced via Ruff. Older
...
+
+If you want to add a dev dependency instead use: `poetry add <name> --group dev`. 
+To regenerate the requirements-dev.txt use: 
+
+```shell
+poetry export --without-hashes --without main --with dev --format=requirements.txt > requirements-dev.txt
+```
+
+Do not forget to actually commit the updated requirement.txt and pyproject.toml and poetry.lock files when you change/add dependencies.
+
```
eivl, line 168: There is an important detail missing from this MD file. 

The explanation of what the lock file is. The troublesome command `poetry update` and en explicit warning that this command updates packages and creates a new .lock file. If the PR is updating dependencies, sure, you would use this and update the .lock file.  

When you add a new dependencies, poetry will use the caret symbol, I personally dislike this behaviour as it is implicit in nature. Are future dependencies going to follow the pinned nature of the repo? I don't know, but poetry will update the pyproject.toml file as it sees best. This might lead to trouble  and worth a note in the readme.

pyproject.toml
```diff
@@ -7,3 +7,57 @@ disable_error_code = ["union-attr", "annotation-unchecked", "import-untyped"]
...
+maintainers = [] # Maybe someone can find a way to export this from GitHub?
+
+[tool.poetry.urls]
+Discord = "https://discord.gg/hRNnVKhF2G"
+
+[tool.poetry.dependencies]
+python = ">=3.11 <3.13"
+humanfriendly = "10.0"
+opencv_python=">=4.8.0.74,<=4.9.0.80"
+Pillow="10.3.0"
```
eivl, line 36: This is a comment for the rest of the pinned versions and this does not reflect a change I want to enforce, but an honest comment. 

`pyproject.toml` in a library should focus on getting the newest version of a package that it can. This means that you should not pin versions here. 

`requirements.txt` is the place to pin a version, as this file that the application should use. this is the file dependabot will use to check for security issues and in another scenario, this is the file that docker would use as well. 

Here lies one of the issues with poetry, it is not 100% compatible with the python ecosystem, and you cant use a constraint file to enforce application versions during install. It will follow the lock file. 

Micromanaging the pyproject toml files pinned versions is a little hassle, but it can ofcourse be done, this is a cost someone has to take; I'm in favor if you can. I use poetry at work so that all developers can have the same experience, at the cost of some extra work for one. 

TL;DR
I don't like pinned versions in pyproject.toml, sometimes that is the only way until poetry complies with pip.

tagstudio/tag_studio.py
```diff
@@ -4,9 +4,14 @@
 
 """TagStudio launcher."""
 
+import os
+import sys
+
+sys.path.insert(
+    0, os.path.abspath(os.path.join(os.path.dirname(__file__)))
+)  # add this so that `poetry run tagstudio` works
```
eivl, line 12: Didn't you add an entrypoint console script? I'm not sure if this is a solution to the problem or just a solution to a xy problem. Can you elaborate about this please?

tagstudio/tag_studio.py
```diff
@@ -4,9 +4,14 @@
...
+
 from src.core.ts_core import TagStudioCore
-from src.cli.ts_cli import CliDriver  # type: ignore
-from src.qt.ts_qt import QtDriver
```
eivl, line 9: Is the added complicated code worth the few ms saved from doing both imports? I would argue that it is not worth it unless there is a very good reason for it. It leads to troublesome code to debug in the future.

eivl (changes requested):
When my comments are resolved I can fully approve this pr.

tagstudio/tag_studio.py
zierf: Since if and else are the same, it could also be simplified.

```Python
    # Driver selection based on parameters.
    if args.ui and args.ui == "cli":
        from src.cli.ts_cli import CliDriver

        driver = CliDriver(core, args)
        ui_name = "CLI"
    else:

        driver = QtDriver(core, args)
        ui_name = "Qt"
```

zierf:
Looking forward to this, _Poetry_ is also currently the best tool for packaging Python into Nix packages. In my attempts I also came across a couple of things and I have also gone over this pull request a bit. Maybe I can share one or two things here.

The [_PySide6_ Python package](https://pypi.org/project/PySide6/) officially states:
"_Please note: this wheel is an alias to other two wheels PySide6_Essentials and PySide6_Addons, which contains a predefined list of Qt Modules._"

I saw something similar again in an info about the package split into Essentials and Addons, but I can't find it anymore.
```TOML
PySide6 = "6.7.2"
# PySide6 includes PySide6_Essentials and PySide6_Addons
# PySide6_Addons= "6.7.2"
# PySide6_Essentials = "6.7.2"
```

`__init__.py`
```Python
# allow local module files as import
# https://github.com/python-poetry/poetry/issues/3928#issuecomment-1399313433
import os, sys; sys.path.append(os.path.dirname(os.path.realpath(__file__)))
```

`pyproject.toml`
```TOML
packages = [
  { include = "tagstudio", from = "." },
]
```

```Python
warning: Git tree '/home/zierf/workspaces/Python/TagStudio' is dirty
  File "/nix/store/k98sinf1gjvbyld65nmqm5zly4lyggrw-python3.12-tagstudio-9.3.2/bin/..tagstudio-wrapped-wrapped", line 6, in <module>
    from tagstudio.tag_studio import main
  File "/nix/store/k98sinf1gjvbyld65nmqm5zly4lyggrw-python3.12-tagstudio-9.3.2/lib/python3.12/site-packages/tagstudio/tag_studio.py", line 8, in <module>
    from .src.cli.ts_cli import CliDriver  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/k98sinf1gjvbyld65nmqm5zly4lyggrw-python3.12-tagstudio-9.3.2/lib/python3.12/site-packages/tagstudio/src/cli/ts_cli.py", line 32, in <module>
    from ..qt.helpers.file_opener import open_file
  File "/nix/store/k98sinf1gjvbyld65nmqm5zly4lyggrw-python3.12-tagstudio-9.3.2/lib/python3.12/site-packages/tagstudio/src/qt/helpers/file_opener.py", line 13, in <module>
    from PySide6.QtWidgets import QLabel
ModuleNotFoundError: No module named 'PySide6.QtWidgets'
```

tagstudio/tag_studio.py
zierf: See my [following comment](https://github.com/TagStudioDev/TagStudio/pull/368#issuecomment-2323405866) or [ModuleNotFoundError for poetry script](https://github.com/python-poetry/poetry/issues/3928#issuecomment-1399313433) for more information.

pyproject.toml
zierf: > `pyproject.toml` in a library should focus on getting the newest version of a package that it can. This means that you should not pin versions here.

I think it's exact the opposite. Defining dependencies too loosely seems to lead to unforeseen problems and non-reproducible builds, especially in Python and Javascript projects. Ultimately, that is what the lock file is trying to avoid. If necessary, it can be updated quickly at any time.

> `requirements.txt` is the place to pin a version, as this file that the application should use. this is the file dependabot will use to check for security issues and in another scenario, this is the file that docker would use as well.

Just an idea, but maybe it's possible to use a git hook to generate a `requirements.txt` from the `pyproject.toml`.
(or can an automatic build system generate a temporary one and call dependabot?)

CONTRIBUTING.md
tintin10q: It is not clear to me what changes you are actually suggesting here. @eivl

CONTRIBUTING.md
tintin10q: The section you comment on explains how to add dependencies. The reader does not need to know what the `poetry.lock` file is because it is updated correctly when they run `poetry add <dependency>` as described. No need to run `poetry update` either. I don't understand why you bring up `poetry update` because I did not tell the reader to run this command. 

I could add a sentence saying that it is better to pin a specific version of a dependency `poetry add package==0.1.0`?

pyproject.toml
tintin10q: I disagree that pyproject.toml should focus on getting the newest version of a package. The important thing is that the project works, not that the latest version is downloaded. From time to time a pr can bump versions after testing. Also in your other comment you say you actually don't like specifying `^` versions. 

Why would you pin versions in requirements.txt and not in pyproject.toml? That just does not make any sense.  The versions should be the same and that is why I explained in the CONTRIBUTING.md  how to generate requirements.txt from pyproject.toml.

If it was possible I would not even have a requirements.txt but we have to have it of course.

tintin10q: I think having a git hook to update requirement.txt from the pyproject.toml is a good idea. I also thought of that but I have never set up git hooks before.

tagstudio/tag_studio.py
tintin10q: I did add an entry point to the script. But this project uses . relative imports so you need to fix that using this. Of course I could not have moved away from these relative imports. You can also have this problem when adding an extra py script in the root it is about where the working directory is. Something you can not change in the entry point definition either. 

I think adding the module to the import path is an elegant solution. This way it will just always work, not just for poetry.

tagstudio/tag_studio.py
tintin10q: I disagree that this is more complicated. Do you never import things lazily? It is a great feature.  
I did it like this because this allows the CLI code which does not work anyways to not be present in the package.

The cli driver can now not in the packages key in the `pyproject.toml`

tintin10q:
Thanks for your comments!

As for specific versions, I created this pr to add poetry not to change versions of dependencies. I suggest you make a new pr to change the PySide6 dependency. But probably after this has merged.

I don't really see the problem of having multiple __init__.py. But moving the code that adds the import path there is a good idea!

Iirc @CyanVoxel mentioned he did not want to do local imports or prepend the package name. It also says it here: https://github.com/TagStudioDev/TagStudio/blob/main/CONTRIBUTING.md#implementations

I agree though having dots or module name imports would be better. I would suggest using the dots. Al tough it does look a little weird if you see it for the first time. But in my opinion: yes it would be better to change the imports.
But again, that is not the point of this pr. I just want to get poetry working. Another pr can try to change the imports. 
I didn't want to do this because it would mean that this pr would touch much more files.

tintin10q:
> It is therefore worth trying out whether the packaged application also works on another system without a QT environment.

But without QT it won't work anyways right?

jack-mil:
Just wanted to chime in as a python developer that has built and maintained several python packages and applications. A pyproject.toml should definitely be the place to list dependencies, as it's the current standard for all python packages. This is preferred over a requirements.txt and manual invocation of pip. Even if end users download the binary releases as a Windows desktop application and this is never published to pypi,  for developers it's good to have a python package definition.

Poetry can be useful, but setuptools is also a perfectly capable build system, and can be simpler and easier to use for some projects. It also has the advantage that developers aren't forced to use a specific tool. They can choose to `pip install -e .` into any virtual environment of their choice (managed by their editor/ide, uv, venv,  virtualenv, pipx, etc, etc). 

Here is the guide on setuptools
https://setuptools.pypa.io/en/latest/userguide/quickstart.html

pyproject.toml
zierf: The Poetry project seems to provide some git hooks.

E.g. I have seen a [poetry-export git hook](https://python-poetry.org/docs/pre-commit-hooks/#poetry-export).

But it requires to install a poetry plugin.
Not sure if it is suitable at all or even working on NixOS systems.

I don't really have any experience with this, there may be better tools out there for creating _requirements.txt_ files.

Maybe a temporary _requirements.txt_ during the build would be sufficient, which would then be used for testing while in CI/CD.

Or perhaps there are better alternatives to `dependabot` that would work without any problems with just a `pyproject.toml`, the industry seems to shift towards it anyway. Unless the _requirements.txt_ is absolutely needed for something.

CONTRIBUTING.md
eivl: `pipx install poetry` is the command to install poetry. the docs has multiple other ways to install it, but I'm in favor of front loading the information instead of back loading it with more links to click on.

CONTRIBUTING.md
eivl: It is a thing I have seen many times in projects that start using poetry. I don't mind the sections that are already added. Im saying I think this detail is missing.  It has burned many people in projects I have worked on the past years.

tintin10q: Can you suggest a text? Like there is not more to it then the poetry.lock pins the versions of each dependency so you get reproducible builds.

pyproject.toml
tintin10q: Cool. But this could also be another pr?

tintin10q:
I would also opt for not having a requirement.txt at all if that was agreed upon. But I like poetry because who likes making virtual enviroments.

seakrueger:
> I would also opt for not having a requirement.txt at all if that was agreed upon. But I like poetry because who likes making virtual enviroments.

I would prefer to keep the requirement*.txt files. I personally don't see the need for an external dependency manager, and would rather keep using virtual envs as it works better for my workflows.

tintin10q:
https://xkcd.com/1172/

But yeah no reason to get rid of the requirements.txt, but a dependency manager that makes the venv and everything for you is also nice.

Sapein:
Why keep the requirements.txt around? You don't need it anymore even for setuptools, which has support for using pyproject.toml for all dependencies.

zierf:
Will this change be implemented? This pull request actually seemed to be finished already. According to the checks, there is only something left in the linter (the logs have already been cleaned up).

If this is merged, [Issue #200](https://github.com/TagStudioDev/TagStudio/issues/200) could also be implemented. At least after I have incorporated the more than 238 commits that have been added in the meantime into my fork. 😁

Without this PR and _Poetry_, _poetry2nix_ will not be able to create an executable version for NixOS.

tintin10q:
I can rebase and run the linter.

tintin10q:
Ok I merged main and updated the dependencies to the current ones. I fixed the formatting thing. `poetry run tagstudio` works for me so it can now be merged.

xarvex:
I've been reading this thread up and down, and there doesn't seem to be a compelling reason to use Poetry. PyPi publishing does not matter for us as an application, and using it for the sake of Nix is not enough to warrant an overhaul that affects everyone. Even with that aside, poetry2nix has been in an unmaintained status since November 2024. There are some people doing the occasional pull, but it is not a project I would rely upon; it seems to be treated as something receiving support for the sake of eventually transitioning *away*.

If we are to switch dependency manager in the future, it will most likely be to uv, stated above, in which I have done some tinkering with uv2nix as an experiment (even though Nix support shouldn't be the reason, it is being considered).
**Please, do not take this as guidance to make a PR adding in uv.** Simply put, there are better things to address at the moment for the project, and a change in package manager would be done at a “quieter” time. When the time comes, I won't mind putting motion into it.

Last note about Nix: I have been revamping the flake and modules as of recently, currently with a shell using a virtualenv, redone with some old dependencies trimmed away. My current task now is packaging TagStudio and the relevant Python dependencies. I will make an issue to track all of this in the morning, so some scattered Nix issues can be “bundled” in one place.

tintin10q:
I dont know or really care about nix.  I fail to see how merging this pr would affect everyone. If you want to keep using requirements.txt then you can. By merging this pr you just give devs the option to use poetry.

tintin10q:
Also I think pypi publising does matter. Why wouldn't you make tagstudio pip installable? It's already setup and it would make tagstudio available to everyone with pip installed which is a lot of people. The name on pypi has been claimed as I was the one who orginally claimed it for CyanVoxel.

CyanVoxel:
#844, which has just been merged, has just significantly refactored the project layout to better align with the "src-layout" structure that was intended from the beginning of this project. Along with this layout change has come several other project-level changes that have been made easier thanks to this corrected layout.

This includes moving all dependencies to the `pyproject.toml` file and removing the `requirements.txt` and `requirements-dev.txt` files. This was done in a dependency manager agnostic way, which was one of my biggest issues with the approach here. I did not want the overhead of having to maintain separate ways of installing packages that included the old method plus a poetry-specific method with all of the numerous [tool.poetry] entries in the `pyproject.toml`. I did not wish to use Poetry as I found it more cumbersome to setup and interface with than a bare venv or an alternative dependency manager such as uv. #844 however circumvents all of this by allowing the dependencies to be declared once and are made installable with either a classic venv + pip approach, uv, or [Poetry 2.0](https://python-poetry.org/blog/announcing-poetry-2.0.0/) since that release fixes a lot of my gripes with the large amount of specialized configuration. Now that Poetry supports PEP 621 I believe everyone should be able to use whichever dependency management solution they prefer without any undue specialized maintenance required on my part.

I believe these changes satisfy the original request in #365 and render this PR as no longer necessary. Most of, if not all the goals set out by this PR should now be accomplishable via the changes in #844. For reasons mentioned previously and reiterated now, I do not plan on supporting Poetry 1.x as it would be too cumbersome and increasingly unnecessary as people upgrade to 2.0. PyPI may be explored in the future, but is not something we're sure of or eager to do at the moment.

Lastly, the [installation documentation](https://docs.tagstud.io/install/) has been greatly revamped and the setup instructions that were previously in the CONTRIBUTING.md have now been now mostly transferred over to the docs site. This includes instructions for various dependency managers, including Poetry.

Please let me know if everything has been addressed and this PR can be safely closed. If you have any requests that weren't addressed by either #844 or related to my comments about increased maintenance overhead, then they are probably best served in a new feature request. If you notice any mistakes in our updated documentation, especially related to Poetry, then you are more than welcome to submit a new PR with any corrections.

zierf:
> PyPi publishing does not matter for us as an application, and using it for the sake of Nix is not enough to warrant an overhaul that affects everyone.

This PR existed before Nix could be packaged. At the same time, _Poetry2Nix_ seemed to be a good basis for packaging under Nix. It would have just fit together very well.

At the same time, it would have also solved Issue #200 and #512 directly. _Poetry2Nix_ would also have made it possible to not use the rigid pre-built Nix packages, but also to use the packages defined in the `pyproject.toml` file.

> Even with that aside, poetry2nix has been in an unmaintained status since November 2024. 

Interesting, good to know, thanks for the info. At that time, it was not yet outdated, a suitable PR could have been launched in October. That's why I asked the question here whether this will be merged or discarded.

> #844, which has just been merged, has just significantly refactored the project layout to better align with the "src-layout" structure that was intended from the beginning of this project. […] This includes moving all dependencies to the `pyproject.toml` file and removing the `requirements.txt` and `requirements-dev.txt` files. This was done in a dependency manager agnostic way, which was one of my biggest issues with the approach here.

Sounds good.

> I believe these changes satisfy the original request in #365 and render this PR as no longer necessary. Most of, if not all the goals set out by this PR should now be accomplishable via the changes in #844.

What was missing for users was a simple way to install the program quickly and easily via their config or to run a flake directly from GitHub via the CLI.

> For reasons mentioned previously and reiterated now, I do not plan on supporting Poetry 1.x as it would be too cumbersome and increasingly unnecessary as people upgrade to 2.0.

I didn't know that changes to the Flake had already been planned/implemented and I finally wanted to provide a way to install/run the program directly. Depending on this PR, that would have been either with _Poetry_ or I would have had to look for an alternative. But since the Flake has already been revised, that at least means less work for me. 😀

> Lastly, the [installation documentation](https://docs.tagstud.io/install/) has been greatly revamped and the setup instructions that were previously in the CONTRIBUTING.md have now been now mostly transferred over to the docs site. 

I think the documentation turned out really well! Big compliment for that. 👍

> Please let me know if everything has been addressed and this PR can be safely closed.

When I took a quick look, I didn't find any serious errors. The `.desktop` file is unfortunately still missing in the flake, but I think I'll address that directly in issues #415 and #847.

When I run it via the CLI, I still get a _Path does not exist._ error (penultimate line), but that may be unrelated to Nix.

```SH
$> tagstudio

2025-03-11 21:59:44 [info     ] [FFMPEG] Using FFprobe location: ffprobe
2025-03-11 21:59:44 [info     ] [FFMPEG] Using FFmpeg location: ffmpeg
2025-03-11 21:59:44 [info     ] [ResourceManager] Resources Registered: count=27
2025-03-11 21:59:44 [info     ] [Config] Config File not specified, using default one filename=~/.config/TagStudio/TagStudio.ini
2025-03-11 21:59:44 [info     ] [Config] Thumbnail cache size limit: 500 MB
qt.multimedia.ffmpeg: Using Qt multimedia with FFmpeg version 7.1 GPL version 3 or later
qt.multimedia.ffmpeg: Available HW decoding frameworks:
qt.multimedia.ffmpeg:      vaapi
qt.multimedia.ffmpeg:      vdpau
qt.multimedia.ffmpeg:      vulkan
qt.multimedia.ffmpeg: Available HW encoding frameworks:
qt.multimedia.ffmpeg:      vaapi
qt.multimedia.ffmpeg:      vdpau
qt.multimedia.ffmpeg:      vulkan
2025-03-11 21:59:46 [error    ] Path does not exist.           open_path=None
2025-03-11 21:59:46 [info     ] FFmpeg found: {self.ffmpeg}, FFprobe found: {self.ffprobe}
```

xarvex:
> > PyPi publishing does not matter for us as an application, and using it for the sake of Nix is not enough to warrant an overhaul that affects everyone.
> 
> This PR existed before Nix could be packaged. At the same time, _Poetry2Nix_ seemed to be a good basis for packaging under Nix. It would have just fit together very well.
> 
> At the same time, it would have also solved Issue #200 and #512 directly. _Poetry2Nix_ would also have made it possible to not use the rigid pre-built Nix packages, but also to use the packages defined in the `pyproject.toml` file.

Not specific to poetry2nix but rather Poetry as a whole that CyanVoxel mentioned, but I forgot to bring up, is that very thing around wanting the development workflow around here to be as agnostic as possible. Contributors should be able to use whatever tools they prefer. Poetry 1.X had the problem of needing dependencies stated its own way, which puts us in a spot of duplicating entries (where there is potential for drift), or “choosing” its methods.

As for the Nix equation, it would be nice to use projects defined in the `pyproject.toml`, I agree. I used poetry2nix in one of my own projects to sort of tinker with, and when it was unmaintained, switched its implementation to [uv2nix](https://pyproject-nix.github.io/uv2nix). There is also [pyproject.nix](https://pyproject-nix.github.io/pyproject.nix), something uv2nix uses under the hood, that I considered using directly as well. Ultimately, I went with the Nix packages because they provide some benefits with their rigidity. In terms of deduplication, if a Nix user use programs in their configuration alongside TagStudio that might share dependencies, that one package is “shared” in the Nix store. It also means the nixpkgs cache is used. Lastly, TagStudio is something I wish to upstream, and that wouldn't be possible if providing a package with one of these schemas.

> > Even with that aside, poetry2nix has been in an unmaintained status since November 2024.
> 
> Interesting, good to know, thanks for the info. At that time, it was not yet outdated, a suitable PR could have been launched in October. That's why I asked the question here whether this will be merged or discarded.

You are correct in that the status of maintenance was not relevant at the time in this PR, my answer was only in the state of things now when it was bumped. Ultimately, there were many other things going on in the project that a change like this fell through the cracks. Apologies on that front.

> I didn't know that changes to the Flake had already been planned/implemented and I finally wanted to provide a way to install/run the program directly. Depending on this PR, that would have been either with _Poetry_ or I would have had to look for an alternative. But since the Flake has already been revised, that at least means less work for me. 😀

There were some rough plans I had with the flake for a long time in terms of improving it. At first, it was experience: my early work on the flake was just getting *anything* functional at all. With university coming in full swing and some of my own side projects popping up, the problem of experience turned into a problem of time constraint. The whole project refactor to make it better compatible with pip, and therefore the Nix builders and tooling that use pip, was also a requirement. All to say, much of what ended up happening in the end was a little spur of the moment once the state of things was fully realized. 

> When I took a quick look, I didn't find any serious errors. The `.desktop` file is unfortunately still missing in the flake, but I think I'll address that directly in issues #415 and #847.

Yes, that still needs to be done. I left that out, as I believe it should be provided in the repository itself. I skimmed through my notifications, but I believe you left a comment on the issue agreeing with that.

> When I run it via the CLI, I still get a _Path does not exist._ error (penultimate line), but that may be unrelated to Nix.
> 
> ```shell
> $> tagstudio
> 
> 2025-03-11 21:59:44 [info     ] [FFMPEG] Using FFprobe location: ffprobe
> 2025-03-11 21:59:44 [info     ] [FFMPEG] Using FFmpeg location: ffmpeg
> 2025-03-11 21:59:44 [info     ] [ResourceManager] Resources Registered: count=27
> 2025-03-11 21:59:44 [info     ] [Config] Config File not specified, using default one filename=~/.config/TagStudio/TagStudio.ini
> 2025-03-11 21:59:44 [info     ] [Config] Thumbnail cache size limit: 500 MB
> qt.multimedia.ffmpeg: Using Qt multimedia with FFmpeg version 7.1 GPL version 3 or later
> qt.multimedia.ffmpeg: Available HW decoding frameworks:
> qt.multimedia.ffmpeg:      vaapi
> qt.multimedia.ffmpeg:      vdpau
> qt.multimedia.ffmpeg:      vulkan
> qt.multimedia.ffmpeg: Available HW encoding frameworks:
> qt.multimedia.ffmpeg:      vaapi
> qt.multimedia.ffmpeg:      vdpau
> qt.multimedia.ffmpeg:      vulkan
> 2025-03-11 21:59:46 [error    ] Path does not exist.           open_path=None
> 2025-03-11 21:59:46 [info     ] FFmpeg found: {self.ffmpeg}, FFprobe found: {self.ffprobe}
> ```

I've gotten this too with both the devshell and the package, need to investigate it more to know where it is coming from and what the cause is; whether it is related to Nix.

I hope that provides some answers and context for you!

PR #372, opened 2024-08-23
feat: Support custom user colors in tags

### LINKED ISSUE #264: [Feature Request] Hex/RGB Color Values for Tags
I've just started using TagStudio and noticed that Tags have hardcoded colour values, for example Red or Blue instead of #FF0000 or #0000FF. Having the option to use Hex or RGB values for Tag Colouring would give a lot of additional flexibility (and it would make the tags even prettier!).

I would suggest keeping the default colour values but also adding a colour ring and a text field for putting in RGB/Hex values.

Also, I looked at other issues to see if someone asked for something like this before and I don't know if I was supposed to use some kind of template, sorry if I had to, I didn't see any!

❤️

seakrueger:
# Hex-based colors

In order to support custom colors, tags must be moved away from the fixed name color system and are instead transitioned to a single primary hex value. All other accents are now up to the front-end to generate (if they are necessary). This allows for the removal of the large tag color dictionary, though the key and primary color hex remain for legacy purporses. A check is added to library loading to replace old fixed-named colors with their primary hex key. 

# Custom colors

QT provides an os-native color picker dialog complete with a full color spectrum, default colors, saving of colors, and an eye-dropper utility. The old color dropdown has been replaced with a push button that opens the color dialog modal and updates to reflect the selected color. The QColor class is also used for generating accent colors as it provides simple helper functions to create off-set colors.

Example gif
closes #264


seakrueger:
Marked as draft until #332 is merged, and I can update the stuff around it

CyanVoxel:
We discussed it in the Discord already, but I'll repost my take on this here to keep everything in one place:

[...] I've intended custom tag colors to still have a name associated with them, so they would be proper objects with an `id`, `name`, and `hex_value` with them, then the `id` gets referenced by the tag. This allows you to update a color value and have that reflected on all of the tags that already use that color, instead of having to go in and manually change each one individually. It also would enable you to search by your named colors among and opens up the door for better color organization

seakrueger:
Very outdated and sounded like Cyan is tackling. May open new PR if I get around to updating

Issue #374, opened 2024-08-23
[Feature Request]: Thumbnails for PDFs


Heiholf:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

Render first page of PDF as thumbnail.

### Solution

I can implement it myself, I just wanted to have a proper Issue so that the PR is not completely random.

Qt has PDF rendering functions, I would use those to prevent adding additional dependencies. 

### Alternatives

There might be faster PDF engines than the builtin in Qt. However, the speedup won't be gigantic and it is easier to use less dependencies.

CyanVoxel:
PDF previews added in #543 (port of #378)

PR #378, opened 2024-08-23
Added thumbnail rendering for PDF-files

Heiholf:
Adds thumbnail support for PDF files as proposed in #374. Uses Qt's inbuilt PDF renderer.

One thing which is kind of weird is the arbitrary scaling of the image with 2.5 but it turns this
Factor_1 into this Factor_2p5
and makes the preview more readable (I think the 6.25x work increase is worth it).

I tried checking if the PDF is password protected. However, the docs don't have any info for the function `passwordRequired()` and it always returns true.

I tested it on [these PDF example files](https://github.com/py-pdf/sample-files). The majority work properly but some don't:
- Some ImageMagick Images
- PDFs with unreadable metadata
- PDFs which even Okular can't render


tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -746,6 +757,46 @@ def _video_thumb(self, filepath: Path) -> Image.Image:
...
+        document.load(file)
+        # Transform page_size in points to pixels with proper aspect ratio
+        page_size: QSizeF = document.pagePointSize(0)
+        ratio_hw: float = page_size.height() / page_size.width()
+        if ratio_hw >= 1:
+            page_size *= size / page_size.height()
+        else:
+            page_size *= size / page_size.width()
+        # Enlarge image to improve image downscaling (kind of arbitrary)
+        page_size *= 2.5
```
eivl, line 780: Use a good variable name instead; magic numbers like this will be confusing in the future. 
```suggestion
        ENLARGE_SCALE_FACTOR = 2.5
        page_size *= ENLARGE_SCALE_FACTOR 
```

The suggestion is not here for you to use as a drop-in replacement. You should save this to the class and, possibly in the future (or right now), move it to a constant namespace.

tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -746,6 +757,46 @@ def _video_thumb(self, filepath: Path) -> Image.Image:
...
+            | QPdfDocumentRenderOptions.RenderFlag.ImageAliased
+            | QPdfDocumentRenderOptions.RenderFlag.PathAliased
+        )
+        # Convert QImage to PIL Image
+        rendered_qimage: QImage = document.render(0, page_size.toSize(), render_options)
+        buffer: QBuffer = QBuffer()
+        buffer.open(QBuffer.OpenModeFlag.ReadWrite)
+        rendered_qimage.save(buffer, "PNG")
+        pil_image: Image.Image = Image.open(BytesIO(buffer.buffer().data()))
+        buffer.close()
```
eivl, line 794: can you use a context manager instead? 
I don't know `QBuffer` so I cant make an inline suggestion, I'm open to research this and help you out.

tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -746,6 +757,46 @@ def _video_thumb(self, filepath: Path) -> Image.Image:
...
+        # Replace transparent pixels with white (otherwise Background defaults to transparent)
+        pixel_array = np.asarray(pil_image.convert("RGBA")).copy()
+        pixel_array[pixel_array[:, :, 3] == 0] = [255, 255, 255, 255]
```
eivl, line 797: this sounds like something very usefull as its own method. would be useful for others in the future. Can you add this as its own method and just call it here?

eivl (changes requested):
I request that changes be made. the implementation is ofc open to you, but ask me for input if you want to and you will get it!

tagstudio/src/qt/widgets/thumb_renderer.py
Heiholf: I am considering creating a utils file for images. But I'm not totally sure how the whole project is laid out.

tagstudio/src/qt/widgets/thumb_renderer.py
Heiholf: AFAIK Qt for Python is directly generate from the C++ code without adding python-like features. I could add a context manager for the buffer but I do not think that is a good idea and would generate too much unnecessary code.
However, I will definitely put it in a "fake context manager" with try-except to close the buffer even with errors occuring.

CyanVoxel:
I'm moved this PR from the deprecated `thumbnails` branch to `Alpha-v9.4`, which is the branch that `thumbnails` was eventually merged into. However the v9.4.x releases are now feature frozen and are now solely reserved for bugfixes. This means that after the changes from `thumbnails` are ported to `main` that I'll once again be moving the target for this PR over to there as a v9.5 feature.

CyanVoxel:
I've ported this functionality to main via #543, thank you for your contribution with this!

PR #381, opened 2024-08-23
Add setup program

Za9-118:
This is yet to be complete, i still have to add it so the pyport.zip (Python3.12) file gets downloaded from the internet as Github cant upload 75MB long file, and the app being compiled, i will do that later since im tired.


Za9-118:
Also the software.zip is the full main branch from TagStudio

CyanVoxel:
I'm not sure the program really needs an installer in it's current state... Plus we use a GitHub Actions release workflow that we trigger to compile new builds, and this doesn't interface with that at all and is just zipping up the source code along with expecting a separate installation of Python 3.12 to be downloaded. Also for future reference, we don't ship releases from the `main` branch and instead use the versioned `Alpha-v9.3`, `Alpha-v9.4`, etc branches.

Za9-118:
Oh ok, i got too deep in the project when i realized you compiled it, Silly me

Za9-118:
Atleast it will be kinda useful for me to use as a template for projects i make if need it.

PR #384, opened 2024-08-24
Complete Readme Revamp

AdamTmHun:
Complete Readme Revamp
- New widgets/shields
- New layout
- New Content Menu
- Clickable stuff
- New menus
- 🐣


eivl:
fix conflicts and tag me for review @AdamTmHun

xarvex:
I don't think there's any interesting in having a rather cookie-cut template for the README, that while I'm sure had the goal of catching attention, actually discards. As potentially the first document a user might read, it is not something that should appear closely identical to several other projects out there in boring and stale conformity.
Also, not to be “that guy” but while there is fun and jokes to be had, there is some objectively bad taste present here that is not fit for the potential first impression I brought up.
Outside of that, we have a documentation site now, where some finer details can go that have trouble being condensed to one view: https://docs.tagstud.io/

Issue #360, opened 2024-08-22
[Feature Request]: EPUB/MOBI/AK3/etc Thumbnails (Ebook extensions)


altroquinine:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

Hi, I recently downloaded Tagstudio and wanted to say that I really like it so far! I saw the original video and the one that just came out, and when I saw that the installation wasn't confusing anymore (I don't know anything about python) I decided to give it a go. I noticed that I have a lot of ebooks in my library, and I often use Calibre to manage them and transfer to my kindle.

### Solution

I'm going to preface this by saying I have no programming or coding experience outside of some rudimentary C# skills. Apologies if I'm completely off base here or using the wrong terminology, hopefully this makes sense!

I was wondering if it'd be possible for Tagstudio to use the cover of the book Calibre pulls from (I think it's just a .jpg file) as a thumbnail instead of it just saying epub? I know there's other programs for managing ebooks, but as far as I know Calibre is the most popular and I'm not really sure if the metadata in the ebook file itself contains a cover image. 

### Alternatives

Not sure! I only recently downloaded it and I wasn't able to find anything in the github regarding epubs.

JorgeRui:
I've tried implementing this on #387, but only works for epubs.

CyanVoxel:
ePub previews implemented in #539 (port of #387)

I'll need to look into other formats such as MOBI and AK3

PR #387, opened 2024-08-24
feat: Add thumbnail extraction for ePub files

JorgeRui:
Implemented _epub_cover method to extract and render the cover image from ePub files. The method scans the contents of the ePub (treated as a ZIP archive) and searches for image files (e.g., .png, .jpg, .jpeg, .gif, .bmp, .svg). This feature enables the generation of thumbnails for ePub files, enhancing the application's ability to display visual representations of eBooks.

Closes partially #360 

Other ebooks file formats are yet not supported, as each one of them have their one unique specifications. This method can be reused for ibooks in theory, but haven't tested it.

image
image


CyanVoxel:
Looks great! Thank you for following the existing patterns for the MediaTypes and render methods, including the alphabetizing! 😁

The only other piece of feedback I have without testing this myself is also echoed by MyPy: I would declare a `None` image at the top of the method like `im: Image.Image = None` and close out the method with `return im`. This fixes the missing return case and mirrors how the other methods work. Also a brief docstring description for `_epub_cover()` would be great!

Leseratte10:
Is it a good idea to just pick the first image that's present in the EPUB file? That's not necessarily the cover, it may also be another random image from the book. 

A better way would be to parse the XML at /META-INF/container.xml (standardized path), follow the path to the OPF file that's mentioned there, and then find any cover files defined in there, like this [python script](https://github.com/marianosimone/epub-thumbnailer/blob/master/src/epub-thumbnailer.py).

tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -764,6 +766,33 @@ def _video_thumb(self, filepath: Path) -> Image.Image:
             )
         return im
 
+    def _epub_cover(self, filepath: Path) -> Image.Image:
```
yedpodtrzitko, line 769: it would be a nice to have test for similar this:

```python
from syrupy.extensions.image import PNGImageSnapshotExtension


def test_svg_to_png(snapshot):
    file_path = Path(__file__).parent / "fixtures" / "sample.epub"

    png_image = ThumbRenderer(...)._epub_cover(file_path, 600)
    assert png_image.read() == snapshot(extension_class=PNGImageSnapshotExtension)
```

CyanVoxel:
I'm moved this PR from the deprecated `thumbnails` branch to `Alpha-v9.4`, which is the branch that thumbnails was eventually merged into. However the v9.4.x releases are now feature frozen and are now solely reserved for bugfixes. This means that after the changes from `thumbnails` are ported to main that I'll once again be moving the target for this PR over to there as a v9.5 feature.

CyanVoxel:
Closed in favor of #539 which ports the commits here over to a clean branch.

PR #391, opened 2024-08-25
Logging, and seperation for cli args.

HEROgold:
This pull request adds a default setting for the logging module, and it setups up a file and stream handler in a nice to read format.

image

There's also a function added that helps to add loggers to different locations.

This request also removes code that's not required anymore due to the swap to logging.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,6 +12,15 @@
 			"console": "integratedTerminal",
 			"justMyCode": true,
 			"args": []
+		},
+		{
+			"name": "TagStudio: Debug",
+			"type": "debugpy",
+			"request": "launch",
+			"program": "${workspaceRoot}/tagstudio/tag_studio.py",
+			"console": "integratedTerminal",
+			"justMyCode": true,
+			"args": ["--debug"]
 		}
 	]
 }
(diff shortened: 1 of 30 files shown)


HEROgold:
There's also a small seperation for the argparse setup, from the main file, this makes it easier to use args in logger.py and main.py as they both using args. (args.debug for logger.py)

CyanVoxel:
Unfortunately this is going to massively conflict with #332, which has been a long time coming and has priority over other PRs at this time. Large scale refactors like this should be communicated in advance to avoid these kinds of conflicts.

There's also been another attempt at a logging refactor with #235 which ran into its own issues and has remained as a draft. I appreciate the work put into this, but it would be better to wait until after #332 is merged into `main` to propose a refactor like this.

HEROgold:
Ahh i understand, i'll wait for the sql update before deciding anything then

CyanVoxel:
Just chiming in to mention that #332 has been merged. I'll go ahead and close this PR to keep things tidy and provide a clean slate for a future logging PR. If you're still interested in taking this up, I would recommend opening up an Issue for it first or at least opening a PR as a draft to provide a place for discussion before commiting any changes. #235 has shown that there's a lot of opinions to be had about how logging should work, and that a PR shouldn't try to take on unrelated tasks that could drag it down.

PR #398, opened 2024-08-25
fix: correct behavior for tag widget search options

### LINKED ISSUE #211: Subtags not working as expected?
Having two tags, `A` (`tag_id:1000`) and `B` (`tag_id:1001`), with `A` having the subtag `B`¹, imagine you have files `a` and `b` with tags `A` and `B` selected respectively.
If you look for `tag_id:1000` or `tag_id:1001`, you only get one single file for each of them, despite the fact one of them should imply the other one. It makes the search function unusable

¹ Reading the documentation, I have a feeling subtag is a *terrible* name for that, as it's kinda the wrong way. “Sub-” implies part of, while it in fact specifies the tags the tag inherits its values from. Maybe it'd be better to name them “supertags” (super-, meaning above or over) or “parent tags”?

CyanVoxel:
This PR addresses three related issues:
1. #211: "Search for Tag" context menu option + left-click behavior for tag widgets did not include tag clusters in results; only instances of that exact tag ID. Expected behavior is to include the entire inheritance cluster, and this has now been implemented. Additional commented-out code has been removed inside the `search_library()` method. 
2. "Add to Search" context menu option was non-functional. This option has been removed from the context menu.
3. Undocumented issue found during testing: The "Search for Tag" context menu and left-click behavior was non-functional for tag widgets inside the "Tag Manager" panel. This functionality has been implemented.

I have not updated the display of tag IDs in the search bar as discussed in 211, as I feel this is much better suited for a future UI revamp of how tags are displayed in the search bar.

Closes #211.

### DIFF
--- a/tagstudio/src/core/library.py
+++ b/tagstudio/src/core/library.py
@@ -1352,10 +1352,13 @@ class Library:
             only_missing: bool = "missing" in query or "no file" in query
             allow_adv: bool = "filename:" in query_words
             tag_only: bool = "tag_id:" in query_words
+            tag_only_ids: list[int] = []
             if allow_adv:
                 query_words.remove("filename:")
             if tag_only:
                 query_words.remove("tag_id:")
+                tag_only_ids.append(int(query_words[0]))
+                tag_only_ids = tag_only_ids + self.get_tag_cluster(int(query_words[0]))
             # TODO: Expand this to allow for dynamic fields to work.
             only_no_author: bool = "no author" in query or "no artist" in query
 
@@ -1382,15 +1385,9 @@ class Library:
                             all_tag_terms.remove(all_tag_terms[i])
                             break
 
-            # print(all_tag_terms)
-
-            # non_entry_count = 0
             # Iterate over all Entries =============================================================
             for entry in self.entries:
                 allowed_ext: bool = entry.filename.suffix.lower() not in self.ext_list
-                # try:
-                # entry: Entry = self.entries[self.file_to_library_index_map[self._source_filenames[i]]]
-                # print(f'{entry}')
 
                 if allowed_ext == self.is_exclude_list:
                     # If the entry has tags of any kind, append them to this main tag list.
@@ -1432,7 +1429,6 @@ class Library:
                     # elif query in entry.path.lower():
 
                     # NOTE: This searches path and filenames.
-
                     if allow_adv:
                         if [q for q in query_words if (q in str(entry.path).lower())]:
                             results.append((ItemType.ENTRY, entry.id))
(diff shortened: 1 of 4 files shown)


tagstudio/src/core/library.py
```diff
@@ -1352,10 +1352,13 @@ def search_library(
             only_missing: bool = "missing" in query or "no file" in query
             allow_adv: bool = "filename:" in query_words
             tag_only: bool = "tag_id:" in query_words
+            tag_only_ids: list[int] = []
             if allow_adv:
                 query_words.remove("filename:")
             if tag_only:
                 query_words.remove("tag_id:")
+                tag_only_ids.append(int(query_words[0]))
+                tag_only_ids = tag_only_ids + self.get_tag_cluster(int(query_words[0]))
```
eivl, line 1361: Will query_words[0] always be a digit? 
Use extend on tag_only_ids instead of adding two lists together.

tagstudio/src/qt/modals/tag_database.py
```diff
@@ -18,17 +19,21 @@
...
+if typing.TYPE_CHECKING:
+    from src.core.library import Tag
+    from src.qt.ts_qt import QtDriver
+
 
 class TagDatabasePanel(PanelWidget):
     tag_chosen = Signal(int)
 
-    def __init__(self, library):
+    def __init__(self, library: "Library", driver: "QtDriver"):
```
eivl, line 31: Literal string `"Library"` and `"QtDriver"` as the type looks like a typo.

tagstudio/src/qt/widgets/tag.py
```diff
@@ -74,8 +74,8 @@ def __init__(
         # search_for_tag_action.triggered.connect(on_click_callback)
         search_for_tag_action.triggered.connect(self.on_click.emit)
         self.bg_button.addAction(search_for_tag_action)
-        add_to_search_action = QAction("Add to Search", self)
-        self.bg_button.addAction(add_to_search_action)
+        # add_to_search_action = QAction("Add to Search", self)
+        # self.bg_button.addAction(add_to_search_action)
```
eivl, line 78: Can these two lines be deleted instead?

eivl (changes requested):
Approved when comments are resolved.

tagstudio/src/core/library.py
CyanVoxel: Was about to confidently say "Yes, the `tag_id:` prefix is only triggered automatically by the program" and then realized that if the user typed in something like "tag_id: gotcha" it would fail...

tagstudio/src/qt/modals/tag_database.py
CyanVoxel: The string literals are there to prevent NameErrors, since QtDriver is not actually imported at runtime (nor should Library be, just caught that) and is only under `if typing.TYPE_CHECKING:` for type checking. It's quirky, but it's what [PEP 484](https://peps.python.org/pep-0484/#forward-references) suggests

PR #403, opened 2024-08-26
Sidecar macro fixes

Computerdores:
This Pull Request addresses 3 Bugs I found while using the gallery_dl-sidecar importing feature:
1. Running macros on files that aren't in a subdirectory was broken (see 1d4570174a3014cfaf348ae5752537528d9ed11e).
2. For `filename.jpg` `get_gdl_sidecar` would try to load `filename.json` instead of `filename.jpg.json`, which is the default filename format used by gallery_dl (see a0562a876630e7bd7fd0b3b91180c8c7bb32975c).
3. `add_generic_data_to_entry` would create a second Content Tags Field if the existing Content Tags field was the first field, because an if-statement used `>` instead of `>=` (see fbd2244a2eb27f9454c8d33006e5f443105fdcea).

I didn't create an Issue first because I had already fixed these issues in order to use the autofill macro to import gdl sidecar files.

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -947,7 +947,7 @@ class QtDriver(QObject):
         """Runs a specific Macro on an Entry given a Macro name."""
         entry = self.lib.get_entry(entry_id)
         path = self.lib.library_dir / entry.path / entry.filename
-        source = entry.path.parts[0]
+        source = "" if entry.path == Path(".") else entry.path.parts[0]
         if name == "sidecar":
             self.lib.add_generic_data_to_entry(
                 self.core.get_gdl_sidecar(path, source), entry_id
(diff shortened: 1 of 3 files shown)


tagstudio/src/qt/ts_qt.py
```diff
@@ -947,7 +947,7 @@ def run_macro(self, name: str, entry_id: int):
         """Runs a specific Macro on an Entry given a Macro name."""
         entry = self.lib.get_entry(entry_id)
         path = self.lib.library_dir / entry.path / entry.filename
-        source = entry.path.parts[0]
+        source = "" if entry.path == Path(".") else entry.path.parts[0]
```
CyanVoxel, line 950: This would ensure that the source folder remains case-insensitive. Preferably, this could be done within the `get_gdl_sidecar()` method instead to sanitize any input that comes in (ex. creating `source_: str = source.lower()` in there).
```suggestion
        source = "" if entry.path == Path(".") else entry.path.parts[0].lower()
```

CyanVoxel (changes requested):
Thanks for fixing this up! I know this is an unfinished + undocumented feature, but it seems to be working back up to its old standard now. After the `.lower()` comment is resolved this should be good to go!

CyanVoxel (approved):

Issue #14, opened 2024-04-23
Ignore folders and file patterns


ShadowJonathan:
Sorta like a `.gitignore`, to be able to ignore folder patterns like (ironically) `.git`, cache folders, and .DS_Store and the likes.

Else all the hundreds or so hidden files will clog up the interface, with no way to hide them.

shadowking123:
i have implemented this feature by entering the "!" in the start of the query and then the file name you want to filter. 
if the file name is in the query, remove it from search.
should it have more than that?

ShadowJonathan:
yes, it should be more-or-less permanent, as a "ignore all these folders all the time" kind of setting, a permanent file/folder blacklist

shadowking123:
How no i get it.
I will work on it

Loran425:
Hiding files with specific extensions was added in https://github.com/TagStudioDev/TagStudio/commit/31f402289565b55d3a84d11f0eb6055e9bf42374

Ignoring specific folders within a library is currently still open.

ItsThatOneJack-Dev:
This would be an amazing feature, and would be useful for many things.

For example, I use a file syncing program called Syncthing to sync my projects and memes folder across devices, and it creates a hidden folder in every synced directory that contains text files with IDs corresponding to files and directories that it uses to ensure files are up-to-date.

It would be great if I could set TagStudio to ignore all hidden directories and files or just add this folder to an ignore list, then TagStudio wouldn't be cluttered by Syncthing files.

yedpodtrzitko:
Currently the list of excluded/allowed file extensions is managed via the GUI. That should remain as is, as not every user is willing/capable to manage that via `.gitignore`-like file.

Having combination of "`.gitignore`" and the GUI interface could have conflicting rules, so I wouldnt go that way of supporting both. 

However there could be an option to have some kind of `.noindex` file (that name is actually taken, so probably something like `.tsnoindex` instead), which would serve as a hard-stop for TS to totally ignore content of given folder. So in the case of syncthing folder, if there would be a file `syncthing/.tsnoindex`, then it would skip that folder.

AwesomeParley:
> However there could be an option to have some kind of `.noindex` file (that name is actually taken, so probably something like `.tsnoindex` instead), which would serve as a hard-stop for TS to totally ignore content of given folder. So in the case of syncthing folder, if there would be a file `syncthing/.tsnoindex`, then it would skip that folder.

[...]

PR #406, opened 2024-08-27
Fix: Correctly ignore files that start with a `.` (eg. `.DS_Store`)

### LINKED ISSUE #399: [Bug]: Ignoring `.DS_Store` files doesn't seem to work
### Checklist

### TagStudio Version

Alpha 9.3.2

macOS 15.0 Beta (24A5327a)

### Description

Using this on macOS, I tried ignoring `.DS_Store` files via the exclusion list for extensions. This however didn't seem to do anything (unless I did something wrong). I'd love to also try and help out debugging, which I may try later today if I've gotten some time 😅.

(If this is duplicated or wrongly filed, super sorry! Couldn't exactly tell if this was a feature req or a bug)

### Expected Behavior

Adding in `.DS_Store` to the ignored extensions should remove them from the UI

### Steps to Reproduce

1. Be on macOS (or just create a file called `.DS_Store` in your folder)
2. Add `.DS_Store` to the exclusion list (I see it gets lowercased but the code also seems to check lowercased anyways)
3. Refresh library and observe

### Logs

_No response_

jlwbr:
This pr adresses 2 issues:

1. #399: `.DS_Store` files where not ignored, this was caused by the pathlib library treating a file starting with a dot the same as a file without an extension. We now treat the whole filename as the extension in this case.
2. When a file is already indexed in the library, but is excluded later, it will stay in the library (only new files are excluded). We now remove these files from the library when refreshing.

Closes #399

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -1183,7 +1183,7 @@ class QtDriver(DriverMixin, QObject):
         all_libs_list = sorted(all_libs.items(), key=lambda item: item[0], reverse=True)
 
         # remove previously saved items
-        self.settings.clear()
+        self.settings.remove("")
 
         for item_key, item_value in all_libs_list[:item_limit]:
             self.settings.setValue(item_key, item_value)


tagstudio/src/core/library.py
```diff
@@ -894,19 +894,30 @@ def refresh_dir(self) -> Generator:
...
                             # print(file)
                             self.files_not_in_library.append(file)
+                    else:
+                        entry_id = self.filename_to_entry_id_map.get(
+                            f.relative_to(self.library_dir), None
+                        )
+
+                        if entry_id:
+                            # A exclude file is in the library, let's remove it.
+                            self.remove_entry(entry_id)
```
CyanVoxel, line 920: Not touching file entries that fall under the exclusion list is intended behavior. This would not only wipe entry data when playing with the exclusion list, but the user wouldn't even be notified of it.

CyanVoxel (changes requested):
Thank you for this! I feel however this would be better as a toggle, perhaps to the left of the existing search options. I would also expect this to handle Windows' hidden file implementation, however I understand if that's out of the scope of what you're able to do.

Example:
image

eivl:
Is this a to broad fix? seems overkill to exclude all files that are hidden under linux/macos/unix just to remove a macos `.DS_Store` file. wouldn't it make more sense to just ignore that file, and continue to extend the list of ignored files as they appear?

jlwbr:
> Is this a to broad fix? seems overkill to exclude all files that are hidden under linux/macos/unix just to remove a macos `.DS_Store` file. wouldn't it make more sense to just ignore that file, and continue to extend the list of ignored files as they appear?

That's not how this fix works, it just treats the name of the dot file the same as an extension of a normal file. (Eg excluding `.json` will exclude `a.json` and `.json` so a file with only an extension is treated the same as a file with an name and an extension.

CyanVoxel:
> > Is this a to broad fix? seems overkill to exclude all files that are hidden under linux/macos/unix just to remove a macos `.DS_Store` file. wouldn't it make more sense to just ignore that file, and continue to extend the list of ignored files as they appear?
> 
> That's not how this fix works, it just treats the name of the dot file the same as an extension of a normal file. (Eg excluding `.json` will exclude `a.json` and `.json` so a file with only an extension is treated the same as a file with an name and an extension.

I also presumed that this was hiding all files with a leading dot as Unix would, my mistake for not reading more closely.

I feel treating files with leading dots as toggleable hidden files instead of as extensions would make more sense, since that's how these files are designed to be treated. Plus if you had a file such as ".file.json", this wouldn't be registered as a proper ".json" file despite it being a valid file extension on a Unix hidden file.

jlwbr:
> Thank you for this! I feel however this would be better as a toggle, perhaps to the left of the existing search options. I would also expect this to handle Windows' hidden file implementation, however I understand if that's out of the scope of what you're able to do.
> 
> 
> 
> Example:
> 
> image
> 
> 

I will try this next week, I had the idea of changing up the system so that uses the same syntax as `.gitignore` this way, you can exclude whole files, all files with a particular extension, etc. But thought that that might be to big of a change for now and might also be to difficult to understand for some people without some solid ui around it.

tagstudio/src/core/library.py
jlwbr: I will remove this then!

CyanVoxel:
> I will try this next week, I had the idea of changing up the system so that uses the same syntax as `.gitignore` this way, you can exclude whole files, all files with a particular extension, etc. But thought that that might be to big of a change for now and might also be to difficult to understand for some people without some solid ui around it.

There's actually an issue associated with this (#14) if you'd be interested in taking a stab at it

CyanVoxel:
@jlwbr Just wanted to touch base on this. A lot has changed in the codebase since this was last worked on, with the biggest being the merging of #332 which introduces a new library backend rendering this PR incompatible in its current state.

Would you still be interested in addressing #399?

PR #407, opened 2024-08-27
Add Tag Modal: Keyboard nav improvements

### LINKED ISSUE #405: [Feature Request]: Better Keyboard Navigation For Add Tag Modal
### Checklist

### Description

The keyboard navigation for the Add Tag Modal is currently a bit wonky. The main problem for me is using Tab to navigate the Modal, because it gets stuck on the Aliases Field.
Also when opening the Add Subtag Modal after closing it previously, the old query is still in the search bar and the search bar isn't in focus.

### Solution

I would like to make the following changes:
- Pressing Tab while in the Aliases Text Box should move focus to the next Element (Subtag Panel / Add Subtag button)
- When opening the Add Tag / Add Subtag Modals the Content of the first text box should be selected. That way it's possible to either start typing normally for a new query or use the left arrow to continue typing at the end of the previous query without using the mouse.

### Alternatives

- Instead of any Tab press in the Aliases Text Box moving the focus past, only a Tab press before typing could be used instead. However this would mean that keyboard navigation is "broken" after typing. And I also don't think Tabs are commonly part of Tag Name Aliases (and they can't be easily used in normal Tag names either, because of the keyboard navigation).
- Instead of selecting the entire Text in the Tag Title / Tag query fields they could instead just receive focus when opening the modals (The tag title field already does), however I think this would make for a worse experience in my opinion, because I have never wanted to add a Tag called "New Tag" and I also most of the time don't want to use the same query again.

Computerdores:
Closes #405.


eivl (approved):
Fantastic change, this introduces the add tag modal with the expected behaviour.

Computerdores:
I made no functional changes, only rebased onto main

CyanVoxel:
I would love to pull this, but unfortunately there was some non-final UI changes to aliases that were attached to #596. If the alias UI gets reverted before #534 is finalized then I'd really like to pull this on top of the old UI.

Computerdores:
As far as I am concerned I could also adapt this to the new UI once that is on main, up to you

CyanVoxel:
> As far as I am concerned I could also adapt this to the new UI once that is on main, up to you

If #534 develops and gets merged then that would be fantastic! If not or if the old UI is restored in the meantime, then I'll likely pull this as-is

Computerdores:
I am reworking the keyboard navigation rn and would like some input regarding how the Tab Focus should move through the window. Main question is how to handle the lists of Aliases and Parent Tags.

The first idea has the advantage of being "what you see is what you get", while the second one has the advantage of always requiring the same number of tabs to go to Alias Add, Subtag Add, Color, Cancel and Save:

I personally lean towards the second idea as it would make the save and cancel buttons more easily reachable, but the first idea with shortcuts for color, save, cancel and the add buttons (save and cancel at the very least) would also be an option imo.

### Idea 1
image

### Idea 2
image

What's your take on this?

CyanVoxel:
> I am reworking the keyboard navigation rn and would like some input regarding how the Tab Focus should move through the window. Main question is how to handle the lists of Aliases and Parent Tags.
> 
> The first idea has the advantage of being "what you see is what you get", while the second one has the advantage of always requiring the same number of tabs to go to Alias Add, Subtag Add, Color, Cancel and Save:
>
> I personally lean towards the second idea as it would make the save and cancel buttons more easily reachable, but the first idea with shortcuts for color, save, cancel and the add buttons (save and cancel at the very least) would also be an option imo.

I think I prefer the second idea as well, and it would pair nicely with the arrow keys being used to navigate back up into the pre-existing alias list so it stays accessible.

I also know there's been some confusion around this, but the alias UI currently in the program is *not* permanent and was never supposed to be added - the "final" alias UI is due to be added in #534, however that shouldn't change the general idea of the navigation here, just the implementation specifics.

CyanVoxel:
Removing the "blocked" status now that #534 has been merged

CyanVoxel (approved):

CyanVoxel:

PR #409, opened 2024-08-28
feat: expanded file deletion/trashing

### LINKED ISSUE #269: [Feature Request] Delete files from within the program

CyanVoxel:
## This is an extension to the work done in #282. This includes:
- Files are now sent to the \*system trash instead of being hard-deleted
- Added ability to delete multiple files at once via multi-selection
- Added confirmation dialogue boxes when deleting files
- Added status bar updates for files being deleted
- Added Edit Menu option for deleting files
- Added Delete key shortcut for deleting files
- Added context menu option to delete file via the Preview Panel
- Clarify context menu action name
- Gracefully fail on files that cannot be deleted, with status bar updates
- Fixed not being able to delete currently-playing videos and GIFs

\* On Windows, I can't find a way to tell when a file will be hard-deleted if the OS can't send it to the Recycle Bin. After spending much time on this, I've decided to settle on providing an extra intimidating confirmation dialogue message in place of being able to handle this case.

Closes #269.

**NOTE:** This PR relies on #390 being pulled *first*.

Windows screenshots:
image
image


requirements.txt
```diff
@@ -10,3 +10,8 @@ numpy==1.26.4
 rawpy==0.21.0
 pillow-heif==0.16.0
 chardet==5.2.0
+pydub==0.25.1
+mutagen==1.47.0
+numpy==1.26.4
+ffmpeg-python==0.2.0
```
eivl, line 16: I have seen this in another PR, but ffmpeg needs to be installed. That should be mentioned or written how you would do that.

eivl:
There are many similar or equal commits as in #390 
I don't want to review the same things twice.

tagstudio/src/qt/resource_manager.py
```diff
@@ -31,6 +31,20 @@ def __init__(self) -> None:
                 )
             ResourceManager._initialized = True
 
+    @staticmethod
```
eivl, line 34: why not a `classmethod` This code breaks if you subclass it.

tagstudio/src/qt/ts_qt.py
```diff
@@ -1449,12 +1521,20 @@ def update_thumbs(self):
...
                     entry = self.lib.get_entry(
                         self.nav_frames[self.cur_frame_idx].contents[i][1]
                     )
-                    filepath = self.lib.library_dir / entry.path / entry.filename
+                    filepath: Path = self.lib.library_dir / entry.path / entry.filename
+
+                    try:
+                        item_thumb.delete_action.triggered.disconnect()
+                    except RuntimeWarning:
+                        pass
```
eivl, line 1586: Why not log this?

tagstudio/src/qt/ts_qt.py
```diff
@@ -896,6 +894,31 @@ def delete_files_callback(self, origin_path: str | Path):
             self.main_window.statusbar.showMessage(f"Deleted {deleted_count} files!")
         self.main_window.statusbar.repaint()
 
+    def delete_file_confirmation(self, count: int, filename: Path | None = None) -> int:
```
eivl, line 907: is this a `staticmethod`?

eivl:
I have now reviewd the last commits as I where told those are the new ones for this PR.

eivl (approved):
Approved when #390 and my comments are resolved.

requirements.txt
CyanVoxel: I'm addressing this in #390 👍

tagstudio/src/qt/ts_qt.py
CyanVoxel: The `.disconnect()` method from Qt is currently a pain point for us, as they've recently changed it from throwing a `RuntimeException` to a `RunetimeWarning` - I'm also not sure if it catches here properly due to that. It's also something that for our purposes, we don't really care if it fails - we just want to ensure that it's disconnected without caring about the previous state, initialized or otherwise. Overall we need a better solution for this, but the recent changes in Qt have made this more difficult to come up with a proper solution for.

tagstudio/src/qt/ts_qt.py
CyanVoxel: It's not, but should it be?

tagstudio/src/qt/resource_manager.py
CyanVoxel: Similar explanation here: https://github.com/TagStudioDev/TagStudio/pull/390#discussion_r1739560040

CyanVoxel:
(Sorry for the commit flood, this PR just got caught up in a a big rebasing mess upstream)

PR #419, opened 2024-08-30
Add Docker support

strenkml:
Adds a docker container that uses [docker-baseimage-kasmvnc](https://github.com/linuxserver/docker-baseimage-kasmvnc) as its base image.  TagStudio's interface will will be available in the browser at port 3000.

Also adds a workflow that will trigger after the Release workflow finishes. The workflow builds and publishes the Docker image to the Github Docker Repo.


seakrueger:
I do not feel this is the right approach to a containerized TagStudio. This goes against the design philosophies of a container, bundling the entire monolithic application and effectively treating it as a mini-VM, which it is in. I think this is best reflected in the massive image size (3.1gb), and the fact that you are accessing the application through a  VNC. 

Past the philosophical problems, there are also implementation problems. The current TagStudio interface is designed around working with multiple different libraries in different directories, which would require either mounting many different volumes (and restarting the container every time more are added) and mounting a broad directory (such as `/`), neither of which feel very intuitive. Most of the UI modals and some other aspects (logo watermark, toolbar, and buttons) are messed up in some way. All the modals are full screen, logo off center, toolbar doesn't align with where the mouse triggers it, etc. It also doesn't seem to handle ctrl+c stop when the container is running in the foreground, the library is not saved (I think this is because TagStudio is not the Entrypoint, and is not capturing SigTerm?). Hotkeys and opening the previously launched library also don't work. I'd also have concerns for security exposing this anywhere, allowing unimpeded access to the VFS and arbitrary file execution as root. 

There have been several discussions in the Discord about better handling the use of TagStudio on servers as well as a proper rewrite of the CLI. I feel it would be better to reproach a containerized variant then, if they ever materialize.

CyanVoxel:
Closing this as unplanned for the reasons @seakrueger described

PR #421, opened 2024-08-30
Added Documantation

ioorik:


tagstudio/src/qt/flowlayout.py
```diff
@@ -23,101 +23,149 @@
...
             item = self.takeAt(0)
 
-    def addItem(self, item):
+    def addItem(self, item) -> None:
+        """Add a given item to the list of items inside the layout."""
         self._item_list.append(item)
 
-    def count(self):
+    def count(self) -> int:
+        """Returns the amount of items in the layout"""
```
yedpodtrzitko, line 62: [PEP-257](https://peps.python.org/pep-0257/#one-line-docstrings)  (Docstring Conventions) says:

> The docstring is a phrase ending in a period. It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”.

tagstudio/src/qt/flowlayout.py
```diff
@@ -23,101 +23,149 @@
...
         if parent is not None:
+            # Set margins to 0.
             self.setContentsMargins(QMargins(0, 0, 0, 0))
 
-        self._item_list = []
-        self.grid_efficiency = False
+        # Initialize the items list to nothing.
+        self._item_list: list = []
+        # Initialize grid efficiency to "False"
+        self.grid_efficiency: bool = False
```
yedpodtrzitko, line 50: This might be just me, but I dont see any value in these comments. The code here is pretty much self-explenatory and doesnt need any comment. Good comments should tell "why", not "what".

tagstudio/src/qt/flowlayout.py
```diff
@@ -23,101 +23,149 @@
...
+            return False
 
-        return None
-
-    def takeAt(self, index):
-        if 0 <= index < len(self._item_list):
+    def takeAt(self, index: int):
+        """
+        Acts like the pop instructions of a list.
+        (Because that's exactly what it does but don't tell anyone.)
```
yedpodtrzitko, line 76: 👎

tagstudio/src/qt/flowlayout.py
ioorik: Sorry, could you maybe um explain it a little further? I am having trouble understanding you. Thank you

tagstudio/src/qt/pagination.py
```diff
@@ -138,6 +138,7 @@ def update_buttons(self, page_count: int, index: int, emit: bool = True):
             raise ValueError("Negative index detected")
 
         # Screw it
+        # Gotta give  it to ya: Nice docs on this one lol
```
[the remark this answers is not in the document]
ioorik: Okay but  come on, it's funny

tagstudio/src/qt/flowlayout.py
ioorik: Absolutly right. I  will get on that

tagstudio/src/qt/flowlayout.py
ioorik: Come on, it is funny

tagstudio/src/qt/pagination.py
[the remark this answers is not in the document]
ioorik: Wait no, it's supposed to be encourging, not like, idk. I looked at most of that file, it is good.

tagstudio/src/qt/flowlayout.py
yedpodtrzitko: yeah but this is not a place for standup comedy

tagstudio/src/qt/flowlayout.py
ioorik: Huh?

xarvex:
This PR does not provide actual value to the project, each comment is either noise or restating self-documenting code, or a combination.

Additionally, ("come on, it is funny")[https://github.com/TagStudioDev/TagStudio/pull/421#discussion_r1738797852] which you've replied to two comments, is not a valid reasoning. This is a serious project.

You are welcome to provide documentation to the project, but such documentation should actually provide additional context. Put bluntly, this doesn't do that.

---

It does seem like this is your first PR, so I would encourage looking at resources for how to contribute to projects. We aren't as strict as some other projects might be, but there needs to be effort and something that can be worked with.

PR #425, opened 2024-09-01
Postpone instantiating modal windows

yedpodtrzitko:
easy backport from #332 which could be actually nice for the upcoming release

- All the modal windows (Folders to Tags etc.) are always instantiated immediately, which takes time.  This PR postpone them until it's actually needed, which speeds up the application start.

~- also reuse `UniversalDetector` instance for files encoding detection.~

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -469,14 +469,24 @@ class QtDriver(QObject):
         window_menu.addAction(check_action)
 
         # Tools Menu ===========================================================
+        def create_fix_unlinked_entries_modal():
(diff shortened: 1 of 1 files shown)


CyanVoxel (commented):
Thank you for this! The UI initialization has been ripe for optimizations like this.

When testing the UniversalDetector however, I noticed that it appears to not play well with the multithreaded nature of ThumbRenderer. When running successive tests on a folder full of differently encoded text files, I was able to get it to consistently mix up the encodings producing a different spotty mix of unrendered thumbnails each time.

yedpodtrzitko:
I reverted the last commit with `UniversalDetector`

PR #428, opened 2024-09-01
Add "Get Started" page to guide new users

Bamowen:
I wanted to try and give new users an entry point to the basic workflow of TagStudio.


Bamowen:
Might not be needed as it can be merged with the "Usage" part of the readme

PR #429, opened 2024-09-02
feat(PreviewPanel): Truncate Path to a single line

seakrueger:
Truncates the path to a single line by removing parent dirs. The line width is calculated based on font size, and since its not a mono-spaced font the calculation under reports a little, so it might be a little over zealous in truncation

### DIFF
--- a/tagstudio/src/qt/helpers/file_opener.py
+++ b/tagstudio/src/qt/helpers/file_opener.py
@@ -142,22 +142,43 @@ class FileOpenerLabel(QLabel):
     def truncate_filepath(self, filepath):
         """Removes parent directories to fit path to a single line"""
         path = Path(filepath)
+        parts = path.parts
         # Since font is not monospace, max chars comes out a little low since average is larger than most ASCII chars
         # Ex: Manual counted max of 50 vs computed 45 chars
         max_chars = self.width() // self.font_metrics.averageCharWidth()
 
-        if len(str(path)) > max_chars:
-            name_size = len(path.name) + 3
-            prev = ""
+        if len(str(path)) < max_chars:
+            return str(path)
 
-            # Reverse as pathlib element 0 is full path
-            for parent in reversed(path.parents):
-                if len(str(parent)) + name_size > max_chars:
-                    if sys.platform == "win32":
-                        return f"{prev}\\\u2026\\{path.name}"
-                    return f"{prev}/\u2026/{path.name}"
-                prev = parent
-        return str(path)
+        length = len(parts[0]) + len(path.name) + 4
+        trunc = [path.drive, "\u2026", path.name]
+
+        j = len(parts) - 2
+        for i in range(1, len(parts) // 2):
+            # Append front dir
+            trunc.insert(i, parts[i])
+            length += len(parts[i])
+            if length > max_chars:
+                # Remove exceding element
+                trunc.pop(i)
+                break
+
+            # Append back dir
+            trunc.insert(2 + i, parts[j])
+            length += len(parts[j])
+            if length > max_chars:
+                # Remove exceding element
+                trunc.pop(2 + i)
+                break
+            j -= 1
+        # There's a small region where all dirs can be added but its still trying to truncate,
+        # remove the ellipsis in that case
+        if len(trunc) == len(parts) + 1:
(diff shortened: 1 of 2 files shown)


tagstudio/src/qt/helpers/file_opener.py
```diff
@@ -134,6 +139,64 @@ def setFilePath(self, filepath):
         """
         self.filepath = filepath
 
+    def truncate_single_filepath(self, filepath):
```
yedpodtrzitko, line 142: please add some tests for this method, thanks.

tagstudio/src/qt/helpers/file_opener.py
```diff
@@ -134,6 +139,64 @@ def setFilePath(self, filepath):
...
+    def leaveEvent(self, event: QEvent) -> None:
+        if self.filepath:
+            # Cancel full path showing if mouse leaves before timer is up
+            self.timer.stop()
+            self.setText(self.truncate_single_filepath(self.filepath))
+        return super().leaveEvent(event)
+
+    def _show_full_path_callback(self):
+        """Shows the full filepath instead of truncated version"""
+        self.setText(str(self.filepath))
```
yedpodtrzitko, line 198: a cleaner solution would be to use `QToolTip` to show the full path instead of doing this, what do you think?

tagstudio/src/qt/helpers/file_opener.py
```diff
@@ -134,6 +139,64 @@ def setFilePath(self, filepath):
...
+                break
+            j -= 1
+        # There's a small region where all dirs can be added but its still trying to truncate,
+        # remove the ellipsis in that case
+        if len(trunc) == len(parts) + 1:
+            trunc.remove("\u2026")
+
+        if sys.platform == "win32":
+            return "\\".join(trunc)
+        return "/".join(trunc)
```
yedpodtrzitko, line 181:
```suggestion
        return os.sep.join(trunc)
```

tagstudio/src/qt/helpers/file_opener.py
seakrueger: There currently doesn't seem to be tests for the front end, at least until 332 is merged

tagstudio/src/qt/helpers/file_opener.py
seakrueger: I found QToolTip to be inconsistent in how long it took to appear and (in-general) the delay longer than desired. I'd be willing to go that route if you would prefer though

tagstudio/src/qt/helpers/file_opener.py
yedpodtrzitko: this function can be tested even without QT, no? `Path` goes inside -> string goes outside, so no frontend is needed here.

tagstudio/src/qt/helpers/file_opener.py
yedpodtrzitko: Just by comparing the amount of code needed to achieve what can be done by calling `.setToolTip()`. If the arguments are  "inconsistent delay" (cant reproduce), "and delay longer than desired" (subjective), then in my eyes the `QTooltip` is winner. 


also I was suspicious that a UI toolkit wouldnt have something to trim a text in element, so I checked their docs, and there is [elidedText](https://doc.qt.io/qtforpython-6/PySide6/QtGui/QFontMetrics.html#PySide6.QtGui.QFontMetrics.elidedText) which should do the same thing you're doing via `truncate_single_filepath`

seakrueger: elidedText does not truncate with knowledge of paths. It simply removes characters in the center, destroying the path names. It also just doesn't work properly, it does not crop to the correct width and falls to the new line

image
image

yedpodtrzitko: makes sense. but the test would be still useful to have.

seakrueger:
This is now based on #430 for now

CyanVoxel:
> This is now based on #430 for now

I've just made some changes to #430 based on yed's feedback and merged it into the 9.4 branch - I won't touch it from there, and you're free to continue this without any further interruption from me 😅

seakrueger:
Decided this is not the right direction.

PR #430, opened 2024-09-02
feat(ui): show file creation/modified dates + restyle path label

CyanVoxel:
This PR displays the file creation and modified dates in the preview panel, as well as restyles the file path label to make it appear more legible.
I understand that #429 is also aiming to restyle the the file path label, and perhaps these approaches can be consolidated. 

*Dark Theme*
image

*Light Theme*
image

### DIFF
--- a/tagstudio/src/qt/widgets/preview_panel.py
+++ b/tagstudio/src/qt/widgets/preview_panel.py
@@ -589,13 +589,13 @@ class PreviewPanel(QWidget):
                         update_on_ratio_change=True,
                     )
                     file_str: str = ""
-                    sep_color: str = "#777777"  # Gray
+                    separator: str = (
+                        f"<a style='color: #777777'><b>{os.path.sep}</a>"  # Gray
+                    )
                     for i, part in enumerate(filepath.parts):
                         part_ = part.strip(os.path.sep)
-                        if i == 0:
-                            file_str += f"{"\u200b".join(part_)}<a style='color: {sep_color}'><b>{os.path.sep}</a></b>"
-                        elif i != 0 and i != len(filepath.parts) - 1:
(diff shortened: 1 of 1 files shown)


tagstudio/src/qt/widgets/preview_panel.py
```diff
@@ -546,7 +588,17 @@ def update_widgets(self):
...
                     )
-                    self.file_label.setText("\u200b".join(str(filepath)))
+                    file_str: str = ""
+                    sep_color: str = "#777777"  # Gray
+                    for i, part in enumerate(filepath.parts):
+                        part_ = part.strip(os.path.sep)
+                        if i == 0:
+                            file_str += f"{"\u200b".join(part_)}<a style='color: {sep_color}'><b>{os.path.sep}</a></b>"
+                        elif i != 0 and i != len(filepath.parts) - 1:
+                            file_str += f"{"\u200b".join(part_)}<a style='color: {sep_color}'><b>{os.path.sep}</a></b>"
```
yedpodtrzitko, line 598: in front of the loop

```python
separator = f"<a style='color: {sep_color}'><b>{os.path.sep}</b></a>"
```


```suggestion
                            file_str += f"{"\u200b".join(part_)}{separator}"
                        elif i != 0 and i != len(filepath.parts) - 1:
                            file_str += f"{"\u200b".join(part_)}{separator}"
```

yedpodtrzitko: actually no, both these blocks are identical, so they can be merged together instead of having two different conditions.

CyanVoxel: I was messing with the three-case structure that I didn't even realized when two of the cases had converged... thanks for pointing this out 🙃👍

PR #436, opened 2024-09-03
fix: Do not create command prompt window on subprocess

seakrueger:
Patches files from abandoned libraries are located and updated in src/qt/helpers/vendored with modified sections labeled PATCHED. A wrapper around subprocess.Popen automatically sets the creation flag to no window on windows.

Must be run via a pyinstaller executable for problem to occur.


pyproject.toml
```diff
@@ -1,9 +1,9 @@
...
-exclude = ["main_window.py", "home_ui.py", "resources.py", "resources_rc.py"]
+exclude = ["main_window.py", "home_ui.py", "resources.py", "resources_rc.py", "**/vendored/"]
 
 [tool.mypy]
 strict_optional = false
 disable_error_code = ["union-attr", "annotation-unchecked", "import-untyped"]
 explicit_package_bases = true
 warn_unused_ignores = true
-exclude = ['tests']
+exclude = ['tests', 'src/qt/helpers/vendored']
```
CyanVoxel, line 9: Any reason why the ruff exclusion pattern for vendored is different than mypy's?

CyanVoxel:
This seems to be working when building on Windows with PyInstaller 6.10.0 and testing with video previews and audio waveform previews. I don't see anything immediately wrong with this besides MyPy complaining about the vendored files, which might not be excluded correctly. Also for some reason the last 4 commits keep creeping forward since they say they were authored in the future 🤷‍♂️

CyanVoxel:
Pulling this now with some ignore comments on the affected files. Couldn't seem to get the folder to exclude either, but this will work for now. Thank you so much for getting a fix for this!

PR #441, opened 2024-09-04
feat: Warn user if FFmpeg is not installed

seakrueger:
Creates a Warning dialog on startup if the program cannot find FFmpeg orFFprobe in the PATH. Other interactions with the program are blocked until the issue is either ignored or resolved.

Uses native OS dialog, so should match Mac and Windows design paradigms on their respective systems (testers required to confirm)
image


tagstudio/src/qt/modals/ffmpeg_checker.py
```diff
@@ -0,0 +1,66 @@
...
+        """Checks if both FFmpeg and FFprobe are installed and in the PATH."""
+        # Same checker that ffmpeg-python uses
+        if which("ffmpeg"):
+            self.ffmpeg_missing = False
+            logging.info(f"FFmpeg found!")
+        if which("ffprobe"):
+            self.ffprobe_missing = False
+            logging.info(f"FFprobe found!")
+        # Reverse from missing to installed
+        return not (self.ffmpeg_missing or self.ffprobe_missing)
```
eivl, line 48: Is there a logical or stylistic reason for checking negatives? wouldn't it be better to have it called `self.ffmpeg` and have it to `False` until you find it in `PATH` and set it to True, chance not needing to do any inversion in the return statement?

I know that I am sharing my own style here, and different styles are fine, I have just not seen this to many times so hence my question.

tagstudio/src/qt/modals/ffmpeg_checker.py
```diff
@@ -0,0 +1,66 @@
...
+        self.setWindowModality(Qt.ApplicationModal)
+        self.setWindowTitle("Warning: Missing dependency")
+
+        self.ffmpeg_missing = True
+        self.ffprobe_missing = True
+
+    def installed(self):
+        """Checks if both FFmpeg and FFprobe are installed and in the PATH."""
+        # Same checker that ffmpeg-python uses
+        if which("ffmpeg"):
```
eivl, line 41: I dont like this implementation of which, it could just have been a method in your class, but since it ships with ffmpegs install it is fine to use. But, it uses `os.name` for platform checking, and I have seen many discussion about this usage and how it is better to use `platform.system()` to get the current platform name. 

It also uses os.path instead of `pathlib` 

I would like @CyanVoxel to comment on this. I can write the method if that is wanted as a new PR after this is merged. there is currently nothing wrong about this usage.

eivl (approved):
Approved once comments are resolved. 
Nice change!

tagstudio/src/qt/modals/ffmpeg_checker.py
seakrueger: No, I'm not sure why I flip-floppde the logic there, I had it the other way around previously. 🙃 Fixed -sorry

seakrueger:
It may better to hold off until #460 is merged and update the url to point to the new docs site

Nginearing:
> It may better to hold off until #460 is merged and update the url to point to the new docs site

As soon as #460 is merged, CyanVoxel plans to use the domain "tagstud.io" that is already bought, so the url will actually be something like docs.tagstud.io but it will still be github pages under the hood.

CyanVoxel:
@seakrueger With #460 having been merged, the docs can be updated here 👍

seakrueger:
The url now points to "https://docs.tagstud.io/help/ffmpeg/". Though it seems the new doc page will need to be cherry-picked to main for the url to work

CyanVoxel:
The `doc` directory was also renamed to `docs` on main, just so you don't miss it

seakrueger:
The actual docs page has now been spun off into its own PR (#473). The actual guide does not need to be part of this PR, and that solves the renamed dir and need to be on a different branch problems.

tagstudio/src/qt/modals/ffmpeg_checker.py
CyanVoxel: Could `shutil.which()` be used here instead?

seakrueger: I did not know the Python stdlib had a built in which command, using that instead.

NeedNot:
Maybe just me but on Macos it says it can't find ffmpeg but the ffmpeg command works fine along with which or where ffmpeg

PR #442, opened 2024-09-04
SVG thumbnail renderer

Tyrannicodin:
Adds an SVG renderer for thumbnails using Qt's QSvgRenderer that writes to a QImage that is written onto a QBuffer that PIL can then read as an image. However, it does have an issue that I have been unable to resolve: Sometimes thumbnails seem to write over each other or otherwise be corrupted (no affect on the actual files though). I think it's due to the QPainter painting in a thread, but I'm not sure.
You can see an example of this problem below:
image

### DIFF
--- a/tagstudio/src/qt/widgets/thumb_renderer.py
+++ b/tagstudio/src/qt/widgets/thumb_renderer.py
@@ -27,8 +27,9 @@ from PIL import (
 from PIL.Image import DecompressionBombError
 from pillow_heif import register_avif_opener, register_heif_opener
 from pydub import AudioSegment, exceptions
-from PySide6.QtCore import QObject, QSize, Qt, Signal
-from PySide6.QtGui import QGuiApplication, QPixmap
+from PySide6.QtCore import QObject, QSize, Qt, Signal, QBuffer
+from PySide6.QtGui import QGuiApplication, QPixmap, QImage, QPainter
+from PySide6.QtSvg import QSvgRenderer
 from src.core.constants import FONT_SAMPLE_SIZES, FONT_SAMPLE_TEXT
 from src.core.media_types import MediaCategories, MediaType
 from src.core.palette import ColorType, get_ui_color
@@ -759,8 +760,24 @@ class ThumbRenderer(QObject):
             filepath (Path): The path of the file.
             size (tuple[int,int]): The size of the thumbnail.
         """
-        # TODO: Implement.
-        im: Image.Image = None
+        
+        # Create an image to draw the svg to and a painter to do the drawing
+        image: QImage = QImage(size, size, QImage.Format.Format_ARGB32)
+        painter: QPainter = QPainter(image)
+
+        # Create an svg renderer, then render to the painter
+        svg: QSvgRenderer = QSvgRenderer(str(filepath))
+        svg.render(painter)
+        del painter
+
+        # Write the image to a buffer as png
+        buffer: QBuffer = QBuffer()
+        buffer.open(QBuffer.OpenModeFlag.ReadWrite)
+        image.save(buffer, "PNG")
+        
+        # Load the image from the buffer
+        im: Image.Image = Image.open(BytesIO(buffer.data()))
+
         return im
 
     def _model_stl_thumb(self, filepath: Path, size: int) -> Image.Image:


CyanVoxel:
> Sometimes thumbnails seem to write over each other or otherwise be corrupted

Yeah, that sounds like a multithreading issue since the `thumb_renderer` is multithreaded

tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -759,8 +760,26 @@ def _image_vector_thumb(self, filepath: Path, size: int) -> Image.Image:
...
+
+        # Create an image to draw the svg to and a painter to do the drawing
+        image: QImage = QImage(size, size, QImage.Format.Format_ARGB32)
+        image.fill("#00000000")
+        painter: QPainter = QPainter(image)
+
+        # Create an svg renderer, then render to the painter
+        svg: QSvgRenderer = QSvgRenderer(str(filepath))
+        svg.render(painter)
+        del painter
```
eivl, line 772: What is the reasons for the manual cleanup?

tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -759,8 +760,26 @@ def _image_vector_thumb(self, filepath: Path, size: int) -> Image.Image:
...
-        im: Image.Image = None
+
+        # Create an image to draw the svg to and a painter to do the drawing
+        image: QImage = QImage(size, size, QImage.Format.Format_ARGB32)
+        image.fill("#00000000")
+        painter: QPainter = QPainter(image)
+
+        # Create an svg renderer, then render to the painter
+        svg: QSvgRenderer = QSvgRenderer(str(filepath))
+        svg.render(painter)
```
eivl, line 771: how does the thumbnail affect with odd aspect ratios in the SVG?

eivl (approved):
Once comments are addressed I approve.

eivl:
@CyanVoxel I wonder if a future change would make sense, add a context manager so that it handles the buildup and teardown of thumbnails.

tagstudio/src/qt/widgets/thumb_renderer.py
Tyrannicodin: I had done it to fix an error, however that error was also resolved by just ending the painter and I forgot to commit that.

tagstudio/src/qt/widgets/thumb_renderer.py
Tyrannicodin: It would have stretched them, but I have now fixed it to keep aspect ratio.

tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -759,8 +760,27 @@ def _image_vector_thumb(self, filepath: Path, size: int) -> Image.Image:
             filepath (Path): The path of the file.
             size (tuple[int,int]): The size of the thumbnail.
         """
-        # TODO: Implement.
-        im: Image.Image = None
+
+        # Create an image to draw the svg to and a painter to do the drawing
+        image: QImage = QImage(size, size, QImage.Format.Format_ARGB32)
+        image.fill("#00000000")
```
yedpodtrzitko, line 766: this would make it a bit more readable:

```suggestion
        image.fill(Qt.transparent)
```

tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -759,8 +760,27 @@ def _image_vector_thumb(self, filepath: Path, size: int) -> Image.Image:
...
+        svg.render(painter)
+        painter.end()
+
+        # Write the image to a buffer as png
+        buffer: QBuffer = QBuffer()
+        buffer.open(QBuffer.OpenModeFlag.ReadWrite)
+        image.save(buffer, "PNG")
+
+        # Load the image from the buffer
+        im: Image.Image = Image.open(BytesIO(buffer.data().data()))
```
yedpodtrzitko, line 781:
```suggestion
        im: Image.Image = Image.open(BytesIO(buffer.data()))
```

yedpodtrzitko:
nice work, but as always - when adding a new feature, please add a test. 


For this one having some sample `.svg` image alongside the test, and then compare the output via `syrupy` that it really provides the expected output. I made a similar test in some other project, so it could look something like this:

```python
from syrupy.extensions.image import PNGImageSnapshotExtension


def test_svg_to_png(snapshot):
    file_path = Path(__file__).parent / "fixtures" / "sample.svg"

    png_image = ThumbRenderer(...)._image_vector_thumb(file_path, 600)
    assert png_image.read() == snapshot(extension_class=PNGImageSnapshotExtension)
```

yedpodtrzitko:
also what's the git branch `thumbnails`? Wouldnt be better to aim this either to the `Alpha-9.4` or `main`?

yedpodtrzitko (changes requested):
missing test

CyanVoxel:
I'm moved this PR from the deprecated `thumbnails` branch to `Alpha-v9.4`, which is the branch that thumbnails was eventually merged into. However the v9.4.x releases are now feature frozen and are now solely reserved for bugfixes. This means that after the changes from `thumbnails` are ported to main that I'll once again be moving the target for this PR over to there as a v9.5 feature.

CyanVoxel:
I've ported this functionality to `main` via #540, thank you for your contribution with this!

PR #447, opened 2024-09-04
Translations

Tyrannicodin:
Adds a translator that handles translation of objects via QCoreApplication.translate(*context*, *key*), and translations are stored as .ini files in the format:
`*context*.*key*="*Translation*"`

Currently has very few actually translated components, just the structure to add more.


CyanVoxel:
Now that #332 has merged, this would be a great time to start separating out the strings to try using with [Weblate](https://weblate.org/)

Bamowen:
Strings separated in #507 
I guess the best course of action would either be to:
- Merge this PR, then #507 
- Add the json of strings to this PR.

Any thoughts?

Tyrannicodin:
> @Tyrannicodin how much is left for this to be a proper PR and not a draft?

The only missing thing is replacing all strings with translation functions, which I want to do in this pr or else it won't really add much.

Bamowen:
[en.json](https://github.com/user-attachments/files/17030068/en.json)
Here you go!

Tyrannicodin:
Hmm, I'm not sure we need to translate logs right? They're not user-facing after all.

seakrueger:
Just wanted to bump this and check if this is still being worked on. Translations are being pulled from webplate, so the sooner we can implement those, the better. Thanks!

PR #455, opened 2024-09-06
More ways to open files

Tyrannicodin:
Adds both double click functionality to thumbnails in the thumbnail grid and an open files shortcut `ctrl+shift+O`.

### DIFF
--- a/tagstudio/src/qt/widgets/item_thumb.py
+++ b/tagstudio/src/qt/widgets/item_thumb.py
@@ -196,9 +196,7 @@ class ItemThumb(FlowWidget):
         self.thumb_button.setContextMenuPolicy(Qt.ContextMenuPolicy.ActionsContextMenu)
 
         self.opener = FileOpenerHelper("")
-        self.thumb_button.double_clicked.connect(
-            lambda: self.opener.open_file()
-        )
+        self.thumb_button.double_clicked.connect(lambda: self.opener.open_file())
         open_file_action = QAction("Open file", self)
         open_file_action.triggered.connect(self.opener.open_file)
         open_explorer_action = QAction("Open file in explorer", self)
--- a/tagstudio/src/qt/widgets/item_thumb.py
+++ b/tagstudio/src/qt/widgets/item_thumb.py
@@ -196,7 +196,7 @@ class ItemThumb(FlowWidget):
         self.thumb_button.setContextMenuPolicy(Qt.ContextMenuPolicy.ActionsContextMenu)
 
         self.opener = FileOpenerHelper("")
-        self.thumb_button.clicked.connect(
+        self.thumb_button.double_clicked.connect(
             lambda: self.opener.open_file() if self.thumb_button.selected else None
         )
         open_file_action = QAction("Open file", self)
--- a/tagstudio/src/qt/widgets/item_thumb.py
+++ b/tagstudio/src/qt/widgets/item_thumb.py
@@ -197,7 +197,7 @@ class ItemThumb(FlowWidget):
 
         self.opener = FileOpenerHelper("")
         self.thumb_button.double_clicked.connect(
-            lambda: self.opener.open_file() if self.thumb_button.selected else None
+            lambda: self.opener.open_file()
         )
         open_file_action = QAction("Open file", self)
         open_file_action.triggered.connect(self.opener.open_file)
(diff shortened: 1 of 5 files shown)


yedpodtrzitko:
Double click makes sense, but `Ctrl+Shift+O` feels very unusual. Maybe it's because I'm on macOS where the convention is `Cmd`+`↓`  (`ControlModifier` + `Key.Key_Down`). Using this combo here would make it feel right.

CyanVoxel:
TIL that [Command-Down Arrow](https://support.apple.com/en-us/102650) opens the selected item in Finder. If there was an equivalent shortcut for Windows Explorer I would've recommended to use each on their respective platforms, but I can't seem to find one.

tagstudio/src/qt/widgets/item_thumb.py
```diff
@@ -218,6 +218,9 @@ def __init__(
 
         self.thumb_button.setContextMenuPolicy(Qt.ContextMenuPolicy.ActionsContextMenu)
         self.opener = FileOpenerHelper("")
+        self.thumb_button.clicked.connect(
+            lambda: self.opener.open_file() if self.thumb_button.selected else None
+        )
```
CyanVoxel, line 223: What is this supposed to be doing? Removing it doesn't change any functionality

CyanVoxel (changes requested):
In your description you say that you've added double-clicking to open files, however I don't see anything in the code ~~or in practice that would suggest that.~~

EDIT: I'm now able to see what the extra click connection was supposed to do, however I would not call this "double-click" to open. Every additional click opens the file an additional time, so triple-clicking the file opens it twice, quadruple-clicking opens it three times, and so on. This also triggers when you make a multi-file selection then return to click the original file with the intention of resetting the selection back to that file. Finally it also seems to break as soon as you interact with the program a bit by searching, which is why I did not see any of this behavior initially.

This may be due to some other changes since #332, but it could also be related to the fact that the connection here is never disconnected. `item_thumb` has the `update_clickable` method that's used to do exactly this, where it disconnects the previous clickable and gets assigned a new one. I would recommend mirroring this for the double-click connection.

I would also expect Ctrl/Command + Down Arrow to open *all* the selected files, however I understand that this was not even a context menu option previously and would likely require additional refactoring (Plus it seems there's now a bug where Ctrl+Click selecting does not update the preview panel).

yedpodtrzitko:
> TIL that [Command-Down Arrow](https://support.apple.com/en-us/102650) opens the selected item in Finder. If there was an equivalent shortcut for Windows Explorer I would've recommended to use each on their respective platforms, but I can't seem to find one.

on Windows it's just Enter, no?

CyanVoxel:
> > TIL that [Command-Down Arrow](https://support.apple.com/en-us/102650) opens the selected item in Finder. If there was an equivalent shortcut for Windows Explorer I would've recommended to use each on their respective platforms, but I can't seem to find one.
> 
> on Windows it's just Enter, no?

Apparently it is... So simple that Microsoft doesn't even consider it a "[shortcut](https://support.microsoft.com/en-us/windows/keyboard-shortcuts-in-windows-dcc61a57-8ff0-cffe-9796-cb9706c75eec)"

yedpodtrzitko (changes requested):

yedpodtrzitko:
please fix the pipeline, thanks.

yedpodtrzitko (approved):

roboticforest:
> In this case I would recommend using the default behavior for each major OS's native file manager (macOS and Windows in this case, with a fallback on one for Linux unless someone can provide more information on that).

Linux file browsers work just like Windows, or at least every version of Ubuntu, Mint, and old versions of RedHat that I've used. The file browsers let you navigate with arrow keys, or by typing the name of the file, then you can just press enter to open what's selected. Or, double-click, of course.

CyanVoxel:
Thank you for your work on this! I apologize for not getting it pulled previously, I was anticipating the platform specific shortcuts to be changed in the background while tending to other high priority areas of the project.

What I've done now is port and update this functionality to a new PR (#858) with a co-author commit that brings over all of the changes added here along with the last bit of review feedback and updates to bring it up to the current codebase standards (translations, formatting, changed imports, etc.). I also have plans to extend some of the functionality here to other areas of the project, such as adding the double-click to the preview panel.

Thank you again for your previous work on this, and I hope you're doing well!

Issue #461, opened 2024-09-06
[Bug]: all files in all libraries are unlinked (red)


satorkain:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

alpha 9.4.0

### Operating System & Version

win 10

### Description

just downloaded and installed TagStudio. Created a new library in which for some reason the all files are unlinked. 

For example, just one music track in the new folder I just created in the root of the disc:

![изображение](https://github.com/user-attachments/assets/a7921ef3-2e08-4a35-a3cf-6a93f577279e)

And so with all libraries and all files. 

Experimenting, I found that only one folder with fonts opened without broken links.

fonts [Proxima Nova 2.zip](https://github.com/user-attachments/files/16906986/Proxima.Nova.2.zip) – files will be linked
track [ssssssssssss.zip](https://github.com/user-attachments/files/16907000/ssssssssssss.zip) – track will be unlinked

### Expected Behavior

there should be no unlinked files

### Steps to Reproduce

this Issue occurs with any folder... any newly created or added to TagStudio.

### Logs

where to find it?

Nginearing:
> ### Logs
> 
> where to find it?

I assume to do debug mode you just run the exe from command prompt

> just downloaded and installed TagSpace.
TagSpace is it called? 😂😂

satorkain:
although not all files in the library are red, e.g. pdf and txt. But mp3's are all red. 
Llibrary with more than 20 thousand files where mp3's are placed in different folders.

Nginearing:
Copy paste all the stuff you see in the command prompt when you run the exe (from within the command prompt)

Nginearing:
> > I assume to do debug mode you just run the exe from command prompt
> 
[...]

PR #471, opened 2024-09-08
fix(ThumbRenderer): Use audio icon when no ffmpeg

seakrueger:
When FFmpeg is missing, Popen raises a FileNotFound error. This would be caught as an Unlinked file and use the broken file icon. The exception is now caught and a more appropriate exception is raised in its place.

Fixes: #461
image


CyanVoxel:
Awesome! This was one thing that bugged me about #461 that I couldn't test the reason for myself. Thank you for the fix!

PR #478, opened 2024-09-09
fix(Flake): Add missing x11 dependency

seakrueger:
Follow #332, tagstudio would no longer start (on wayland) without adding x11 as a dependency to the library path.


xarvex (approved):
Thanks for catching this! @CyanVoxel can you merge as I am on mobile? (tested earlier but in lecture now)

Issue #386, opened 2024-08-24
[Bug]: GLIBC version mismatch on Rocky/RHEL 9 (requires 2.35)


egg82:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Alpha v9.3.2

### Operating System & Version

Rocky Linux 9.4

### Description

Portable binary fails with the following error:
```
Failed to load Python shared library '/tmp/_MEIe8CWw8/libpython3.12.so.1.0': dlopen: /lib64/libm.so.6: version `GLIBC_2.35' not found (required by /tmp/_MEIe8CWw8/libpython3.12.so.1.0)
```

Non-portable fails with a similar error:
```
Failed to load Python shared library '/home/X/.local/bin/_internal/libpython3.12.so.1.0': dlopen: /lib64/libm.so.6: version `GLIBC_2.35' not found (required by /home/X/.local/bin/_internal/libpython3.12.so.1.0)
```

If this is expected, the installation instructions should be updated to reflect that only specific GLIBC versions are supported.

The latest Rocky release comes with 2.34, but 2.35+ is required:
```
ldd --version 

ldd (GNU libc) 2.34
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
```

### Expected Behavior

Program to launch

### Steps to Reproduce

1. Download + extract latest release on a system with an older GLIBC version
2. Run

Leseratte10:
Same issue happens on Ubuntu 20.04 which is still supported by Canonical and comes with glibc 2.31. 

2.35 is the glibc version that comes with Ubuntu 22.04. Looking at the release.yml which is used to build the project, that seems to be using the ubuntu 22.04 image which would explain the issue.

Ideally, you'd build software for Linux on the oldest supported distribution. Github is still supporting Ubuntu 20.04 for a while, and even once they remove that, you can always use a Docker container. Unless there's a particular reason other than "it's the default" as to why TagStudio uses 22.04. 

[...]

PR #479, opened 2024-09-09
Switch Linux CI to python docker image

Leseratte10:
Currently, the build CI for Linux uses Github's Ubuntu 22.04 VM image. Ubuntu 22.04 comes with Glibc 2.35, so this means that the Linux builds of TagStudio will only run on distributions that also come with Glibc >= 2.35. 

This PR changes that around so Githubs VM image is still being used, but inside that image it uses a Debian Bullseye docker container image with Python 3.12. Bullseye is a fair bit older and comes with Glibc 2.31. 

This change should add support for running TagStudio under Debian Bullseye (11), Ubuntu 20.04 (Focal) and Rocky Linux 9 (see #386), and any other Linux distributions coming with glibc < 2.35 and >= 2.31.  

I'm not sure if this change will have unintended side effects, but at least for me, TagStudio built with this change will actually start on Ubuntu 20.04 where before it just complained about a wrong Glibc version.

EDIT: Is this the correct target branch or should this target alpha-v9.4? Looks like the CI is a bit different between the two branches.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -6,17 +6,22 @@ jobs:
   pytest:
     name: Run tests
     runs-on: ubuntu-latest
-    container:
-      image: python:3.12-bullseye
 
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
 
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+          cache: 'pip'
+
       - name: Install system dependencies
         run: |
-          apt-get update && \
-          apt-get install -y --no-install-recommends \
+          # dont run update, it is slow 
+          # sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
             libxkbcommon-x11-0 \
             x11-utils \
             libyaml-dev \
@@ -29,9 +34,7 @@ jobs:
             libxcb-xinerama0 \
             libopengl0 \
             libxcb-cursor0 \
-            libpulse0 \
-            xvfb \
-            xauth
+            libpulse0
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -6,22 +6,17 @@ jobs:
   pytest:
     name: Run tests
     runs-on: ubuntu-latest
+    container:
+      image: python:3.12-bullseye
 
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
 
-      - name: Setup Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: '3.12'
-          cache: 'pip'
-
       - name: Install system dependencies
         run: |
-          # dont run update, it is slow 
-          # sudo apt-get update
-          sudo apt-get install -y --no-install-recommends \
+          apt-get update && \
+          apt-get install -y --no-install-recommends \
             libxkbcommon-x11-0 \
             x11-utils \
             libyaml-dev \
@@ -34,7 +29,9 @@ jobs:
             libxcb-xinerama0 \
             libopengl0 \
             libxcb-cursor0 \
-            libpulse0
+            libpulse0 \
+            xvfb \
+            xauth
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,12 +18,11 @@ jobs:
             build-flag: --portable
             suffix: _portable
     runs-on: ubuntu-22.04
+    container:
+      image: python:3.12-bullseye
+
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
-        with:
-          python-version: '3.12'
-          cache: 'pip'
       - run: pip install -Ur requirements.txt pyinstaller
       - run: pyinstaller tagstudio.spec -- ${{ matrix.build-flag }}
       - run: tar czfC dist/tagstudio_linux_x86_64${{ matrix.suffix }}.tar.gz dist tagstudio


seakrueger:
Have you test that TagStudio works on Debian 11/Rocky 9/Ubuntu 20.04? 
TagStudio interface is written using PySide 6 (Qt6 for Python). Neither [Debian 11](https://pkgs.org/search/?q=qt6&on=name), [Rocky 9](https://pkgs.org/search/?q=qt6&on=name), or [Ubuntu 20.04](https://packages.ubuntu.com/search?suite=focal&arch=any&searchon=names&keywords=qt6) have packages for qt6. I'm not able to test this myself presently, but one of the top results for "qt6 Ubuntu 20.04" is the OBS project dropping support for Ubuntu 20.04 because of Qt6 [obsproject/obs-studio/discussions/9055]. 
Thanks!

.github/workflows/pytest.yaml
```diff
@@ -6,22 +6,17 @@ jobs:
...
-        with:
-          python-version: '3.12'
-          cache: 'pip'
-
       - name: Install system dependencies
         run: |
-          # dont run update, it is slow 
-          # sudo apt-get update
-          sudo apt-get install -y --no-install-recommends \
+          apt-get update && \
```
yedpodtrzitko, line 18: Has this stopped working without running the update? If not, can we keep it disabled? There is nothing essential in the updates to run `pytest`, so it needlessly increases the action running time.

.github/workflows/pytest.yaml
```diff
@@ -6,22 +6,17 @@ jobs:
...
 
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
 
-      - name: Setup Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: '3.12'
-          cache: 'pip'
```
yedpodtrzitko, line 18: why is this removed? I understand that Python is pre-installed in the image, but this also takes care of caching the installed dependencies.

Leseratte10:
> Have you test that TagStudio works on Debian 11/Rocky 9/Ubuntu 20.04? TagStudio interface is written using PySide 6 (Qt6 for Python). Neither [Debian 11](https://pkgs.org/search/?q=qt6&on=name), [Rocky 9](https://pkgs.org/search/?q=qt6&on=name), or [Ubuntu 20.04](https://packages.ubuntu.com/search?suite=focal&arch=any&searchon=names&keywords=qt6) have packages for qt6. I'm not able to test this myself presently, but one of the top results for "qt6 Ubuntu 20.04" is the OBS project dropping support for Ubuntu 20.04 because of Qt6 [[obsproject/obs-studio/discussions/9055](https://github.com/obsproject/obs-studio/discussions/9055)]. Thanks!

I have not fully tested every functionality but I did make sure that it does start and the GUI shows up on my 20.04 install. If needed I can give it another go in a clean 20.04 VM.

.github/workflows/pytest.yaml
Leseratte10: Yes, this has stopped working. The Python docker images don't have the package metadata cached because it's usually just wasted disk space so if you do need to install something inside that docker container you will need to run apt update first. 

Theoretically you could decide to run pytest on ubuntu-latest like before (which would fix that issue), but then you would test on a different OS than you do the actual compiling / bundling on, not sure if that's a good idea.

.github/workflows/pytest.yaml
Leseratte10: With the build process now running inside a container, I don't think this type of caching still works. I did run into some weird issues with the setup-python action combined with the Docker container so I decided to remove it. Maybe someone with more knowledge about Github CI has an explanation for that.

yedpodtrzitko: Whatever is the issue there, removing the caching here is a net loss, so I'm not a fan of this change. If `setup-python` is not working, then please set up https://github.com/actions/cache as a replacement instead, thanks.

.github/workflows/pytest.yaml
yedpodtrzitko: I understand the rationale behind the change, but slowing down the test suite (which runs quite frequently) feels unreasonable when the actual change you want to achieve is relevant for the artifact build. 

Considering we're building also macOS and Windows versions, which dont run the testsuite at all and we still trust the build will be okay, then running *Python* tests on different version than where the build happens should be okay as well.

Leseratte10: Okay, then I'll revert that step and have the tests run on a newer version.

Leseratte10: I have reverted the changes to pytest.yaml which should fix both of your concerns, now it's just the release.yml that has changed.

.github/workflows/pytest.yaml
Leseratte10: I have reverted the changes to pytest.yaml, now it's just the release.yml that has changed. Since a release isn't happening that often, do we still need working caching for that?

yedpodtrzitko: the release workflow can stay without cache.

yedpodtrzitko (approved):
I dont have linux machine to test this, but it looks legit to me 🤷

xarvex:
Getting around to some of these stale issues. I believe the reason we ever used Ubuntu 22.04 as the runner in the beginning is that the 20.04 runner had some sort of issue in building (it has been some time, so details have slipped my mind). Whatever it was, it wouldn't even get to the point of building, which I guess something about this container works around.

To gauge the scope of things, here is a list of distros that are still supported and shipping a GLIBC below 3.35:
| Distribution         | End-of-Life Date | Upgrade Available |
| ---------------------|------------------|-------------------|
| CentOS Stream 9      | 31 May 2027      | ✅                |
| Debian 11 (Bullseye) | 31 August 2026   | ✅                |
| Rocky Linux 9        | 31 May 2032      | ❌                |
| Ubuntu 20.04         | 31 May 2025      | ✅                |

That essentially leaves Rocky Linux 9 being the sole distro with any significant support window, and no upgrade available, with the reason for its behavior seemingly for the sake of supporting enterprise usage. Unfortunately, we aren't going to shift CI to support a tiny group of users on this system, to give potential issues for everyone else. We do have build information on our documentation page if one does wish to have an executable for their outdated system: https://docs.tagstud.io/install/#building

PR #481, opened 2024-09-09
Fix tag search to not require wildcards

Tyrannicodin:
Previously, searching for tags required % wildcards to be added either side in order to search for a tag without an exact match, however this removes that requirement.

### DIFF
--- a/tagstudio/src/core/library/alchemy/library.py
+++ b/tagstudio/src/core/library/alchemy/library.py
@@ -429,8 +429,8 @@ class Library:
             if search.tag:
                 query = query.where(
                     or_(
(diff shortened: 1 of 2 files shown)


yedpodtrzitko:
please add a test for this scenario, so there wont be any regression in future, thanks.

yedpodtrzitko (changes requested):
please add a test for this scenario, so there wont be any regression in future, thanks.

yedpodtrzitko (approved):

CyanVoxel:
Mergeable once the Ruff error is addressed

PR #483, opened 2024-09-09
fix(QtDriver): Retain filter on directory refresh

### LINKED ISSUE #482: [Bug]: Directory refresh ignores search query
### Checklist

### TagStudio Version

Alpha 9.4.0

NixOS 24.05

### Description

If the displayed images are filtered (i.e. only matching images are shown) and then the directories are refreshed, all images are shown. 

After every refresh, ALL thumbnails are regenerated. Thumbnail generation is already slow for me, so this makes adding images 1-by-1 very unpleasant. I thought I could fix this by just entering the query `untagged` which now only generates thumbnails for new images. But the refresh behavior makes this hack useless.

### Expected Behavior

Refreshing directories does not clear the search query, so I would expect that it is immediately applied after refreshing. Right now I have to press enter in the search query bar again after every refresh.

E.g. if I have the query `untagged` and it shows 10 images, I add an image to a folder, I refresh: Now 11 images are shown. But in reality all my images are shown.

### Steps to Reproduce

1. filter displayed images (e.g. `untagged`)
2. (Optional) add an image to a directory in the library
3. Refresh directories `Ctrl + R`

### Logs

_No response_

seakrueger:
Keeps the filtered query thumbnails in the main window, rather than before where the search would remain in the bar but the filter would be reset to none.

Fixes #482

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -1091,7 +1091,13 @@ class QtDriver(QObject):
             )
         )
         r = CustomRunnable(lambda: iterator.run())
-        r.done.connect(lambda: (pw.hide(), pw.deleteLater(), self.filter_items(self.main_window.searchField.text())))
+        r.done.connect(
+            lambda: (
+                pw.hide(),
+                pw.deleteLater(),
+                self.filter_items(self.main_window.searchField.text()),
+            )
+        )
         QThreadPool.globalInstance().start(r)
 
     def new_file_macros_runnable(self, new_ids):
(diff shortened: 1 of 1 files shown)


CyanVoxel (approved):
Approving this as I've come to agree that this behavior is more coinvent, but I wanted to share a little quirk about doing this that was the reason why I avoided initially avoided it.

When refreshing a library and refiltering with the previous search term like this, you're now creating an additional navigation frame that may just be a duplicate of the previous one, if not one that strangely blends into previous versions of similar search results. Honestly it's not like the navigation frame system is exactly flawless these days anyway, and it's something that needs to be replaced regardless. So, story time aside, you've got a thumbs up from me on this change.

PR #495, opened 2024-09-11
Autocomplete tags in the preview panel

Tyrannicodin:
Adds autocomplete to tags in the preview panel, for all tag fields.
You can now add tags through several methods:
1. Beginning to type a tag's name and
    a. Clicking enter to get the first tag that matches the text
    b. Using the arrow keys to select a tag from the autocomplete popup
    c. Clicking on a tag from the autocomplete popup
2. Clicking the plus button and using the previous tag modal.

Requires #481

### DIFF
--- a/tagstudio/src/qt/modals/tag_search.py
+++ b/tagstudio/src/qt/modals/tag_search.py
@@ -4,6 +4,7 @@
 
 
 import math
+from typing import Optional
 
 import structlog
 from PySide6.QtCore import QSize, Qt, Signal
@@ -16,13 +17,11 @@ from PySide6.QtWidgets import (
     QVBoxLayout,
     QWidget,
 )
-from typing import Optional
-
 from src.core.library import Library
 from src.core.library.alchemy.enums import FilterState
 from src.core.palette import ColorType, get_tag_color
 from src.qt.widgets.panel import PanelWidget
-from src.qt.widgets.tag import TagWidget, Tag
+from src.qt.widgets.tag import Tag, TagWidget
 
 logger = structlog.get_logger(__name__)
 
--- a/tagstudio/src/qt/modals/tag_search.py
+++ b/tagstudio/src/qt/modals/tag_search.py
@@ -16,12 +16,13 @@ from PySide6.QtWidgets import (
     QScrollArea,
     QFrame,
 )
+from typing import Optional
 
 from src.core.library import Library
 from src.core.library.alchemy.enums import FilterState
 from src.core.palette import ColorType, get_tag_color
 from src.qt.widgets.panel import PanelWidget
-from src.qt.widgets.tag import TagWidget
+from src.qt.widgets.tag import TagWidget, Tag
 
 logger = structlog.get_logger(__name__)
 
@@ -33,7 +34,7 @@ class TagSearchPanel(PanelWidget):
         super().__init__()
         self.lib = library
         # self.callback = callback
-        self.first_tag_id = None
+        self.first_tag: Optional[Tag] = None
         self.tag_limit = 100
         # self.selected_tag: int = 0
         self.setMinimumSize(300, 400)
@@ -88,9 +89,9 @@ class TagSearchPanel(PanelWidget):
     # 	self.search_field.setFocus()
 
     def on_return(self, text: str):
-        if text and self.first_tag_id is not None:
-            # callback(self.first_tag_id)
-            self.tag_chosen.emit(self.first_tag_id)
+        if text and self.first_tag is not None:
+            # callback(self.first_tag)
+            self.tag_chosen.emit(self.first_tag.id)
             self.search_field.setText("")
             self.update_tags()
         else:
@@ -103,10 +104,13 @@ class TagSearchPanel(PanelWidget):
 
         found_tags = self.lib.search_tags(
             FilterState(
-                path=name,
+                tag=name,
                 page_size=self.tag_limit,
             )
         )
+
+        if len(found_tags) > 0:
+            self.first_tag = found_tags[0]
 
         for tag in found_tags:
             c = QWidget()
--- a/tagstudio/src/qt/modals/tag_search.py
+++ b/tagstudio/src/qt/modals/tag_search.py
@@ -109,8 +109,7 @@ class TagSearchPanel(PanelWidget):
             )
         )
 
-        if len(found_tags) > 0:
-            self.first_tag = found_tags[0]
+        self.first_tag = found_tags[0] if found_tags else None
 
         for tag in found_tags:
             c = QWidget()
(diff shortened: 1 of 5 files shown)


Tyrannicodin:
Yep that's fine, I made this before I looked at that pr so can add some tests to both of them.

yedpodtrzitko:
@CyanVoxel can you please enable the workflow to run on this PR? Thanks.

tagstudio/src/qt/modals/tag_search.py
```diff
@@ -103,11 +104,14 @@ def update_tags(self, name: str | None = None):
 
         found_tags = self.lib.search_tags(
             FilterState(
-                path=name,
+                tag=name,
                 page_size=self.tag_limit,
             )
         )
 
+        if len(found_tags) > 0:
```
yedpodtrzitko, line 112:
```suggestion
        if found_tags:
```

tagstudio/src/qt/widgets/tag_box.py
```diff
@@ -27,6 +27,19 @@
 logger = structlog.get_logger(__name__)
 
 
+class TagCompleter(QCompleter):
+    def __init__(self, parent, lib):
```
yedpodtrzitko, line 31: add type annotations please

yedpodtrzitko (changes requested):
.

yedpodtrzitko:
here's a bug:

- only `Archived` and `Favorite` tags exist
- type `meow` into the field to add new tag & press Enter

Expected behaviour:

- since no matching tag exists, no tag should be added

Current behaviour:

- tag `Archived` is added

tagstudio/src/qt/widgets/tag_box.py
```diff
@@ -27,6 +27,19 @@
...
+        super().__init__(parent)
+        self.lib = lib
+        self.update([])
+    
+    def update(self, exclude:typing.Iterable[str]):
+        tags = [tag.name for tag in self.lib.tags]
+        for value in exclude:
+            tags.remove(value)
```
yedpodtrzitko, line 39: for performance reasons please use `set` instead of `list`, eg:
```suggestion
    def update(self, exclude: set[str]):
        tags = {tag.name for tag in self.lib.tags}
        tags -= exclude
```

yedpodtrzitko (changes requested):
Please add some tests for the widget, thanks.

Tyrannicodin:
Added tests and fixed the bugs 👍

tagstudio/tests/qt/test_tag_widget.py
```diff
@@ -108,3 +108,27 @@ def test_tag_widget_edit(qtbot, qt_driver, library, entry_full):
...
+    tag_widget.tag_entry.returnPressed.emit()
+
+    entry = next(library.get_entries(with_joins=True)) # Update entry
+    assert len(entry.tags) == 2
+
+    # Test unmatched autocomplete
+    tag_widget.tag_completer.activated.emit("missing")
+
+    entry = next(library.get_entries(with_joins=True)) # Update entry
+    assert len(entry.tags) == 2
```
yedpodtrzitko: I'm not sure I understand the check here. Is it that searching for `arch` adds new tag to the Entry (thus len(tags) == 2), but then searching for `missing` afterwards doesnt yield any results, so the `len(tags) == 2` is still valid? 
If so, could you add a check `len(tags) == 1` before adding the first tag to confirm there is only one tag before adding `arch`? thanks.

tagstudio/src/qt/widgets/tag_box.py
```diff
@@ -73,13 +93,33 @@ def __init__(
...
-        self.add_button.clicked.connect(
-            lambda: (
-                tsp.update_tags(),
-                self.add_modal.show(),
+
+        self.add_button.clicked.connect(self.add_modal.show)
+        self.tag_entry.textChanged.connect(
+            lambda text: (
+                tsp.search_field.setText(text),
+                tsp.update_tags(text),
```
yedpodtrzitko, line 110: I dont understand the logic here - you have initiated `TagCompleter`, which contains all tag names for the autocompleter. Why do you need also query the database each time when the input value is changed?

tagstudio/src/qt/widgets/tag_box.py
```diff
@@ -73,13 +93,33 @@ def __init__(
...
+                self.tag_completer.setCompletionPrefix(text),
+                self.tag_completer.complete(),
+            )
+        )
+        self.tag_entry.returnPressed.connect(
+            lambda: self.tag_completer.activated.emit(self.tag_entry.text())
```
yedpodtrzitko, line 116: pressing Enter behaves differently than I'd expect in a scenario when no match is found:

Steps to reproduce:

- tags "Archived" and "Favorite" in DB.
- type "Favorited" (with extra `d` on the end), so there no match
- press Enter
- no tag is added, which is expected behaviour, since there was no match (thanks for fixing that)
- the value of the search field is emptied, which is not the expected behaviour when there was no match - right now I might be confused why the tag "Favorite" was not added, since I am convinced I typed there "Favorite". So in the case of no match, I would expect Enter wont empty the value.

does it make sense?

tagstudio/src/qt/widgets/tag_box.py
```diff
@@ -50,6 +63,13 @@ def __init__(
         self.base_layout.setContentsMargins(0, 0, 0, 0)
         self.setLayout(self.base_layout)
 
+        self.tag_entry = QLineEdit()
```
yedpodtrzitko, line 66: it looks weird when the input field is jumping left and right depending on how does it momentarily fit next to the existing tags. Maybe it would be better to keep the input field on a separate line, so it's always aligned on the left? Up to you.

yedpodtrzitko (commented):
it looks good, beside the Enter behaviour on no-match. good job.

tagstudio/tests/qt/test_tag_widget.py
Tyrannicodin: Yep, that's exactly it. I will add the extra check today.

tagstudio/src/qt/widgets/tag_box.py
Tyrannicodin: It was because if it didn't update, the add tag modal wouldn't have the updated tags, but I can probably change it to update only when the modal opens to reduce calls to the database.

tagstudio/src/qt/widgets/tag_box.py
Tyrannicodin: If I can figure out how, I will.

Tyrannicodin:
Changes made

yedpodtrzitko:
please fix the pipeline

yedpodtrzitko:
it's quite misbehaving 

https://github.com/user-attachments/assets/960adbbe-9aa7-48a8-a37a-4ec6d372380c

Tyrannicodin:
What. I've never seen it do that.
I'll look into it in the morning

CyanVoxel:
Closed in favor of the new tagging workflow introduced in #655 which provides the same benefits

PR #498, opened 2024-09-11
feat: make search results more ergonomic

yedpodtrzitko:
the current return type of `search_library()` is a bit unergonomic - it returns a tuple of:
`(total_count, page_items)`, so the result needs to be assigned into two items, which is error-prone

This should make it better - it returns `SearchResult` which is a wrapper for `total_count` and `page_items`, but it can still be treated just as list of items

### DIFF
--- a/tagstudio/src/core/library/alchemy/library.py
+++ b/tagstudio/src/core/library/alchemy/library.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from datetime import datetime, UTC
 import shutil
 from os import makedirs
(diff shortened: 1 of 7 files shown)


tagstudio/src/core/library/alchemy/library.py
```diff
@@ -92,6 +93,33 @@ def get_default_tags() -> tuple[Tag, ...]:
     return archive_tag, favorite_tag
 
 
+@dataclass(frozen=True)
+class SearchResult:
+    """Wrapper for search results.
+
+    :param total_count: total number of items for given query, might be different than len(items)
+    :param items: items for current page (size matches filter.page_size)
+    """
```
CyanVoxel, line 102: Just a note, and I discovered one instance of this in #322 as well: Most of the existing codebase is using Google-style docstrings rather than reST, however this is yet to be checked or enforced by a workflow. This is something I'm looking to address in #499, so I'm accepting this as-is and will defer further docstring conversation over to that one.

CyanVoxel (approved):

PR #499, opened 2024-09-12
ci: update ruff linter config

CyanVoxel:
This PR aims to make the following changes/additions to the Ruff linter workflow:
- Increase line length to 100 columns
- Enforce Google-style docstrings and formatting
- Enforce import order

Changing this workflow will require a format refactor of the codebase. I could either preform this within the same PR or open a second one to perform this.

The [CONTRIBUTING.md](https://github.com/TagStudioDev/TagStudio/blob/main/CONTRIBUTING.md) will not need to be updated, as it already outlines these recommended practices.

Discussion on these changes is welcome.


yedpodtrzitko:
Adding the missing docstring everywhere will be PITA, otherwise it looks like how I'd prefer it myself. 

I'm just not sure we need docstrings in tests, but there's probably no way how to exclude that directory from that particular rule, so let's do it.

CyanVoxel:
> Adding the missing docstring everywhere will be PITA, otherwise it looks like how I'd prefer it myself.

I attempted to ignore missing docstrings for the time being by excluding `D103`, however I realize I wasn't [thorough enough](https://docs.astral.sh/ruff/rules/#pydocstyle-d) with the rule selection here. I can exclude `D100` though `D107` to suppress these warnings for now, then docstrings can be added to existing code at a more reasonable pace over time until these exclusions are ready to be lifted. In the meantime the formatting can still be enforced for any pre-existing and future docstrings.

> I'm just not sure we need docstrings in tests, but there's probably no way how to exclude that directory from that particular rule, so let's do it.

It looks like I can ignore docstring linting in the tests directory with:
```toml
[tool.ruff.per-file-ignores]
"tagstudio/tests/**" = ["D"]
```

yedpodtrzitko:
> I could either preform this within the same PR or open a second one to perform this.

I'd do it within the same PR, to see everything actually looks and work as expected with the chosen rules.

yedpodtrzitko:
please add the reformatting commit into [.git-blame-ignore-revs ](https://graphite.dev/guides/git-blame-ignore-revs)

pyproject.toml
```diff
@@ -1,9 +1,28 @@
...
+convention = "google"
 
 [tool.ruff.lint]
-select = ["E", "F", "UP", "B", 'SIM']
-ignore = ["E402", "E501", "F541"]
+select = ["E", "F", "D", "I", "UP", "B", "SIM"]
+ignore = [
+	"E402",
+	"E501",
+	"F541",
```
yedpodtrzitko, line 16: I originally put there this rule to avoid too many unrelated changes, but since we're going nuclear now, I'd enable this rule as well:

F541 - https://docs.astral.sh/ruff/rules/f-string-missing-placeholders/

pyproject.toml
```diff
@@ -1,9 +1,28 @@
...
+[tool.ruff.lint.pydocstyle]
+convention = "google"
 
 [tool.ruff.lint]
-select = ["E", "F", "UP", "B", 'SIM']
-ignore = ["E402", "E501", "F541"]
+select = ["E", "F", "D", "I", "UP", "B", "SIM"]
+ignore = [
+	"E402",
+	"E501",
```
yedpodtrzitko, line 15: this can (should be enabled as well)

pyproject.toml
```diff
@@ -1,9 +1,28 @@
...
+
+[tool.ruff.lint.pydocstyle]
+convention = "google"
 
 [tool.ruff.lint]
-select = ["E", "F", "UP", "B", 'SIM']
-ignore = ["E402", "E501", "F541"]
+select = ["E", "F", "D", "I", "UP", "B", "SIM"]
+ignore = [
+	"E402",
```
yedpodtrzitko, line 14: probably also enable it?

pyproject.toml
```diff
@@ -1,9 +1,28 @@
...
+[tool.ruff.lint.per-file-ignores]
+"tagstudio/tests/**" = ["D"]
+
+[tool.ruff.lint.pydocstyle]
+convention = "google"
 
 [tool.ruff.lint]
-select = ["E", "F", "UP", "B", 'SIM']
-ignore = ["E402", "E501", "F541"]
+select = ["E", "F", "D", "I", "UP", "B", "SIM"]
```
yedpodtrzitko, line 12: should be added:
- `FBT` https://docs.astral.sh/ruff/rules/#flake8-boolean-trap-fbt
- `CPY` https://docs.astral.sh/ruff/rules/#flake8-copyright-cpy
- `T20` https://docs.astral.sh/ruff/rules/#flake8-print-t20

nice to have:
- `SIM` to catch unnecessarily complex code https://docs.astral.sh/ruff/rules/#flake8-simplify-sim


nice to have but will need some more work:
- `N` https://docs.astral.sh/ruff/rules/#pep8-naming-n 
- `AN001` although I'm not sure if that would be possible to apply for new code, as there is a lot of annotations missing in existing code  https://docs.astral.sh/ruff/rules/#flake8-annotations-ann

pyproject.toml
CyanVoxel: Ah yes, now's a good time to remove `F541`, I actually missed that it was still there.
`E402` shouldn't be a problem to remove, but I know there's some runtime imports for the v9.4 Blender thumbnailer... (problem for future me)
`E501` I agree should be removed, *but it doesn't mean I'm looking forward to the refactoring that it entails...*

yedpodtrzitko: >  E501 I agree should be removed, but it doesn't mean I'm looking forward to the refactoring that it entails...

that should be handled automatically by `ruff format` anyway, no?

CyanVoxel: Not in the cases of f-strings or long docstrings apparantly, which seem to make up most if not all of the cases being flagged. It's stuff that needs to happen, it's just a bit more involved.

pyproject.toml
CyanVoxel: - `FBT` seems well-intentioned but would require a *significant* amount of  additional refactoring work that's I'm not prepared for
- `CPY` seems good, looks like I just need to do some configuring work beforehand
- `T20` looks great
- `N` I like a lot, but will require a whole lot of `noqa: N802` comments for overriding native Qt methods that insist on sticking with camelCase. But I don't mind doing this and reformatting non-Qt methods that may mistakenly be using the wrong case
- `ANN001` I agree with you on, it would be great to have but adds on a lot of work to catch up the existing codebase

yedpodtrzitko: I'll do `FBT` in a separate PR

yedpodtrzitko: looks like only `FBT003` is really feasiable

yedpodtrzitko (approved):

CyanVoxel:
I've implemented the rest of the discussed rules besides `AN001` and `CPY`. `AN001` should be possible over time, there's just a lot of arguments to cover right now. I also didn't realize that the `CPY` rule is in preview and requires the `--preview` flag, so I'll probably just keep an eye out for when this fully releases.

yedpodtrzitko:
well done. I'd probably squash the commits, so this whole thing can be a single entry in `.git-blame-ignore-revs`?

CyanVoxel:
> well done. I'd probably squash the commits, so this whole thing can be a single entry in `.git-blame-ignore-revs`?

I was thinking the same 👍

yedpodtrzitko:
> I was thinking the same 👍

and by "I", I didnt really referred to myself, just so we're on the same page :) unless you want me to do that

PR #507, opened 2024-09-16
Add en.json with strings found in code

Bamowen:
This file compiles all strings found in the code in one json.
Surely I've missed some, but this should be a good enough start for the translation effort.


tagstudio/resources/translations/en.json
```diff
@@ -0,0 +1,144 @@
...
+  "generic.open_all": "Open All",
+  "generic.close_all": "Close All",
+  "generic.refresh_all": "Refresh_All",
+  "generic.apply": "Apply",
+  "generic.mirror": "Mirror",
+  "generic.search_tags": "Search Tags",
+  "build_tags.parent_tags": "Parent Tags",
+  "build_tags.add_parent_tags": "Add Parent Tags",
+  "delete_unlinked.delete_unlinked": "Delete Unlinked Entries",
+  "delete_unlinked.confirm": "Are you sure you want to delete the following %{len(self.lib.missing_files)} entries?",
```
yedpodtrzitko, line 100: @Bamowen if the underlying implementation changes (for example from variable `self.lib.missing_files` to `missing_files_count`), will it break the translation? If so, then that's quite brittle way how to make these translation string.

In such case it would be better to change the strings in code from f-string to use `format()` like this: 

```"Are you (...) {files_count} entries?".format(files_count=len(self.lib.missing_files))```, so the translation is not directly coupled to the variable name.

Tyrannicodin: I've been doing this in #447 as I go through changing the code

yedpodtrzitko: nice

PR #513, opened 2024-09-18
Update requirements.txt

myxxmikeyxx:
Changed the version to 6.7.0 because it is marked as a fixed version for the but where it says:
```
No QtMultimedia backends found. Only QMediaDevices, QAudioDevice, QSoundEffect, QAudioSink, and QAudioSource are available.
Failed to create QVideoSink "Not available"
Failed to initialize QMediaPlayer "Not available"
```
Bug Report is attached below and the change log from 6.7.0 -> 6.7.1 is listed. two CVE's were listed but they are not directly used by QT and only used if the developer is using them. I did not check to see if the code contains, but since you are trying to use sqlite DB and are not networking this should be in the clear.

https://bugreports.qt.io/browse/PYSIDE-2656
https://code.qt.io/cgit/qt/qtreleasenotes.git/about/qt/6.7.1/release-note.md
https://forum.qt.io/topic/157026/no-qtmultimedia-backends-found-on-macos-with-pyside-6-7-1


CyanVoxel:
6.7.0 has historically had issues for us including a bug involving a TypeError in the main window class, plus an issue where it doesn't play nicely with MyPy causing the workflow to always fail.
Quoting @xarvex's TypeError investigation from April:
> i am currently trying to get 6.7.0 working for PySide6, the issue stems from tagstudio/src/qt/main_window.py on line 37

> self.setupUi(self) is called, which in turn self gets used as the value to construct QWidget, where the program freaks out about it not being a QWidget
heres the thing: the signature of that method does not change, and the class self is for (Ui_MainWindow) extends QMainWindow, which extends QWidget

> there is an issue filed [here](https://bugreports.qt.io/browse/PYSIDE-2683) that has not had activity from the person who filed it, which seems to correspond to whats going on. i dont want to jump to it being a bug in the library yet though, as the possibility remains that Ui_MainWindow is not being constructed properly

However I am currently unable to reproduce the same TypeError issue with the current codebase, so maybe that particular bug isn't an issue anymore. Either way we may just need to try/wait for a future PySide version instead of downgrading to versions with preexisting bugs.

myxxmikeyxx:
No problem, was not sure if any of the other tweaks I made were worth it. I am on a Mac Studio (Last intel model) so the error I was getting on running the script could just be a me thing is someone else is running it on an M1.

CyanVoxel:
> No problem, was not sure if any of the other tweaks I made were worth it. I am on a Mac Studio (Last intel model) so the error I was getting on running the script could just be a me thing is someone else is running it on an M1.

I think there's been some reports on Linux about a similar issue, if not then I'm thinking about other Mac reports. Either way I'll keep an eye on future PySide versions to fix this. According to one of the forum posts you linked it looks like 6.7.3 might include a fix, which looks like it's scheduled for release... wait, [tomorrow?](https://wiki.qt.io/Qt_6.7_Release) Okay I'll *really* be keeping an eye out for that then...

myxxmikeyxx:
Not sure if you are able to commit just the Build_Mack)S_app.sh file but I downloaded the 9.4 branch again and tried to compile but was getting an error because the directories don't have quotes around them.

myxxmikeyxx:
Creating a new pull request. Looks like the updated Qt fixes the ffmpeg not found problem

PR #516, opened 2024-09-19
feat: add option to ignore dir content via `.ts_noindex` file

yedpodtrzitko:
- add ability to ignore directory content via `.ts_noindex` file
- if the `.ts_noindex` file is created later, remove already registered the files from the library on refresh

### DIFF
--- a/tagstudio/src/core/library/alchemy/library.py
+++ b/tagstudio/src/core/library/alchemy/library.py
@@ -15,7 +15,6 @@ from sqlalchemy import (
     and_,
     create_engine,
     delete,
-    exists,
     func,
     or_,
     select,
@@ -34,6 +33,7 @@ from ...constants import (
     TAG_ARCHIVED,
     TAG_FAVORITE,
     TS_FOLDER_NAME,
+    TS_FOLDER_NOINDEX,
 )
 from ...enums import LibraryPrefs
 from .db import make_tables
@@ -362,6 +362,8 @@ class Library:
         if not full_ts_path.exists():
             logger.info("creating library directory", dir=full_ts_path)
             full_ts_path.mkdir(parents=True, exist_ok=True)
+            # create noindex file to ignore the folder
+            (full_ts_path / TS_FOLDER_NOINDEX).touch()
 
     def add_entries(self, items: list[Entry]) -> list[int]:
         """Add multiple Entry records to the Library."""
@@ -389,10 +391,10 @@ class Library:
             session.query(Entry).where(Entry.id.in_(entry_ids)).delete()
             session.commit()
 
-    def has_path_entry(self, path: Path) -> bool:
+    def get_path_entry(self, path: Path) -> Entry | None:
         """Check if item with given path is in library already."""
         with Session(self.engine) as session:
-            return session.query(exists().where(Entry.path == path)).scalar()
+            return session.scalar(select(Entry).where(Entry.path == path))
 
     def search_library(
         self,
--- a/tagstudio/src/core/utils/refresh_dir.py
+++ b/tagstudio/src/core/utils/refresh_dir.py
@@ -4,7 +4,7 @@ from pathlib import Path
 from time import time
 
 import structlog
-from src.core.constants import TS_FOLDER_NAME
+from src.core import constants
 from src.core.library import Entry, Library
 
 logger = structlog.get_logger(__name__)
@@ -39,7 +39,11 @@ class RefreshDirTracker:
         self.files_not_in_library = []
 
     def refresh_dir(self, lib_path: Path) -> Iterator[int]:
-        """Scan a directory for files, and add those relative filenames to internal variables."""
+        """Scan a directory for changes.
+
+        - Keep track of files which are not in library.
+        - Remove files from library which are in ignored dirs.
+        """
         if self.library.library_dir is None:
             raise ValueError("No library directory set.")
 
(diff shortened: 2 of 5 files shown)


tagstudio/src/core/utils/refresh_dir.py
```diff
@@ -49,24 +53,43 @@ def refresh_dir(self, lib_path: Path) -> Iterator[int]:
...
+            # - if directory contains file `.ts_noindex` then skip the directory
+            if constants.TS_FOLDER_NOINDEX in files:
+                logger.info("TS Ignore File found, skipping", directory=root)
+                # however check if the ignored files aren't in the library; if so, remove them
+                entries_to_remove = []
+                for file in files:
+                    file_path = root / file
+                    entry_path = file_path.relative_to(lib_path)
+                    if entry := self.library.get_path_entry(entry_path):
+                        entries_to_remove.append(entry.id)
```
CyanVoxel, line 69: I'm not really a fan of the idea of deleting users' library data without asking for a confirmation first. Perhaps a safeguard could be a confirmation asking if the user wants to "delete x entries, with y of them having data" (assuming that more than 0 have any user data to begin with)?

tagstudio/src/core/utils/refresh_dir.py
```diff
@@ -49,24 +53,43 @@ def refresh_dir(self, lib_path: Path) -> Iterator[int]:
         self.files_not_in_library = []
         dir_file_count = 0
 
-        for path in lib_path.glob("**/*"):
-            str_path = str(path)
-            if path.is_dir():
+        for root, _, files in lib_path.walk():
+            if "$RECYCLE.BIN" in str(root).upper():
```
CyanVoxel, line 57: Any specific reasoning for changing from `glob()` to `walk()`?

tagstudio/src/core/library/alchemy/library.py
```diff
@@ -362,6 +362,8 @@ def verify_ts_folders(self, library_dir: Path) -> None:
         if not full_ts_path.exists():
             logger.info("creating library directory", dir=full_ts_path)
             full_ts_path.mkdir(parents=True, exist_ok=True)
+            # create noindex file to ignore the folder
+            (full_ts_path / TS_FOLDER_NOINDEX).touch()
```
CyanVoxel, line 366: It seems like the purpose of this is the leverage the `noindex` system to ignore built-in folders. I would either recommend reinforcing the presence of a `.ts_noindex` file inside this folder in case it gets removed then, or just account for the folder path without the need to reinforce and then detect the presence of a `.ts_noindex` file.

CyanVoxel (commented):
This seems like a nice way to set up folders to ignore in a portable way, especially in a way that other programs can easily integrate with. Although, I still feel that central a `.gitignore` style file with glob-like pattern matching would have the advantages of:
- Also being portable, being located inside the .TagStudio directory
- Easier to edit the directories all at once
- Easier to insert/apply templates from known project folder types (e.g. VSCode, Godot, Unity, etc.)
- Could integrate with the file extension exclusion process
	- The current GUI for the file extensions could be replaced by a combined text editor UI to avoid [conflicting rules](https://github.com/TagStudioDev/TagStudio/issues/14#issuecomment-2346850256) or multiple interfaces
		- This new GUI could also present basic examples for how to ignore file types and folders, along with showing any defaults we may include
- In conjunction with the ability to specify an external location for the .TagStudio folder, would still allow for libraries on write-only drives

I’d like to hear more of your thoughts on this.

***

On my first test I encountered a strange grid desync, but haven’t been able to reproduce it. The steps I took were as follows:
1. Start with existing library “Example”
2. Add `.ts_noindex` file to “Example/Subdirectory”
3. Open TagStudio & load library - all files are still present in grid, despite the log stating that it’s ignoring the directory:
```
2024-11-11 11:37:26 [info     ] TS Ignore File found, skipping directory=WindowsPath('F:/Example/Subdirectory')
```
4. Click on any thumbnail for a file that should otherwise be excluded
5. Preview is not rendered and the following traceback occurs:
```
2024-11-11 11:31:32 [info     ] selecting item                 append=False bridge=False grid_index=170
2024-11-11 11:31:32 [info     ] update_widgets                 selected=[170]
2024-11-11 11:31:32 [info     ] searching library              filter=FilterState(page_index=0, page_size=500, search_mode=<SearchMode.AND: 0>, tag=None, tag_id=None, id=171, path=None, name=None, query=None) query_full=SELECT entries.id, entries.folder_id, entries.path, entries.suffix
FROM entries
WHERE entries.id = 171
 LIMIT 500 OFFSET 0
2024-11-11 11:31:32 [info     ] found item                     entries=0 grid_idx=170 lookup_id=171   
  File "F:\GitHub\TagStudio/tagstudio\src\qt\ts_qt.py", line 998, in <lambda>
    lambda checked=False, index=idx: self.select_item(
                                     ^^^^^^^^^^^^^^^^^
  File "F:\GitHub\TagStudio/tagstudio\src\qt\ts_qt.py", line 946, in select_item
  File "F:\GitHub\TagStudio/tagstudio\src\qt\widgets\preview_panel.py", line 559, in update_widgets   
    self.driver.frame_content[grid_idx] = results[0]
                                          ~~~~~~~^^^
  File "F:\GitHub\TagStudio/tagstudio\src\core\library\alchemy\library.py", line 113, in __getitem__  
    return self.items[index]
           ~~~~~~~~~~^^^^^^^
IndexError: list index out of range
```
6. Upon refreshing the library, the grid no longer seems aligned with the internal data

I attempted this again with an “Example/3D” directory, but everything seemingly worked as intended the second time around.

When trying to add a `.ts_noindex` file to a directory at runtime (in this case “Example/Vector”, I noticed that refreshing did not seem to update the library to filter out these results, despite the logs saying that they were ignoring both the “Example/3D” and “Example/Vector” directories.

yedpodtrzitko:
> I’d like to hear more of your thoughts on this.

Having the include/exclude patterns in two different places -that is in the DB and in plaintext file- sounds like it will lead to scenarios where the rules will be clashing among themselves, and/or it will be complex to decide in which order the rules should be evaluated. 

This is a solution for scenario someone mentioned on Discord - that is having a directory with thousand of metadata files from some other application you want to avoid having indexed.

CyanVoxel:
> Having the include/exclude patterns in two different places -that is in the DB and in plaintext file- sounds like it will lead to scenarios where the rules will be clashing among themselves, and/or it will be complex to decide in which order the rules should be evaluated.

I've been under the impression that a `.gitignore` list would outright replace the current include/exclude lists, and it's a superset of that functionality. 
 
> This is a solution for scenario someone mentioned on Discord - that is having a directory with thousand of metadata files from some other application you want to avoid having indexed.

I'll try to find that message on the Discord. I *feel* glob patterns would be powerful enough to address that, assuming the folders containing this metadata have known and predictable paths.

yedpodtrzitko:
> I've been under the impression that a .gitignore list would outright replace the current include/exclude lists, and it's a superset of that functionality.

ok then, I will close this PR

PR #472, opened 2024-09-08
fix(PreviewPanel): Use birthtime for creation time on Mac & Windows

seakrueger:
st_ctime does not provide accurate creation time on MacOS, and as Python 3.12 is [deprecated for Windows](https://docs.python.org/3.12/library/os.html#os.stat_result.st_ctime). On these two platforms st_birthtime is used instead.
Tested on Linux and Windows

fixes #469

PR #373, opened 2024-08-23
Add support for VALVe Source 1 + 2 files

FavroiteGamers:
This PR adds support for VMF (Valve Material File) thumbnails as well as changing the plaintext to include all of these plain text formats used by Source games. 

        ".vmt",
        ".fgd",
        ".nut",
        ".cfg",
        ".conf",
        ".vdf",
        ".vcfg",
        ".gi",
        ".inf",
        ".vqlayout",
        ".qss",
        ".vsc",
        ".kv3",
        ".vsnd_template"

PR #370, opened 2024-08-23
Change "show file in explorer" to be different depending on the OS

SupKittyMeow:
Windows says explorer, Mac says Finder, and Linux just says filesystem (as I don't know what it is on Linux)

PR #522, opened 2024-09-22
feat: port thumbnail (#390) and related features to v9.5

### LINKED ISSUE #270: [BUG]: Video Playback Icons Missing in Executables

CyanVoxel:
This PR ports the following thumbnail and related PRs from the `Alpha-v9.4` branch to `main` (v9.5+):
- Blender thumbnail support (#273) by @050011-code
- Add font thumbnail preview support (#307) by @Thesacraft 
- refactor: move type constants to new media classes (#331) by @CyanVoxel 
- feat(ui): expanded thumbnail and preview features (#390) by @CyanVoxel 
- ui: "open in explorer" action follows os name (#370) by @SupKittyMeow
- feat(ui): preview support for source engine files (#373) by @FavroiteGamers
- Refactor video_player.py (Fix #270) (#274) by @CyanVoxel 
- feat(ui): show file creation/modified dates + restyle path label (#430) by @CyanVoxel 
- fix(ui): use default audio icon if ffmpeg is absent (#471) by @seakrueger 
- fix(ui): use birthtime for creation time on mac & win (#472) by @seakrueger
- *other miscellaneous commits*

Many of these features and changes were closely intertwined with the focused PR (#390) to where it didn't make sense to spend the extra time and effort to try and comb them apart just for a cleaner git history. If I missed any other PRs or attributions, please let me know.


requirements.txt
```diff
@@ -12,3 +12,9 @@ pillow-heif==0.16.0
 chardet==5.2.0
 structlog==24.4.0
 SQLAlchemy==2.0.34
+pydub==0.25.1
+mutagen==1.47.0
+numpy==1.26.4
+ffmpeg-python==0.2.0
+Send2Trash==1.8.3
```
yedpodtrzitko, line 19: this dependency looks unused

tagstudio/src/core/media_types.py
```diff
@@ -0,0 +1,495 @@
...
+        extensions: set[str],
+        is_iana: bool = False,
+    ) -> None:
+        self.media_type: MediaType = media_type
+        self.extensions: set[str] = extensions
+        self.is_iana: bool = is_iana
+
+
+class MediaCategories:
+    """Contains pre-made MediaCategory objects as well as methods to interact with them."""
```
yedpodtrzitko, line 71:
```suggestion
    """Contain pre-made MediaCategory objects as well as methods to interact with them."""
```

tagstudio/src/core/media_types.py
```diff
@@ -0,0 +1,495 @@
...
+
+    def __init__(
+        self,
+        media_type: MediaType,
+        extensions: set[str],
+        is_iana: bool = False,
+    ) -> None:
+        self.media_type: MediaType = media_type
+        self.extensions: set[str] = extensions
+        self.is_iana: bool = is_iana
```
yedpodtrzitko, line 67: Since this is just a data wrapper, I'd suggest to use `dataclass` instead. It will save you from writing the constructor and assigning the values in it manually.

```python
@dataclass(frozen=True)
class MediaCategory:
    media_type: MediaType
    extensions: set[str]
    is_iana: bool # maybe assign default value, so you dont have to initialize it explicitly half of the times?
```

example of usage eg. here:

https://github.com/TagStudioDev/TagStudio/blob/49d071cf2b2955c3a5dfb96e7df0661d9cac0b44/tagstudio/src/qt/modals/folders_to_tags.py#L33-L37

tagstudio/src/core/media_types.py
```diff
@@ -0,0 +1,495 @@
...
+        SHORTCUT_TYPES,
+        SOURCE_ENGINE_TYPES,
+        SPREADSHEET_TYPES,
+        TEXT_TYPES,
+        VIDEO_TYPES,
+    ]
+
+    @staticmethod
+    def get_types(ext: str, mime_fallback: bool = False) -> set[MediaType]:
+        """Returns a set of MediaTypes given a file extension.
```
yedpodtrzitko, line 477:
```suggestion
        """Return a set of MediaTypes given a file extension.
```

tagstudio/src/core/media_types.py
```diff
@@ -0,0 +1,495 @@
...
+        """
+        types: set[MediaType] = set()
+        # mime_guess: bool = False
+
+        for cat in MediaCategories.ALL_CATEGORIES:
+            if ext in cat.extensions:
+                types.add(cat.media_type)
+            elif mime_fallback and cat.is_iana:
+                type: str = mimetypes.guess_type(Path("x" + ext), strict=False)[0]
+                if type and type.startswith(cat.media_type.value):
```
yedpodtrzitko, line 491: `type` is a name of builtin, so it would be better to use different variable name.

tagstudio/src/core/palette.py
```diff
@@ -295,3 +334,9 @@ def get_tag_color(color_type: ColorType, color_id: TagColor) -> str:
         traceback.print_stack()
         logger.error("Color not found", color_id=color_id)
         return "#FF00FF"
+
+
+def get_ui_color(color_type: ColorType, color: str):
+    """Returns a hex value given a color name and ColorType."""
```
yedpodtrzitko, line 340:
```suggestion
    """Return a hex value given a color name and ColorType."""
```

tagstudio/src/core/palette.py
```diff
@@ -295,3 +334,9 @@ def get_tag_color(color_type: ColorType, color_id: TagColor) -> str:
         traceback.print_stack()
         logger.error("Color not found", color_id=color_id)
         return "#FF00FF"
+
+
+def get_ui_color(color_type: ColorType, color: str):
```
yedpodtrzitko, line 339: add return type annotation please

tagstudio/src/core/palette.py
```diff
@@ -283,6 +283,45 @@ class ColorType(IntEnum):
...
 }
 
+_UI_COLORS: dict = {
+    "": {
+        ColorType.PRIMARY: "#333333",
+        ColorType.BORDER: "#555555",
+        ColorType.LIGHT_ACCENT: "#FFFFFF",
+        ColorType.DARK_ACCENT: "#1e1e1e",
+    },
+    "red": {
```
yedpodtrzitko, line 293: how about using an enum for the keys instead of string. Then in the places where these variables are used, it would be obvious where that value comes from, and that it's not just some random string. Something like this:


```python
class UIColors(enum.StrEnum):
    DEFAULT = ""
    RED = "RED"
    GREEN = "GREEN"
    ...
```

and then


```suggestion
    UIColors.RED: {
```

tagstudio/src/qt/helpers/silent_popen.py
```diff
@@ -0,0 +1,64 @@
+import subprocess
```
yedpodtrzitko, line 1: add a docstring here what is this file/function good for?

yedpodtrzitko (commented):
I'd recommend doing smaller PRs instead, which would be easier to review.

requirements.txt
CyanVoxel: Send2Trash was accidentally added due to its inclusion in future a v9.4 PR, fixed 👍

CyanVoxel:
Implemented suggested feedback, thanks for the review 👍

tagstudio/src/qt/widgets/video_player.py
```diff
@@ -115,14 +118,21 @@ def __init__(self, driver: "QtDriver") -> None:
...
         open_file_action = QAction("Open file", self)
         open_file_action.triggered.connect(self.opener.open_file)
-        open_explorer_action = QAction("Open file in explorer", self)
+
+        system = platform.system()
+        open_explorer_action = QAction("Open in explorer", self)  # Default, usually Linux
+        if system == "Darwin":
+            open_explorer_action = QAction("Reveal in Finder", self)
+        elif system == "Windows":
+            open_explorer_action = QAction("Open in Explorer", self)
```
yedpodtrzitko, line 134: ❓ this code already exists in `preview_panel.py`, cant it be reused here somehow?

yedpodtrzitko (commented):
almost there :)

tagstudio/src/qt/resource_manager.py
```diff
@@ -23,9 +24,26 @@ def __init__(self) -> None:
...
             with open(Path(__file__).parent / "resources.json", encoding="utf-8") as f:
                 ResourceManager._map = ujson.load(f)
-                logger.info("resources registered", count=len(ResourceManager._map.items()))
+                logger.info(
+                    f"[ResourceManager] {len(ResourceManager._map.items())} resources registered"
+                )
             ResourceManager._initialized = True
 
+    @staticmethod
+    def get_path(id: str) -> Path | None:
```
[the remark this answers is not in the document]
CyanVoxel: This gets used by another feature in 9.4 (file deleting) - I could remove it here but I would be adding it back in the next PR here anyways

tagstudio/src/qt/widgets/preview_panel.py
```diff
@@ -91,22 +96,69 @@ def __init__(self, library: Library, driver: "QtDriver"):
...
+        self.trash_term: str = "Trash"
+        if platform.system() == "Windows":
+            self.trash_term = "Recycle Bin"
+        self.delete_action = QAction(f"Send file to {self.trash_term}", self)
+
+        self.open_explorer_action = QAction("Open in explorer", self)  # Default text (Linux, etc.)
+        if platform.system() == "Darwin":
+            self.open_explorer_action = QAction("Reveal in Finder", self)
+        elif platform.system() == "Windows":
+            self.open_explorer_action = QAction("Open in Explorer", self)
```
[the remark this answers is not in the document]
CyanVoxel: I've removed the trash_term/delete_action as those were part of a separate PR that got mixed in here. Otherwise I've split out the explorer action string so it (and future strings) can be reused

CyanVoxel:
@yedpodtrzitko Implemented the new feedback 👍 Let me know how you feel about the PlatformStrings file and if there's anything else I may have missed

tagstudio/src/qt/platform_strings.py
```diff
@@ -0,0 +1,16 @@
...
+import platform
+
+
+class PlatformStrings:
+    open_file_str: str = "Open in file explorer"
+
+    if platform.system() == "Windows":
+        open_file_str = "Open in Explorer"
+    elif platform.system() == "Darwin":
+        open_file_str = "Reveal in Finder"
```
yedpodtrzitko, line 16: Having the logic directly in the class isnt the usual way how to do it, but I think we can live with it here :thinking:

yedpodtrzitko (approved):

tagstudio/src/qt/platform_strings.py
CyanVoxel: For future reference what would be the expected way to do something like this? I was mulling over different implementations and just went with this since it minimizes the performance impact of the checks while keeping the imports somewhat clean

yedpodtrzitko: I have to board plane now, so just very quickly - I would probably go slightly different way - rather than doing everything in one class, I'd create a class with common phrases, and then inherit from that class like `WindowStrings` where platform-specific phrases would be overriden.

```
class SharedStrings:
   open_file_str: str = 'something'


class WindowsStrings(SharedStrings):
    open_file_str: str = 'windows something'  

TRANSLATIONS = # `platform.system()` logic to decide which class to use

```

PR #527, opened 2024-09-28
Fix typo in fix_unlinked.py

Bamowen:
"De&lete" should be "&Delete"


CyanVoxel:
This is intentional, otherwise the <kbd>Alt</kbd> shortcut conflicts with "&Done" in the same menu, both mapping to <kbd>Alt</kbd>+<kbd>D</kbd>.

Bamowen:
> This is intentional, otherwise the Alt shortcut conflicts with "&Done" in the same menu, both mapping to Alt+D.

Ok, didn't know. Seemed odd. Closing.

Issue #450, opened 2024-09-05
[Feature Request]: Audio Playback


AwesomeParley:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

Basically the same way that videos have playback when you press on them, audio should also playback. For right now, we have to press on the audio file, then press the icon on the right to open up another app to play the audio, instead, it should just play in the app.

csponge:
Hi there!

I'm interested in this feature as well. I'll probably take a crack at implementing it, unless someone else has already done so.

csponge:
Just wanted to give an update. I haven't forgotten about this!

I currently have audio files playing. I have a slider to change position, and buttons for play, pause, and mute.

There are a few more minor things to fix, but I'm hoping to have that done relatively soon.

After I get that all working, I think it would be cool to add a queue feature in. Not sure if that makes sense to include in this issue. Any thoughts on that?

CyanVoxel:
> Just wanted to give an update. I haven't forgotten about this!
> 
> I currently have audio files playing. I have a slider to change position, and buttons for play, pause, and mute.
> 
> There are a few more minor things to fix, but I'm hoping to have that done relatively soon.
> 
> After I get that all working, I think it would be cool to add a queue feature in. Not sure if that makes sense to include in this issue. Any thoughts on that?

That sounds awesome! Not sure about a queue feature though, I feel that's outside the scope of the program. Playing back audio files is great for previewing but queues are more so the territory of a music player in my opinion.

AwesomeParley:
> Just wanted to give an update. I haven't forgotten about this!
> 
> I currently have audio files playing. I have a slider to change position, and buttons for play, pause, and mute.
> 
> There are a few more minor things to fix, but I'm hoping to have that done relatively soon.

Awesome! I'm glad that it's being worked on and seems to be working for you. 

> After I get that all working, I think it would be cool to add a queue feature in. Not sure if that makes sense to include in this issue. Any thoughts on that?

I do agree that would be very cool, but seeing as CyanVoxel thinks it's outside the scope of the project, we can maybe wait a bit before adding that. I do believe it would be useful, but there's a lot of stuff we still have to do that's more important. 

Thanks for working on it!

Cool-Game-Gev:
[...]

PR #529, opened 2024-09-29
feat: Audio Playback

csponge:
This PR adds the ability to play audio files. Additionally, it adds a few new UI elements for interacting with audio. Below is a list of the newly added elements:

- Slider (used for changing the current position in the file)
- Play/Pause
- Mute/Unmute

**NOTE:** There are a few bugs to work out, but I think they would be better suited for new issues because they affect more than just the audio player.

Refs: https://github.com/TagStudioDev/TagStudio/issues/450

### DIFF
--- a/tagstudio/src/core/constants.py
+++ b/tagstudio/src/core/constants.py
@@ -57,7 +57,6 @@ VIDEO_TYPES: list[str] = [
 ]
 AUDIO_TYPES: list[str] = [
     ".mp3",
-    ".mp4",
     ".mpeg4",
     ".m4a",
     ".aac",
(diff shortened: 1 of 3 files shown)


CyanVoxel:
Any reason why this was closed?

Nginearing:
> Any reason why this was closed?

Seems like they didn't realise deleting their fork would close the commit.

CyanVoxel:
> Ah, that's my bad. Didn't realize that would happen.
> 
> Would it be possible to fix this by recreating the branch? If not, maybe I can open a new PR?
> 
> Sorry for the confusion.

I believe a new PR would be needed if the original fork and/or branch is deleted, unless there's a way to undo that in a way that GitHub is pleased with that I'm unaware of

csponge:
Okay, I'll work on making a new PR.

Again, sorry for the confusion.

PR #534, opened 2024-10-07
fix: subtags/parent tags & aliases update the UI for building a tag

DandyDev01:
The following has moved to https://github.com/TagStudioDev/TagStudio/pull/596
<del>subtags can be added and removed when creating or editing a Tag
 <del>  - parent tags would not be listed before 
 <del>  - parent tags could not be removed before

<del>can exclude tags from the tag search panel
 <del>  -   it was possible to make a tag parent itself before. Now when selecting a parent tag, the tag being created or edited is <del>excluded from the tag search panel, so it cannot be added as a parent to itself. 
<del>   - before when a tag could parent itself, that tag would be corrupted and users would be unable to edit the tag or add it <del>as a parent to another tag (this would corrupt that tag too)
  
  <del> - aliases can be removed and added
 <del>  - aliases can be edited
<del>   - added support for display name on Tags. This is a parity feature from the json DB.</del>

This PR is meant to update the UI for aliases to use a table like the exclude file extensions.

changed how aliases are managed (UI is a table):
image


tagstudio/tests/qt/test_tag_widget.py
```diff
@@ -99,12 +98,14 @@ def test_tag_widget_edit(qtbot, qt_driver, library, entry_full):
...
-    panel = tag_box_widget.edit_modal.widget
-    assert isinstance(panel, BuildTagPanel)
-    assert panel.tag.name == tag.name
-    assert panel.name_field.text() == tag.name
+    # NOTE: currently tag_box_widget.edit_modal is never set because its
+    #       edit_tag method is never called
+    # panel = tag_box_widget.edit_modal.widget
+    # assert isinstance(panel, BuildTagPanel)
+    # assert panel.tag.name == tag.name
+    # assert panel.name_field.text() == tag.name
```
yedpodtrzitko, line 111: commenting out existing tests isnt really convincing that everything works as expected

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "cSpell.words": [
+    "Abendshien",
+    "qtbot",
+    "Voxel"
+  ]
+}
```
yedpodtrzitko, line 7: I'd say we want to have whole `.vscode` in `.gitignore`

yedpodtrzitko (changes requested):
please add tests for newly added methods, thanks.

.vscode/settings.json
DandyDev01: .vscode is now ignored

yedpodtrzitko:
@DandyDev01 the PR is still marked as Draft, so am I supposed to review it already?

DandyDev01:
> @DandyDev01 the PR is still marked as Draft, so am I supposed to review it already?

@yedpodtrzitko Yes, its ready. Sorry I thought I took it off draft when I sent the review request.

seakrueger:
If adding a tag as a parent of itself corrupts the database entry, IMO a check should be added to make sure that isn't possible rather than hiding the tag in the UI.

yedpodtrzitko:
looks good overall, I just noticed some issue with the alias:

- open dialog for adding tag
- add new input for Alias
- click `[-]` to remove alias (see screenshot below)
- it will produce the following error:

```
  File "/Users/yed/dev/python/TagStudio/tagstudio/src/qt/modals/build_tag.py", line 202, in <lambda>
    new_field = TagAliasWidget(on_remove_callback=lambda a="": self.remove_alias_callback(a))
                                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yed/dev/python/TagStudio/tagstudio/src/qt/modals/build_tag.py", line 220, in remove_alias_callback
    self.alias_ids.remove(alias_id)
KeyError: None
```







another issue with the same button when editing existing tag - clicking the `[-]` on alias seemingly doesnt do anything (ie. the input doesnt disappear), but when saving it, it will actually remove it as expected

yedpodtrzitko (changes requested):
Also could you please squash the commits? thanks.

yedpodtrzitko:
@CyanVoxel can you approve the CI run here?

yedpodtrzitko (approved):

CyanVoxel:
Thank you for this PR! I apologize for not getting around to commenting on it earlier.

From my understanding, this PR is doing three things:
1. Connecting the add/edit tag UI for tag aliases and subtags to the SQL backend
2. Reworking the tag alias UI
3. Fixing an undocumented bug where a tag could be assigned as its own parent

However I also understand that this PR is *not:*
- Implementing the functionality of tag aliases and subtags (i.e. integrating them into the search system as with 9.4 and below)

The DB connection is still an appreciated addition, and is one step closer to reaching tag feature parity between 9.4 and 9.5. I'd just like to confirm that this was your intention rather than a full re-implementation of these features. If you *did* intend to re-implement these features so that they're functioning on par with 9.4, then please let me know.

As for the alias UI changes, I have some feedback on the direction taken here. While the previous newline separated textbox wasn't completely ideal, I feel that this individually separated line approach makes things a bit more cumbersome to use. My thoughts are:
- New aliases can no longer be quickly created via an Enter press
- More vertical screen space is now being taken up by the separated lines
- The (-) button overlaps with the beginning of the text, making the text difficult to read and interact with
- The (-) button is not aligned with the textbox and does not match the rest of the app theme

My recommendations to solve these issues while keeping the spirit of the changes that improve upon the original implementation would be:
- Use some form of table for the alias list, similar to the Include/Exclude filetypes table
- Allow the use of the Enter key to automatically create new aliases
- Keep a persistent (-) button to the left of each alias, with the default Qt button styling
- Allow a backspace on an entry alias text box to remove the alias

DandyDev01:
@CyanVoxel 
Yes that is correct, I did not have intentions to integrate this into the search system as I don't know any of the plans for the search system.

In terms of your feedback, the following is my proposed implementation. Please let me know if you approve of them.

- New aliases can no longer be quickly created via an Enter press
       - create a shortcut CTRL+n+a to add new aliases for when no other alias has focus. When an alias text box has focus, pressing enter will add a new text box and give it focus
- More vertical screen space is now being taken up by the separated lines 
       - rather than use the include/exclude table, I suggest having the aliases behave like the tags field on entries. I would also make this change to the parent tags. This would avoid having multiple interaction paradigms for tags, helping to keep things simple.
         
image


- The (-) button overlaps with the beginning of the text, making the text difficult to read and interact with
       - your recommendation "Keep a persistent (-) button to the left of each alias, with the default Qt button styling" or on hover will add a button to the left that does not overlap anything.
- Allow a backspace on an entry alias text box to remove the alias
      - create a shortcut CTRL+ DEL   XOR    CTRL + d + a  to remove the alias that has focus. Having just backspace could lead to user frustration when editing an alias as they might just delete the alias.

yedpodtrzitko:
how do I make a query using the tag alias?

DandyDev01:
> how do I make a query using the tag alias?

Search integration is not meant to be a part of this PR, but I have just added it. 

Alias
1. create a tag
2. add an alias
3. add the tag to an entry
4. search for the alias

Subtag
1. create two tags
2. add one tag as a parent to the other
3. add the tag with a parent to an entry
4. search for the parent tag

CyanVoxel:
> @CyanVoxel Yes that is correct, I did not have intentions to integrate this into the search system as I don't know any of the plans for the search system. **edit: I ended up adding search integration for aliases and subtags**

I’ll test this for feature parity against 9.4 then 👍

> In terms of your feedback, the following is my proposed implementation. Please let me know if you approve of them.
>
> - New aliases can no longer be quickly created via an Enter press
> - ✔create a shortcut CTRL+a to add new aliases for when no other alias has focus. When an alias text box has focus, pressing enter will add a new text box and give it focus

The “created via an Enter press” was more so referring to the old behavior when the alias box was already selected, although we *do* want interfaces like this to be support full keyboard navigation down the line, so I’ll continue this thought with that in mind.

I don’t feel that an additional shortcut is necessary here, and would add an extra layer of confusion and complexity (ex. CTRL+A is typically reserved for selecting all text/elements on a page). A simpler solution in my mind would be the ability to use the arrow/maybe tab key(s) to navigate up and down through the UI to select different fields and buttons. If using the table method as mentioned before, then the table could be selected via the arrow keys and Enter could be used to create a new row from there. In the case of a different approach with a “+” button, then that button could be what the user presses Enter on to activate it (in addition to and as an alternative to a mouse click).

> - More vertical screen space is now being taken up by the separated lines
> - ✔rather than use the include/exclude table, I suggest having the aliases behave like the tags field on entries. I would also make this change to the parent tags. This would avoid having multiple interaction paradigms for tags, helping to keep things simple.
> 
> image

I appreciate the idea of having the two “lists” of items in the tag screen be visually similar, however there’s some fundamental differences between tags and aliases that I feel warrant having the UI style of aliases be different:
- Aliases are editable strings of text, rather than representations of multi-field objects that require a separate interface to edit (i.e. tags)
- Strings of text tend to be easier to edit and compare when they’re left aligned (in English’s case) and separated by newlines rather than being wrapped on the same line
- The UX for editing tags is different from editing aliases/text, which would make it confusing to present two identical interface styles that act very differently from each other

> - The (-) button overlaps with the beginning of the text, making the text difficult to read and interact with
> - ✔your recommendation "Keep a persistent (-) button to the left of each alias, with the default Qt button styling" or on hover will add a button to the left that does not overlap anything.

No hover, just have a minus button always to the left for each Alias:
image
*(This is an example of the layout, not the styling)*

> - Allow a backspace on an entry alias text box to remove the alias
> -✔ create a shortcut  CTRL + d  to remove the alias that has focus. Having just backspace could lead to user frustration when editing an alias as they might just delete the alias.

(Same thoughts as above with the additional shortcut) I was referring to having backspace only remove the entry when the field became empty, meaning only empty fields would be deleted once backspace was pressed.

tagstudio/src/qt/helpers/file_opener.py
```diff
@@ -34,11 +34,11 @@ def open_file(path: str | Path, file_manager: bool = False):
             normpath = Path(path).resolve().as_posix()
             if file_manager:
                 command_name = "explorer"
-                command_args = '/select,"' + normpath + '"'
+                command_arg = '/select,"' + normpath + '"'
                 # For some reason, if the args are passed in a list, this will error when the
                 # path has spaces, even while surrounded in double quotes.
                 subprocess.Popen(
-                    command_name + command_args,
```
CyanVoxel, line 41: What are the changes here for?

tagstudio/src/qt/modals/build_tag.py
```diff
@@ -155,60 +169,163 @@ def __init__(self, library: Library, tag: Tag | None = None):
...
+            new_item = QTableWidgetItem(alias_name)
+            new_item.setData(Qt.UserRole, alias_id)  # type: ignore
+
+            self.aliases_table.insertRow(row)
+            self.aliases_table.setItem(row, 1, new_item)
+            self.aliases_table.setCellWidget(row, 0, remove_btn)
 
     def set_tag(self, tag: Tag):
+        self.tag = tag
+
```
CyanVoxel, line 298: Accidental duplicate line here?
```suggestion
```

tagstudio/src/qt/widgets/tag.py
```diff
@@ -29,6 +105,7 @@ def __init__(
         tag: Tag,
         has_edit: bool,
         has_remove: bool,
+        tag_display_name: str = None,
```
CyanVoxel, line 108: `tag_display_name` can be made concise as `display_name`

CyanVoxel (commented):
This new UI approach is definitely a step in the right direction for what I was trying to convey, but still breaks some of the previous ease of use. In v9.4 pressing the <kbd>Enter</kbd> key would in effect complete one alias and move your cursor to start typing another, just with a single key press. Yes, this is just writing text on a new line thanks to the return character, but in effect it made it very quick and seamless to create new aliases.

In this PR, hitting the <kbd>Enter</kbd> key *will* complete your current alias, but it will not create a new one. It appears a *second* <kbd>Enter</kbd> press will generate a new alias slot, however it is not pre-selected - meaning the user now has no choice but to use a mouse to select the new alias box. Furthermore, it seems that the new alias box is inserted in a seemingly random order in the table. I've seen it be added at the bottom, at the top, and in the middle so far.
image

Also I noticed that blank alias boxes are allowed here, when before they were removed upon saving the tag - I would suggest not allowing empty aliases here as well.

As for the subtag and alias searching, it seems to be working as intended from my initial testing. We should be revamping the search in #595 soon anyway, but this will be a much-need stopgap.

Display names for tags seem to also behave strangely, where editing tags in the tag management panel seems to update display names to show randomly different subtags, if any at all. I don't think I've seen display names appear in the preview panel, either. My guess is that the SQL database doesn't enforce an order for subtags like the JSON db did, so queries may return subtags in different orders. I honestly don't see a quick fix for this, and would likely involve requiring a new piece of UI to let the user select which subtag they'd like as the display name specifier along with a new column in the database (probably out of the scope of this PR, and might be best served as part of a larger tag interface rework).

***

Honestly I would *very much* like to be able to pull the database and search work in this right away, while still giving the UI changes some time to be iterated upon. Having subtags and aliases in the database is the final piece required for the JSON to SQL migration system for 9.5 before testing begins. Would it be possible to split this into another PR, with one focusing on the backend and the other focusing on the frontend?

*And not to sound overbearing, but one last note: If you find yourself wanting to mark this or a future PR as "not ready yet", please mark it as a draft rather than close and re-open it. Closing and re-opening a PR causes issues for me pulling it, which is why it's taken me some time to get back to reviewing this. I was under the impression that I added this notice in the CONTRIBUTING.md but was mistaken, so I'll be adding it in there soon regardless.*

DandyDev01:
> This new UI approach is definitely a step in the right direction for what I was trying to convey, but still breaks some of the previous ease of use. In v9.4 pressing the Enter key would in effect complete one alias and move your cursor to start typing another, just with a single key press. Yes, this is just writing text on a new line thanks to the return character, but in effect it made it very quick and seamless to create new aliases.
> 
> In this PR, hitting the Enter key _will_ complete your current alias, but it will not create a new one. It appears a _second_ Enter press will generate a new alias slot, however it is not pre-selected - meaning the user now has no choice but to use a mouse to select the new alias box. Furthermore, it seems that the new alias box is inserted in a seemingly random order in the table. I've seen it be added at the bottom, at the top, and in the middle so far. image
> 
> Also I noticed that blank alias boxes are allowed here, when before they were removed upon saving the tag - I would suggest not allowing empty aliases here as well.
> 
> As for the subtag and alias searching, it seems to be working as intended from my initial testing. We should be revamping the search in #595 soon anyway, but this will be a much-need stopgap.
> 
> Display names for tags seem to also behave strangely, where editing tags in the tag management panel seems to update display names to show randomly different subtags, if any at all. I don't think I've seen display names appear in the preview panel, either. My guess is that the SQL database doesn't enforce an order for subtags like the JSON db did, so queries may return subtags in different orders. I honestly don't see a quick fix for this, and would likely involve requiring a new piece of UI to let the user select which subtag they'd like as the display name specifier along with a new column in the database (probably out of the scope of this PR, and might be best served as part of a larger tag interface rework).
> 
I'll get this all sort as soon as possible and try my best to be more thorough with my testing before having a review done. 

> Honestly I would _very much_ like to be able to pull the database and search work in this right away, while still giving the UI changes some time to be iterated upon. Having subtags and aliases in the database is the final piece required for the JSON to SQL migration system for 9.5 before testing begins. Would it be possible to split this into another PR, with one focusing on the backend and the other focusing on the frontend?
> 

I'd be happy to split this into separate PRs. Would you want the back-end PR to only touch on the back-end or would it be okay to include changes to the front end that have it in a better functioning state? I am asking because I have a branch with minor changes to the front-end. It uses shortcuts we don't want and displays the aliases in a way we don't want, but the process of creating and updating tags with this UI is fairly nice until the front-end PR is ready.

If you would be okay with the minor front-end changes, I will create a new PR using that branch and rename this PR so we know it is about the front-end and mark it as a draft.

> _And not to sound overbearing, but one last note: If you find yourself wanting to mark this or a future PR as "not ready yet", please mark it as a draft rather than close and re-open it. Closing and re-opening a PR causes issues for me pulling it, which is why it's taken me some time to get back to reviewing this. I was under the impression that I added this notice in the CONTRIBUTING.md but was mistaken, so I'll be adding it in there soon regardless._

This is not overbearing at all. I'd like to apologize for the closing and re-opening, this happened twice due to reading error on my part. I am sorry for the inconvenience this has caused.

CyanVoxel:
> I'd be happy to split this into separate PRs. Would you want the back-end PR to only touch on the back-end or would it be okay to include changes to the front end that have it in a better functioning state? I am asking because I have a branch with minor changes to the front-end. It uses shortcuts we don't want and displays the aliases in a way we don't want, but the process of creating and updating tags with this UI is fairly nice until the front-end PR is ready.
>
> If you would be okay with the minor front-end changes, I will create a new PR using that branch and rename this PR so we know it is about the front-end and mark it as a draft.

That sounds like it would work great! A bit of functioning UI stuff will be okay, even with some shortcuts in place. With the backend additions pulled into `main` I can build on top of that to complete the work I'm doing on the [`json-to-sql`](https://github.com/TagStudioDev/TagStudio/tree/json-to-sql) branch in the meantime, which is currently the bottleneck of the project release cycle.

> This is not overbearing at all. I'd like to apologize for the closing and re-opening, this happened twice due to reading error on my part. I am sorry for the inconvenience this has caused.

No worries! It's my fault for not having it written down anywhere, and I know sometimes things like that can happen unintentionally too. Really it's a bug on GitHub's part that's at fault here...

Thank you so much for your work on this so far, it's going to be a huge help! And I'm very much looking forward to how the final alias list turns out!

DandyDev01:
> In this PR, hitting the Enter key _will_ complete your current alias, but it will not create a new one. It appears a _second_ Enter press will generate a new alias slot, however it is not pre-selected - meaning the user now has no choice but to use a mouse to select the new alias box. Furthermore, it seems that the new alias box is inserted in a seemingly random order in the table. I've seen it be added at the bottom, at the top, and in the middle so far. 
> 
this was happening because a set was being used to store ids and sets are unordered. I changed it to a list and that solved the issue.


> Display names for tags seem to also behave strangely, where editing tags in the tag management panel seems to update display names to show randomly different subtags, if any at all. I don't think I've seen display names appear in the preview panel, either. My guess is that the SQL database doesn't enforce an order for subtags like the JSON db did, so queries may return subtags in different orders. I honestly don't see a quick fix for this, and would likely involve requiring a new piece of UI to let the user select which subtag they'd like as the display name specifier along with a new column in the database (probably out of the scope of this PR, and might be best served as part of a larger tag interface rework).
> 
The issue here is the same as the above. A set is being used to store parent tags, because sets are unordered this results in getting a random element at index 0 each time. This issue was also solved by using a list instead of the original set.

CyanVoxel (changes requested):
The new alias UI seems to work great! I have no additional comments or requests for that 👍 

I noticed that you tried getting the display names working though, which has a couple of issues:
1. The SQL database structure does not support a consistent parent tag order. Unfortunately switching from a set to a list will not work as a substitute for what will ultimately be needed in the end: which is a an additional db column referencing which parent tag ID should be use to disambiguate it as the display name.
2. These display names don't seem to function regardless, causing an exception on my machine when trying to access the tag manager panel:
```
  File "/Users/cyanvoxel/Files/GitHub/TagStudio/tagstudio/src/qt/ts_qt.py", line 368, in <lambda>
    tag_database_action.triggered.connect(lambda: self.show_tag_database())
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cyanvoxel/Files/GitHub/TagStudio/tagstudio/src/qt/ts_qt.py", line 666, in show_tag_database
    self.modal.show()
  File "/Users/cyanvoxel/Files/GitHub/TagStudio/tagstudio/src/qt/modals/tag_database.py", line 119, in showEvent
    self.update_tags()
  File "/Users/cyanvoxel/Files/GitHub/TagStudio/tagstudio/src/qt/modals/tag_database.py", line 87, in update_tags
    display_name = tag.display_name
                   ^^^^^^^^^^^^^^^^
  File "/Users/cyanvoxel/Files/GitHub/TagStudio/tagstudio/src/core/library/alchemy/models.py", line 81, in display_name
    if len(self.parent_tags) == 0:
           ^^^^^^^^^^^^^^^^
  File "/Users/cyanvoxel/Files/GitHub/TagStudio/.venv/lib/python3.12/site-packages/sqlalchemy/orm/attributes.py", line 566, in __get__
    return self.impl.get(state, dict_)  # type: ignore[no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cyanvoxel/Files/GitHub/TagStudio/.venv/lib/python3.12/site-packages/sqlalchemy/orm/attributes.py", line 1086, in get
    value = self._fire_loader_callables(state, key, passive)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cyanvoxel/Files/GitHub/TagStudio/.venv/lib/python3.12/site-packages/sqlalchemy/orm/attributes.py", line 1121, in _fire_loader_callables
    return self.callable_(state, passive)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cyanvoxel/Files/GitHub/TagStudio/.venv/lib/python3.12/site-packages/sqlalchemy/orm/strategies.py", line 918, in _load_for_state
    raise orm_exc.DetachedInstanceError(
sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <Tag at 0x171e0ea80> is not bound to a Session; lazy load operation of attribute 'parent_tags' cannot proceed (Background on this error at: https://sqlalche.me/e/20/bhk3)
```


I would suggest just reverting any additions regarding display names here. If you would like to eventually tackle them, I would also suggest taking that on in a separate PR *(it's gonna involve some nasty work though so I don't blame you if you'd want to pass on that)*

And finally, the bug I mentioned [here](https://github.com/TagStudioDev/TagStudio/pull/596#issuecomment-2487661201) on #596 is still present. Since this is currently a live bug on main now I would suggest trying to fix that in a new PR that can be merged in the meantime, or at least let me know if you wouldn't like to work on it so that I can open it up to others.

But other than the backend stuff, the UI and UX improvements to the alias list are great!

DandyDev01:
> I noticed that you tried getting the display names working though, which has a couple of issues:

>     2. These display names don't seem to function regardless, causing an exception on my machine when trying to access the tag manager panel:
> 
> I would suggest just reverting any additions regarding display names here. If you would like to eventually tackle them, I would also suggest taking that on in a separate PR _(it's gonna involve some nasty work though so I don't blame you if you'd want to pass on that)_
> 
I've just removed them, as you said that is definitely something for a separate PR

> And finally, the bug I mentioned [here](https://github.com/TagStudioDev/TagStudio/pull/596#issuecomment-2487661201) on #596 is still present. Since this is currently a live bug on main now I would suggest trying to fix that in a new PR that can be merged in the meantime, or at least let me know if you wouldn't like to work on it so that I can open it up to others.
> 
I'll create a PR for those changes then

tagstudio/src/core/library/alchemy/models.py
```diff
@@ -52,8 +52,8 @@ class Tag(Base):
 
     parent_tags: Mapped[set["Tag"]] = relationship(
         secondary=TagSubtag.__tablename__,
-        primaryjoin="Tag.id == TagSubtag.child_id",
-        secondaryjoin="Tag.id == TagSubtag.parent_id",
+        primaryjoin="Tag.id == TagSubtag.parent_id",
+        secondaryjoin="Tag.id == TagSubtag.child_id",
```
CyanVoxel, line 56: Any specific reason for this change?

tagstudio/src/core/library/alchemy/library.py
```diff
@@ -618,7 +618,6 @@ def search_tags(
                 results=len(res),
             )
 
-            session.expunge_all()
```
CyanVoxel, line 621: Any reason for this as well?

tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -12,7 +12,6 @@
 
 import cv2
 import numpy as np
-import pillow_jxl  # noqa: F401
```
CyanVoxel, line 15: Was removing this import accidental? This is used for JPEG XL support

CyanVoxel:
Not sure why the MyPy check is failing here as it passes on my machines, looking into it now.
Maybe this just needs an update from `main`?

CyanVoxel:
The MyPy workflow is failing on this, though I still don't know why I can't reproduce it locally:
```
/home/runner/work/TagStudio/TagStudio/tagstudio/src/qt/modals/tag_database.py:87:32: error: Argument "subtag_ids" to "add_tag" of "Library" has incompatible type "list[int]"; expected "set[int] | None"  [arg-type]
/home/runner/work/TagStudio/TagStudio/tagstudio/src/qt/modals/tag_database.py:88:33: error: Argument "alias_names" to "add_tag" of "Library" has incompatible type "list[str]"; expected "set[str] | None"  [arg-type]
/home/runner/work/TagStudio/TagStudio/tagstudio/src/qt/modals/tag_database.py:89:31: error: Argument "alias_ids" to "add_tag" of "Library" has incompatible type "list[int]"; expected "set[int] | None"  [arg-type]
Found 3 errors in 1 file (checked 109 source files)'
  ```

CyanVoxel:
> The MyPy workflow is failing on this, though I still don't know why I can't reproduce it locally:
> 
> ```
> /home/runner/work/TagStudio/TagStudio/tagstudio/src/qt/modals/tag_database.py:87:32: error: Argument "subtag_ids" to "add_tag" of "Library" has incompatible type "list[int]"; expected "set[int] | None"  [arg-type]
> /home/runner/work/TagStudio/TagStudio/tagstudio/src/qt/modals/tag_database.py:88:33: error: Argument "alias_names" to "add_tag" of "Library" has incompatible type "list[str]"; expected "set[str] | None"  [arg-type]
> /home/runner/work/TagStudio/TagStudio/tagstudio/src/qt/modals/tag_database.py:89:31: error: Argument "alias_ids" to "add_tag" of "Library" has incompatible type "list[int]"; expected "set[int] | None"  [arg-type]
> Found 3 errors in 1 file (checked 109 source files)'
> ```

I'm suspecting that something is wrong with the workflow and isn't running on this PR properly, as these line numbers only make sense when looking at the current `main` code and not this PR.

CyanVoxel:
So it's not a workflow issue - MyPy just creates a new merge commit and sees that the type hint changes here are conflicting with a call to `add_tag()` that uses lists that was [merged into main](https://github.com/TagStudioDev/TagStudio/commit/2903dd22c45c02498687073d075bb88886de6b62) a couple days ago. MyPy can't comment on what's wrong here since this PR isn't up to date with main, and thus doesn't have access to the new code that conflicts with these changes.

CyanVoxel (approved):
Thank you for all your work on this! I apologize for the long turnaround time and roundabout PRs for this, but I appreciate you working with me and I love the final results here!

PR #536, opened 2024-10-08
fix: dont check db version with new library

### LINKED ISSUE #535: [Bug]: Can't Create Libraries
### Checklist

### TagStudio Version

main branch

Windows 10

### Description

Attempting to create a library results in a version mismatch error instead of falling back to new library creation. This is likely due to the code introduced in #503 that creates a library version mismatch error when no version value is found, which would also be true if there is no library in the first place.

### Expected Behavior

Creating a new library/opening a library at a location which does not already contain a library should result in a new library being created.

### Steps to Reproduce

1. Go to File -> Create/Open Library
2. Select a location that does not already contain a TagStudio library

Alternative:
2. Pass "-o" "<path of library that does not exist>" as launch arguments before running the program

### Logs

```
2024-10-07 15:48:37 [info     ] resources registered           count=24
2024-10-07 15:48:37 [info     ] Config File not specified, using default one filename=C:/Users/Faceboji/AppData/Roaming/TagStudio/TagStudio.ini
2024-10-07 15:48:41 [info     ] update_widgets                 selected=[]
2024-10-07 15:48:41 [info     ] opening library                connection_string=sqlite:///F:\Example 2\.TagStudio\ts_library.sqlite library_dir=WindowsPath('F:/Example 2')
2024-10-07 15:48:41 [info     ] creating db tables
```

yedpodtrzitko:
fixes #535

### DIFF
--- a/tagstudio/src/core/library/alchemy/library.py
+++ b/tagstudio/src/core/library/alchemy/library.py
@@ -1,6 +1,5 @@
 import re
 import shutil
-import sys
 import unicodedata
 from dataclasses import dataclass
 from datetime import UTC, datetime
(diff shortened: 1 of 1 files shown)


CyanVoxel:
Thanks for the quick fix! Seems to work just fine on my end now

PR #537, opened 2024-10-09
Add Location Fields

Nginearing:
- Add location (text line) fields


seakrueger:
Yes, per https://discord.com/channels/1229183630228848661/1232037962116759603/1282597048692117524
Unless this qualifies as a "teeny tiny QoL feature", but IMO while small doesn't qualify and I don't know if another 9.4 release is planned

seakrueger:
The json library is no longer used on main, new fields should be added under library/alchemy/fields.py

Nginearing:
> The json library is no longer used on main, new fields should be added under library/alchemy/fields.py

Of course, I just clicked through the folders, I didn't even look at the folder names, oops.

Nginearing:
I think that's correct now. @seakrueger

CyanVoxel:
This is indeed the correct branch for new features at the moment.

While location fields would be beneficial (as we chatted about a few weeks ago) I think the field would make more sense to hold actual EXIF-style GPS location data rather than just be a cosmetic text field. Custom names for fields are coming in the future too, so there wouldn't be a need for hardcoded names for fields such as this when that eventually lands.

Nginearing:
> This is indeed the correct branch for new features at the moment.
> 
> While location fields would be beneficial (as we chatted about a few weeks ago) I think the field would make more sense to hold actual EXIF-style GPS location data rather than just be a cosmetic text field. Custom names for fields are coming in the future too, so there wouldn't be a need for hardcoded names for fields such as this when that eventually lands.

How about: TagStudio shows location (gps) metadata alongside the usual Date Created and Date Modified metadata and then there is the location field, so that if you can't remember where you took a photo, you could search up the gps coordinates shown in tagstudio and then put in the location name into the location field. What do you think about this?

I feel like GPS coordinates on their own are quite unhelpful but alongside location fields both location fields and GPS coordinates make a lot more sense.

CyanVoxel:
> How about: TagStudio shows location (gps) metadata alongside the usual Date Created and Date Modified metadata, what you're describing, sounds similar to the ability to modify metadata which afaik has been mentioned in an issue before.

This is something I would like also like, along with showing any other EXIF data in a similar manner. This PR/feature request would more so be for adding custom GPS/location data for your own purposes and/or having multiple GPS values per file entry. As I think about it though, I'm struggling to come up with practical use cases for this... If @Nginearing is okay with just having the EXIF data display with the potential for EXIF editing further down the line, then I'm okay with that too.

Nginearing:
I'll get started on programming the implementation to show the gps coordinates exif data.

PR #539, opened 2024-10-10
feat: add ePub thumbnail support (port #387)

CyanVoxel:
*This PR is a port of #387 which previously targeted `thumbnails`/`Alpha-v9.4` to `main` (aka v9.5x).*

This adds thumbnail/preview support for ePub files.


tagstudio/tests/qt/test_thumb_renderer.py
```diff
@@ -0,0 +1,19 @@
...
+
+def test_epub_to_png():
+    # sample.epub, aka "Garrish, Matt. Epub 3 Best Practices (Excerpt). O’Reilly, 2013."
+    # provided by https://idpf.github.io/epub3-samples/30/samples.html
+    # under the CC-BY-SA 3.0 license: https://creativecommons.org/licenses/by-sa/3.0/
+    file_path = Path(__file__).parents[1] / "fixtures" / "sample.epub"
+    tr = ThumbRenderer()
+    cover_image: Image.Image = tr._epub_cover(file_path)
+
+    assert cover_image is not None
```
yedpodtrzitko, line 19: I'd suggest to make a test which will actually check the final result as well similarly as done here for pdf:

https://github.com/TagStudioDev/TagStudio/commit/b1bae1f28aa4be6f54896cf0943dcea0d2b25bcd

CyanVoxel:
Implemented the improved test 👍 I'll do the same for the other PRs as well.

Also we're having unrelated issues with the pytest workflow, so these tests are going to fail here regardless until that's resolved

yedpodtrzitko:
nit: the `.epub` file has almost 4MB, that feels like an overkill a bit. `.epub` is just .zip file, so maybe its content can be gutted a bit? The embedded fonts are not really necessary, and the image itself can be shrinked a bit too.

PR #543, opened 2024-10-11
feat: add pdf thumbnail support (port #378)

CyanVoxel:
This PR is a port of #378 which previously targeted thumbnails/Alpha-v9.4 to main (aka v9.5x).

This adds thumbnail/preview support for PDF files.


tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -786,6 +798,52 @@ def _model_stl_thumb(self, filepath: Path, size: int) -> Image.Image:
 
         return im
 
+    def _pdf_thumb(self, filepath: Path, size: int) -> Image.Image:
+        """Render a thumbnail for a PDF file.
+
+        filepath (Path): The path of the file.
+            size (tuple[int,int]): The size of the icon.
```
yedpodtrzitko, line 805: this type doesnt match the parameter (`int`)

tagstudio/src/qt/widgets/thumb_renderer.py
```diff
@@ -786,6 +798,52 @@ def _model_stl_thumb(self, filepath: Path, size: int) -> Image.Image:
 
         return im
 
+    def _pdf_thumb(self, filepath: Path, size: int) -> Image.Image:
```
yedpodtrzitko, line 801: it would be nice to have a test for this. Rather than explaining what and how, I made the test, so feel free to cherry-pick/merge it into the branch. But it looks like the final png isnt really transparent 🤔 

https://github.com/TagStudioDev/TagStudio/commit/b1bae1f28aa4be6f54896cf0943dcea0d2b25bcd

yedpodtrzitko (changes requested):

CyanVoxel:
I noticed that on my machine (Windows 10) I had to regenerate the png preview to get the test to pass locally  - it seems that for whatever reason, the text is aliased slightly differently when generating on my machine, resulting in a few different pixels that cause the test to fail.

We may need to go with a different line of testing for this, or introduce a margin of error. I'd like to get your thoughts on this.

yedpodtrzitko:
> I noticed that on my machine (Windows 10) I had to regenerate the png preview to get the test to pass locally - it seems that for whatever reason, the text is aliased slightly differently when generating on my machine, resulting in a few different pixels that cause the test to fail.
> 
> We may need to go with a different line of testing for this, or introduce a margin of error. I'd like to get your thoughts on this.

perhaps generating the image in a higher resolution (parameter `size`) could eliminate the discrepancies caused by different antialiasing. Or maybe having a PDF containing just a black square (so no antialiasing) could do the trick.

CyanVoxel:
> > I noticed that on my machine (Windows 10) I had to regenerate the png preview to get the test to pass locally - it seems that for whatever reason, the text is aliased slightly differently when generating on my machine, resulting in a few different pixels that cause the test to fail.
> > We may need to go with a different line of testing for this, or introduce a margin of error. I'd like to get your thoughts on this.
> 
> perhaps generating the image in a higher resolution (parameter `size`) could eliminate the discrepancies caused by different antialiasing. Or maybe having a PDF containing just a black square or some other simple shape (so no antialiasing) could do the trick.
> 
> Another option could be to make the values for `render_options.setRenderFlags()` somehow dynamic, so in tests there would be no antialiasing, thus no discrepancies.

Unfortunately it's not *anti*-aliasing causing the discrepancies, but rather the way the *aliased* text itself is rendered. The downscaling/anti-aliasing step comes later down the line.

Maybe the *font* is different from platform to platform, causing the discrepancies? That would at least be easier to work around than the idea of PIL just having slightly different rendering techniques between each platform. If it's indeed an issue with fonts specifically, then the black square approach shouldn't raise issues - same goes for the SVG previews. 

Top: Windows, Second: macOS (presumably), Middle: Difference, Bottom: Overlay
image

CyanVoxel:
Changed the test to a black square, seems to be working both on my Windows machine and inside the workflow

yedpodtrzitko (approved):

PR #366, opened 2024-08-22
Add support for mscz thubmnails

Joshua-Beatty:
Adds support for mscz thumbnails. Closes https://github.com/TagStudioDev/TagStudio/issues/363
image
Some of the older mscz I found don't have a thumbnail file so I default to the regular image.

This zip file has some example mscz files, 2 with thumbnails, one without.
[MSCZ Examples.zip](https://github.com/user-attachments/files/16715306/MSCZ.Examples.zip)

PR #540, opened 2024-10-11
feat: add svg thumbnail support (port #442)

CyanVoxel:
*This PR is a port of #442 which previously targeted thumbnails/Alpha-v9.4 to main (aka v9.5x).*

This adds thumbnail/preview support for SVG files.

PR #545, opened 2024-10-13
feat: add OpenDocument thumbnail support (port #366)

### LINKED ISSUE #379: [Feature Request]: Add general thumbnail support for file types with thumbnails in zip file (i.e. Libre-/OpenOffice)
### Checklist

### Description

LibreOffice files only show up as icons 
image

### Solution

As described in #363 the thumbnails are part of the file. The file is stored as a zip and the thumbnail is stored under "Thumbnails/thumbnail.png". 

It would be quite easy to do the same as #366 for these files. However, as there are probably more files like these it might be useful to bundle the functionality for file types with this structure.

I don't know how easy it is right now or if it is more useful to wait until [thumbnails](/TagStudioDev/TagStudio/tree/thumbnails) is merged into [main](/TagStudioDev/TagStudio/tree/main).

Some file types which have this kind of pattern are (some where added afterwards):
- [x] .mscz (MuseScore)
- [ ] .odf .odp .ods .odt .odg (LibreOffice/OpenOffice)
- [ ] .ora (OpenRaster [Qronikarz](https://github.com/TagStudioDev/TagStudio/issues/379#issuecomment-2307582992))
- [ ] .kra (Krita [blipdrifter](https://github.com/TagStudioDev/TagStudio/issues/379#issuecomment-2336399524))

### Alternatives

Alternatively, one could handle all of these files individually which would create an unnecessary amount of repetition in the code.

CyanVoxel:
*This PR is a port of #366 which previously targeted thumbnails/Alpha-v9.4 to main (aka v9.5x).*

This adds thumbnail/preview support for OpenDocument file types. This uses the code from #366 and allows any other OpenDocument file type to be rendered with the same method as MuseScore files. 

Closes #379.


yedpodtrzitko (approved):

tagstudio/tests/qt/test_thumb_renderer.py
```diff
@@ -0,0 +1,32 @@
...
+    img_bytes = io.BytesIO()
+    img.save(img_bytes, format="PNG")
+    img_bytes.seek(0)
+    assert img_bytes.read() == snapshot(extension_class=PNGImageSnapshotExtension)
+
+
+def test_ods_preview(cwd, snapshot):
+    file_path: Path = cwd / "fixtures" / "sample.ods"
+    renderer = ThumbRenderer()
+    img: Image.Image = renderer._open_doc_thumb(file_path)
```
yedpodtrzitko, line 27: these two tests (and the tests in #543 #540 #539 ) can be eventually DRY-ed by using `pytest.mark.parametrize`. It executes the test with values in the supplied list, which works great in case the only thing which changes is the fixture name (and thumbnailing method in case of the other PRs). So it can look like this:


```suggestion
@pytest.mark.parametrize(["fixture_file", "thumbnailer"], [
    ("sample.odt", ThumbRenderer._open_doc_thumb), # _open_doc_thumb needs to be changed to @classmethod
    ("sample.ods", ThumbRenderer._open_doc_thumb), # _open_doc_thumb needs to be changed to @classmethod
])
def test_document_preview(cwd, fixture_file, thumbnailer, snapshot):
    file_path: Path = cwd / "fixtures" / fixture_file
    img: Image.Image = thumbnailer(file_path)
```

so instead of two almost-identical tests, there will be just one, and the fixture file + thumbnailing method is passed inside via the parameters.

CyanVoxel: > these two tests (and the tests in #543 #540 #539 ) can be eventually DRY-ed by using `pytest.mark.parametrize`. It executes the test with values in the supplied list, which works great in case the only thing which changes is the fixture name (and thumbnailing method in case of the other PRs). So it can look like this:
> 
> so instead of two almost-identical tests, there will be just one, and the fixture file + thumbnailing method is passed inside via the parameters.

I've tried merging the tests as shown here, including marking `open_doc_thumb` as a `@classmethod`, however it seems that the snapshots can not be found causing the tests to fail. Here's how I have the test set up:

```py
@pytest.mark.parametrize(
    ["fixture_file", "thumbnailer"],
    [
        (
            "sample.odt",
            ThumbRenderer._open_doc_thumb,
        ),
        (
            "sample.ods",
            ThumbRenderer._open_doc_thumb,
        ),
    ],
)
def test_document_preview(cwd, fixture_file, thumbnailer, snapshot):
    file_path: Path = cwd / "fixtures" / fixture_file
    img: Image.Image = thumbnailer(file_path)

    img_bytes = io.BytesIO()
    img.save(img_bytes, format="PNG")
    img_bytes.seek(0)
    assert img_bytes.read() == snapshot(extension_class=PNGImageSnapshotExtension)
```
And the error from each run (same output as if the original tests had missing snapshots):
```
E       AssertionError: assert [+ received] == [- snapshot]
E         Snapshot 'test_document_preview[sample.odt-_open_doc_thumb]' does not exist!
E         + b'\x89PNG\r\n\x1
```

yedpodtrzitko: Since the test was renamed, it's looking for a fixture matching the new test name + parameters, which doesnt exist yet. You have to run pytest with `--snapshot-update` the first time when a new test is created to generate the fixtures.

CyanVoxel:
I've addressed the merge conflicts by combining some of the new and old tests using the `@pytest.mark.parametrize` decorator. I'm not aware of a good way to pass optional parameters using this method (for extra arguments such as `size`), so this results in two sort-of-similar tests. Let me know if there's any way to combine these further, or if this looks good to you.

Thank you for your review so far, and your help with the tests 👍

yedpodtrzitko:
> I've addressed the merge conflicts by combining some of the new and old tests using the `@pytest.mark.parametrize` decorator. I'm not aware of a good way to pass optional parameters using this method (for extra arguments such as `size`), so this results in two sort-of-similar tests. Let me know if there's any way to combine these further, or if this looks good to you.

you can use `functools.partial` to pass a parameter into the function upfront:

```
from functools import partial


def foo(template, size=1000):
    print(template, size)


foo_200 = partial(foo, size=200)
foo_200('xxx')
# will print `xxx, 200`
```

so it will look like this:

```python
from functools import partial

@pytest.mark.parametrize(
    ["fixture_file", "thumbnailer"],
    [
        (
            "sample.pdf",
            partial(ThumbRenderer._pdf_thumb, size=200),
        ),
```

yedpodtrzitko (approved):

yedpodtrzitko:
> I'm not aware of a good way to pass optional parameters using this method (for extra arguments such as `size`)

The question for future consideration is if all the thumbnailing methods shouldnt accept the same parameters (ie. `size`)...

Issue #550, opened 2024-10-19
[Bug]: Image dimensions do not show in Preview Panel


CyanVoxel:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Alpha 9.5.0 (main)

### Operating System & Version

Windows 10

### Description

The image dimensions are currently missing from the preview panel:
![image](https://github.com/user-attachments/assets/80ec0ed1-08fb-4a40-bbfc-d8bbe921b4aa)

### Expected Behavior

The image height and width dimensions in pixels should be displayed under the file type and file size labels. This currently still works for video files. Example running TagStudio Alpha v9.4:
![image](https://github.com/user-attachments/assets/38799b85-3dc0-4a7a-a77a-0c2c51d4a7b3)

### Steps to Reproduce

Click on an image in your TagStudio library to preview it, then observe the missing image dimensions.

Roc25:
Found where problem. Error in if statement
![image](https://github.com/user-attachments/assets/c9608791-27d3-4f53-a18c-77d1b96c91cf)

PR #551, opened 2024-10-19
fix resolution info #550

Roc25:
Fix for #550

### DIFF
--- a/tagstudio/src/core/media_types.py
+++ b/tagstudio/src/core/media_types.py
@@ -392,6 +392,11 @@ class MediaCategories:
         extensions=_IMAGE_VECTOR_SET,
         is_iana=False,
     )
+    IMAGE_RASTER_TYPES: MediaCategory = MediaCategory(
+        media_type=MediaType.IMAGE,
+        extensions=_IMAGE_SET,
+        is_iana=False,
+    )
     IMAGE_TYPES: MediaCategory = MediaCategory(
         media_type=MediaType.IMAGE,
         extensions=_IMAGE_SET | _IMAGE_RAW_SET | _IMAGE_VECTOR_SET,
@@ -531,11 +536,3 @@ class MediaCategories:
             if mime_type and mime_type.startswith(media_cat.media_type.value):
                 return True
         return False
-
-    @staticmethod
-    def is_image_ext_raster(ext: str) -> bool:
-        return (
-            MediaCategories.is_ext_in_category(ext, MediaCategories.IMAGE_TYPES)
-            and not MediaCategories.is_ext_in_category(ext, MediaCategories.IMAGE_RAW_TYPES)
-            and not MediaCategories.is_ext_in_category(ext, MediaCategories.IMAGE_VECTOR_TYPES)
-        )
--- a/tagstudio/src/core/media_types.py
+++ b/tagstudio/src/core/media_types.py
@@ -207,7 +207,7 @@ class MediaCategories:
         ".rw2",
     }
     _IMAGE_VECTOR_SET: set[str] = {".svg"}
-    _IMAGE_SET: set[str] = {
+    _IMAGE_RASTER_SET: set[str] = {
         ".apng",
         ".avif",
         ".bmp",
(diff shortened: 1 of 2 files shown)


Roc25:
I want to add a check to see if the extension is raster in MediaCategories, but Ruff doesn't allow it. If I make it inline the string length will be 250. Or it's bad idea to move it to MediaCategories? What i need to do ? Help please

```
(.venv) ➜  TagStudio git:(resolution-info) ✗ git commit -m "some refactor"
ruff-format..............................................................Passed
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1

tagstudio/src/core/media_types.py:537:9: SIM103 Return the negated condition directly
    |
535 |       @staticmethod
536 |       def is_image_ext_raster(ext: str) -> bool:
537 |           if (
    |  _________^
538 | |             not MediaCategories.is_ext_in_category(ext, MediaCategories.IMAGE_TYPES)
539 | |             or MediaCategories.is_ext_in_category(ext, MediaCategories.IMAGE_RAW_TYPES)
540 | |             or MediaCategories.is_ext_in_category(ext, MediaCategories.IMAGE_VECTOR_TYPES)
541 | |         ):
542 | |             return False
543 | |         else:
544 | |             return True
    | |_______________________^ SIM103
    |
    = help: Inline condition

Found 1 error.
No fixes available (1 hidden fix can be enabled with the `--unsafe-fixes` option).
```

CyanVoxel:
Thank for the fix! I think it would be more worth it to separate out raster images into a new `IMAGE_RASTER_TYPES` object that helps compose `IMAGE_TYPES` similar to how the vector and raw images work, that way there's no need for the `is_image_ext_raster()` function.

Roc25:
You're right it will be better. But i'm not sure must i set is_iana in IMAGE_RASTER_TYPES (Its false at this moment)

CyanVoxel:
> You're right it will be better. But i'm not sure must i set `is_iana` in `IMAGE_RASTER_TYPES` (Its false at this moment)

I (roughly) go by the [IANA categories](https://www.iana.org/assignments/media-types/media-types.xhtml#image) for this, which seems to include things like SVGs under their "image" category - so I would have `IMAGE_TYPES` set to true as that should include the union of all other image types, while `IMAGE_RASTER_TYPES` would be set to false. This should prevent situations where leveraging the mimetype for something like an svg might result in it thinking that it's a *raster* image specifically, while keeping the ability for a previously unknown image type to attempt to be rendered through normal means.

tagstudio/src/core/media_types.py
```diff
@@ -392,6 +392,11 @@ class MediaCategories:
...
         is_iana=False,
     )
+    IMAGE_RASTER_TYPES: MediaCategory = MediaCategory(
+        media_type=MediaType.IMAGE,
+        extensions=_IMAGE_SET,
+        is_iana=False,
+    )
     IMAGE_TYPES: MediaCategory = MediaCategory(
         media_type=MediaType.IMAGE,
         extensions=_IMAGE_SET | _IMAGE_RAW_SET | _IMAGE_VECTOR_SET,
```
CyanVoxel, line 402: The `IMAGE_TYPES` media category should now include the new _IMAGE_RASTER_SET
```suggestion
        extensions=_IMAGE_SET | _IMAGE_RASTER_SET | _IMAGE_RAW_SET | _IMAGE_VECTOR_SET,
```

tagstudio/src/core/media_types.py
```diff
@@ -392,6 +392,11 @@ class MediaCategories:
         extensions=_IMAGE_VECTOR_SET,
         is_iana=False,
     )
+    IMAGE_RASTER_TYPES: MediaCategory = MediaCategory(
+        media_type=MediaType.IMAGE,
+        extensions=_IMAGE_SET,
```
CyanVoxel, line 397: If `IMAGE_RASTER_TYPES` is going to pull from the same `IMAGE_SET`, then that set should probably be renamed to `_IMAGE_RASTER_SET` to avoid future confusion

CyanVoxel (approved):

Issue #555, opened 2024-10-20
[Feature Request]: change images on pressing the mouse instead of on releasing


BPplays:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

when i press my mouse down when selecting an image it doesn't select until i release, and holding doesn't seem to serve a function, so it adds an extra delay for no reason.

Roc25:
I have implemented this functionality, but I think it is better to move this into separate settings, since not all users may like it. You can test it [here](https://github.com/Roc25/TagStudio/tree/press-select-instead-release)

CyanVoxel:
A click-and-drag and/or box select function would give this more purpose, but I have no issue with this idea as it stands

PR #556, opened 2024-10-20
Change item select to press instead of release (click)

Roc25:
For #555. I think it's better to add this to the settings, but leave it as default

### DIFF
--- a/tagstudio/src/qt/widgets/item_thumb.py
+++ b/tagstudio/src/qt/widgets/item_thumb.py
@@ -381,9 +381,9 @@ class ItemThumb(FlowWidget):
     def update_clickable(self, clickable: typing.Callable):
         """Updates attributes of a thumbnail element."""
         if self.thumb_button.is_connected:
-            self.thumb_button.clicked.disconnect()
+            self.thumb_button.pressed.disconnect()
         if clickable:
-            self.thumb_button.clicked.connect(clickable)
+            self.thumb_button.pressed.connect(clickable)
             self.thumb_button.is_connected = True
 
     def refresh_badge(self, entry: Entry | None = None):


BPplays:
@Roc25 seems to work fine in my testing.
⁤⁤
> I think it's better to add this to the settings

i couldn't find any global settings menu, though there seems to be an ini file in `%appdata%\TagStudio` on windows.
i think some discussion on what the settings menu should look like probably needs to happen first,
@CyanVoxel are there any plans for how the options menu should look?

Roc25:
I quickly wrote a simple setting, if it’s approved I’ll refactor it later. It's a checkbox in "window" dropdown menu on top (idk what it's called). At this moment it has some bugs

BPplays:
@Roc25 even though it stays selected after a restart it needs to be turned off and on again to actually work.
also i feel like it should select on press by default and release should be opt-in.

CyanVoxel:
> i couldn't find any global settings menu, though there seems to be an ini file in `%appdata%\TagStudio` on windows. i think some discussion on what the settings menu should look like probably needs to happen first, @CyanVoxel are there any plans for how the options menu should look?

@VasigaranAndAngel was last working on a settings menu implementation - this was hopefully coming in #515 however due to the urgency of such a feature I think it might be best to go forward with one ourselves.

The .ini/QSettings file is intended for application-wide settings, while library-specific preferences should be stored separately. There was a sizable [thread in the discord](https://discord.com/channels/1229183630228848661/1279024877113708576) about potential settings implementations, but mostly from a backend perspective. For the UI I would image things just read from the TOML-formatted .ini/.plist file into a tab-organized window. I'll draft up a feature request for this to better organize this and facilitate further discussion and feedback

seakrueger:
Personally, I don't feel this feature needs a settings option. It's a small implementation detail that most users won't notice or see a need to change the behavior of.  It also introduces unnecessary code duplication checking whether .press or .click should be called.

CyanVoxel:
After testing I think this should just be the default and only behavior. Windows Explorer, Finder, and at least some Linux file explorer(s) (Dolphin confirmed) all have "select on mouse down" behavior.

Roc25:
Squashed several commits

PR #560, opened 2024-10-24
[fix] bug where preview panel ui would not update when fields where removed from an entry

DandyDev01:
When removing fields from an entry the database would properly update, but the ui would not reflect those changes until a different entry was selected then switched back to the entry that was edited. This PR fixes that.


yedpodtrzitko (approved):

CyanVoxel:
Thank you for the fix!

Issue #561, opened 2024-10-25
[Feature Request]: Add Alembiс for db migrations


Roc25:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

Add Alembic for migrations. For example, if I want to add the ability to pin tags, I must create a column in the database for the tags if they are pinned, and the database should automatically migrate (#542).

### Solution

Add Alembic

PR #562, opened 2024-10-25
Add Alembiс for db migrations

Roc25:
For #561

Migrations should be created and applied automatically when user opens library


Roc25:
Currently, every time the library is opened, a migration file is created

yedpodtrzitko:
see this commit addressing some issues mentioned
https://github.com/yedpodtrzitko/TagStudio/commit/17e2df1927be13e32078dbdf6222dc1feb99599b

- `migrations/versions/*` shouldnt be in `.gitignore` because we actually need to keep the generated migrations inside repository

- there's a problem with `alembic.ini` location. it's outside the `tagstudio` folder, which assumes the alembic `import` statements should start with `tagstudio.` Howver in `env.py`, you use `import` starting with `src.` (subfolder of `tagstudio`), which implies that alembic is executed from within the `tagstudio` folder... but that doesnt actually match the content of `alembic.ini`. So either the imports need to change, or the content (and ideally location) of `alembic.ini` needs to be inside `tagstudio`. In the commit above I changed content of `alembic.ini`.

- to generate the migrations, there needs to be some database to compare its content with the models to generate the difference for migrations. I added an empty database file into the `alembic` folder and assigned its path to `sqlachemy.url` in `alembic.ini`
- after that it's possible to generate the initial migration via:


```
# inside `tagstudio` folder
alembic -c ../alembic.ini revision --autogenerate -m "initial migration"
```

Roc25:
If I understand you correctly, we can also not change imports and not move `alembic.ini`, but we can add `/tagstudio` to `sys.path`

In `alembic.ini`
`prepend_sys_path = . tagstudio`

In that way we can use alembic in root dir and imports will be ok. If I missed something let me know

yedpodtrzitko:
it doesnt work for me. When I try to run TS from your branch and open a library, I end up with this error:
```
  File "/Users/yed/dev/python/TagStudio/tagstudio/src/qt/widgets/preview_panel.py", line 374, in <lambda>
    return lambda: self.driver.open_library(Path(path))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yed/dev/python/TagStudio/tagstudio/src/qt/ts_qt.py", line 1118, in open_library
    open_status = self.lib.open_library(path)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yed/dev/python/TagStudio/tagstudio/src/core/library/alchemy/library.py", line 170, in open_library
    self.migrate(connection_string.render_as_string(hide_password=False))
  File "/Users/yed/dev/python/TagStudio/tagstudio/src/core/library/alchemy/library.py", line 147, in migrate
    command.stamp(alembic_cfg, "head")
  File "/Users/yed/.pyenv/versions/312-tags/lib/python3.12/site-packages/alembic/command.py", line 665, in stamp
    script = ScriptDirectory.from_config(config)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yed/.pyenv/versions/312-tags/lib/python3.12/site-packages/alembic/script/base.py", line 170, in from_config
    raise util.CommandError(
alembic.util.exc.CommandError: No 'script_location' key found in configuration.
```

Roc25:
Fixed it .bat and .sh files. We must run the application from `tagstudio`, not from the root directory, otherwise the built program (.exe) will give an error

yedpodtrzitko:
> We must run the application from `tagstudio`, not from the root director


That sounds like an arbitrary decision with no benefit. The path to `alembic.ini` can be location-agnostic (`alembic_cfg = Config(str(Path(__file__).parent.parent / "alembic/alembic.ini"))`), and so can be `script_location` in `alembic.ini` (`script_location = src:core/library/alembic/migrations
`), so it doesnt matter where TS is executed from.

PR #563, opened 2024-10-26
feat: allow multiple directories per library

yedpodtrzitko:
- this PR allows to have multiple individual directories in the same library. This means there's no such thing as "root" of the library, so there's a new dialog where metadata of the library should be stored

- this also means that "open library" and "create library" are now two different actions as "open library" doesnt implicitly create the library subfolder (ie `.TagStudio`) automatically in given folder

- each directory can be then toggled to show/hide its content in the library

Screenshot 2024-10-26 at 14 26 39

Screenshot 2024-10-26 at 14 27 00

### DIFF
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -25,7 +25,7 @@ jobs:
         run: |
           python -m pip install --upgrade uv
           uv pip install --system -r requirements.txt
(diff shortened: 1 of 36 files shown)


yedpodtrzitko:
this reiles on #516, so I will close this PR for now.

PR #564, opened 2024-10-29
fix: Corrected link to lead to the roadmap.

Hissymaster:
Correcting the link from Planned features to Roadmap.


CyanVoxel:
Ah thank you for the fix! I knew I was probably missing a reference somewhere...

PR #575, opened 2024-11-05
feat: add filetype and mediatype searches

### LINKED ISSUE #574: [Feature Request]: File/Media Type Search
### Checklist

### Description

_this issue is from the tagstud.io roadmap_

Being able to search for files by file type (i.e. extension, like jpg, png, etc) and media type (image, plaintext, etc) would be helpful, especially for untagged files.

### Solution

Two new search methods:
`filetype:` - a search method that searches by each entry's suffix (this functionality may be able to be done by using `path:`)
`mediatype:` - a search method that searches by extensions included in each MediaCategory's `extensions` property

For example, searching `filetype:jpg` would only show files with a suffix of `jpg` (NOT `jpeg`, though that could be added), and searching by `mediatype:image` would show all files with an extension that is included in `_IMAGE_SET`, `_IMAGE_RAW_SET`, and `_IMAGE_VECTOR_SET`. 

### Alternatives

_No response_

python357-1:
fixes #574 

added search types `filetype:` and `mediatype:`

example library:
image

filetype search example:
image

mediatype search example:
image


python357-1:
i did try and add some tests based on what i was seeing from other tests, but i couldn't really tell how a library gets passed to tests so if they aren't correct, feel free to fix/point me to some information on how to fix them

tagstudio/src/core/library/alchemy/library.py
```diff
@@ -438,6 +439,18 @@ def search_library(
                 )
             elif search.path:
                 statement = statement.where(Entry.path.ilike(f"%{search.path}%"))
+            elif search.filetype:
+                statement = statement.where(Entry.suffix.ilike(f"%{search.filetype}%"))
```
CyanVoxel, line 443: This introduces a quirk where filetypes containing the name of others are returned along with the original search. As an example, I had `.apng` files returned alongside `.png` files when searching for `filetype:png`

CyanVoxel (changes requested):
Looks great, thank you for this! Approved when the comment is addressed *(might be a good addition to the test cases, too)*

tagstudio/src/core/library/alchemy/library.py
python357-1: fixed that and added a test that should catch that if it ever happens in the future

CyanVoxel (approved):

Issue #577, opened 2024-11-06
[Feature Request]: Filename Search


python357-1:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

_this is from the tagstud.io roadmap_
add a new search filter, `filename:`, that is used to show only entries with a given filename, ex: `filename:skeleton` would only show entries with "skeleton" in the filename

this could potentially also use positional searches, like regex's ^ and $, for better searching:

`filename:^skeleton` would only return filenames that start with skeleton, and `filename:skeleton$` would only return filenames that end with skeleton.

### Alternatives

this could also be implemented in the `path:` filter by using globs, but it would probably be more complicated to implement:

`path:*skeleton*.*` would return all files whose filename includes skeleton
`path:skeleton*.*` would return all files whose filename starts with skeleton
`path:*skeleton.*` would return all files whose filename ends with skeleton

CyanVoxel:
Globs seem the way to go 👍

CyanVoxel:
Completed in #575 & #582

PR #582, opened 2024-11-11
feat: make path search use globs

python357-1:
Implements the alternative searching mentioned in #577


CyanVoxel (approved):
Looks good to me! The only comment I have would be to add support for quotes, but that's included with #575 so it's effectively taken care of. Thank you for this!

PR #583, opened 2024-11-12
remove queued thumnail jobs when closing library

yedpodtrzitko:
No need to keep rendering the queued thumbnails when closing library and/or quitting the app


CyanVoxel (approved):
Seems to work as intended to me, thanks for the fix!

PR #590, opened 2024-11-17
feat: Shortcut manager

VasigaranAndAngel:
For customize key bindings.

this will be easily accessible once the [settings window and key bindings](https://github.com/orgs/TagStudioDev/projects/1?pane=issue&itemId=86303015) in it are implemented. until then, we'll need to manually edit the `config.ini` file to set key bindings.

To test:
Add a new group and values as shown below to your config.ini file:
```
[Shortcuts]
open=Ctrl+1
save=Ctrl+2, Ctrl+3, Ctrl+4
new_tag="Ctrl+,", Ctrl+\"
deselect=
```
- `open=Ctrl+1`: Sets the shortcut for "open" to `Ctrl+1`
- `save=Ctrl+2, Ctrl+3, Ctrl+4`: You can add multiple shortcuts by separating with `,`. All of these shortcuts will trigger the action.
- `new_tag="Ctrl+,", Ctrl+\"`: Use `"` to specify `,` as a key. use `\` before `"` to specify `"` as a key (`QSettings` supported format)
- `deselect=`: An empty value means no shortcut keys will trigger the action.

### DIFF
--- a/tagstudio/src/qt/helpers/ini_helpers.py
+++ b/tagstudio/src/qt/helpers/ini_helpers.py
@@ -1,0 +1,64 @@
+from string import ascii_letters, digits
+
+import structlog
+
+logger = structlog.get_logger("ini_helpers")
+
+
+def is_valid_ini_key(key: str) -> bool:
+    """Check if a given string is a valid INI key for QSettings.
+
+    This function validates whether the provided key is suitable for use in an INI file
+    managed by QSettings. Valid INI keys are those that are human-readable and do not
+    require URL encoding when saved.
+
+    A valid key can contain letters (both uppercase and lowercase), digits,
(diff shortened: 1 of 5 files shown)


Computerdores:
Hmm, this doesn't work for me on Windows 10.

Also I noticed that the Shortcuts Group gets removed from the config after starting and doesn't get readded on closing

CyanVoxel:
Could this be updated now that #622 is merged to make is easier to review?

CyanVoxel:
Likewise with #587 I'm sorry that this PR ended up getting left behind, having come before the settings system was implemented and eventually becoming incompatible. While this is a great feature to have I'm unfortunately going to have to close this due to age and incompatibility at this time.

PR #592, opened 2024-11-18
feat: Make the create tag panel have empty tag name field

Cool-Game-Dev:
Made tag name empty by default and added placeholder "New Tag" text.


python357-1:
Seems like it works as intended

Computerdores (approved):
Looks good!

Cool-Game-Dev:
I can't get the signals to activate, can anyone take a look?

Cool-Game-Dev:
If anyone could spare debugging time that would be greatly appreciated.

VasigaranAndAngel:
> If anyone could spare debugging time that would be greatly appreciated.

using signals for this isn't the best approach. because we'd need to connect the signal every time `BuildTagPanel` is created. it's being created from tag_box (which show tags in the preview panel and has the signals are connected right now. when you right click a tag in the preview panel to edit it, the save buttons works as expected.), ts_qt (for creating new tags via the menu bar or shortcut) and tag_database (used for managing tags).

`BuildTagPanel` inherits from `PanelWidget` and then it's passed to `PanelModal` as the `widget` parameter. this is how the widgets are constructed right now.

so, instead of using signals, you can access the buttons of `PanelModal` by adding new class variables in `PanelWidget`:
```py
class PanelWidget(QWidget):
    """Used for widgets that go in a modal panel, ex. for editing or searching."""

    done = Signal()
    panel_save_button: QPushButton | None = None
    panel_cancel_button: QPushButton | None = None
    panel_done_button: QPushButton | None = None
```
then assign the buttons to those variables in `PanelModal`
    image
3. now it's easy to access the buttons
    image

i've added variables to cancel and done buttons also. that's not necessary.

Additionally, set the fixed height of the `self.name_field` to 24 to fix the flickering when adding and removing the border.

let me know if i'm not explained correctly.

python357-1:
Yeah @Cool-Game-Dev I did this locally and it works perfectly. Just copy what he has and commit it, and this issue will be done.
@VasigaranAndAngel Thanks so much for the fix!

tagstudio/src/qt/modals/build_tag.py
```diff
@@ -204,6 +208,16 @@ def set_tag(self, tag: Tag):
 
         self.tag = tag
 
+    def on_name_changed(self):
+        is_empty = not self.name_field.text().strip()
+
+        self.name_field.setStyleSheet(
+            "border: 1px solid red; border-radius: 2px" if is_empty else ""
+        )
```
CyanVoxel, line 216: Superficial nitpick: Use the universal "red" UI color defined in `palette.py`
```suggestion
        self.name_field.setStyleSheet(
            f"border: 1px solid {get_ui_color(ColorType.PRIMARY, UiColor.RED)}; border-radius: 2px"
            if is_empty
            else ""
        )
```

tagstudio/src/qt/modals/build_tag.py
```diff
@@ -51,6 +51,8 @@ def __init__(self, library: Library, tag: Tag | None = None):
         self.name_title.setText("Name")
         self.name_layout.addWidget(self.name_title)
         self.name_field = QLineEdit()
+        self.name_field.setFixedHeight(24)
+        self.name_field.textChanged.connect(self.on_name_changed)
```
CyanVoxel, line 55: Suggestion to add some placeholder text to signal to users who don't or [can't notice](https://en.wikipedia.org/wiki/Congenital_red%E2%80%93green_color_blindness) the red outline that a name is required
```suggestion
        self.name_field.setFixedHeight(24)
        self.name_field.textChanged.connect(self.on_name_changed)
        self.name_field.setPlaceholderText("Tag Name (Required)")
```

CyanVoxel (commented):
Looks fantastic! This perfectly balances the benefits of having a default tag name for quick creation and quick lookup with the benefit of being able to immediately start typing a desired tag name. I've just got a couple additional suggestions here, but once those are addressed then this is good to be merged!
Thank you @Cool-Game-Dev and everyone else who's helped!

tagstudio/src/qt/modals/build_tag.py
Cool-Game-Dev: Seems reasonable to me.

Cool-Game-Dev:
Both of those seem great! Thanks everyone for the help!

CyanVoxel:
Thank you all again for your work on this PR!

Issue #593, opened 2024-11-18
[Bug]: link to 'Planned Features' in docs leads to a missing page


jakehurst:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

main

### Operating System & Version

macOS 14.7.1

### Description

On https://github.com/TagStudioDev/TagStudio/blob/main/CONTRIBUTING.md, under 'Getting Started,' there is a link to "Check the _Planned Features_ page" which leads to https://github.com/TagStudioDev/TagStudio/blob/main/docs/updates/planned_features.md.

That page was removed with commit https://github.com/TagStudioDev/TagStudio/commit/237638024507d31d98a054fa46c2e62ab73dff12

### Expected Behavior

Expected behavior is that by clicking or visiting the link, it should lead to a page that shows the planned features or the feature roadmap instead of a broken link.

The likely fix is to replace the link to instead point to the 'Roadmap' page, added with commit https://github.com/TagStudioDev/TagStudio/commit/04149f6454de3639823be52e0e56438a5a9d63d5

Additionally, search for all other instances of links to 'planned_features.md' and replace those to instead point to 'roadmap.md'.

The Roadmap page title is 'Roadmap'. Should it be 'Feature Roadmap'?  We could then make the links to the Feature Roadmap page consistent throughout the documentation, and encourage new contributors to read that page instead.

### Steps to Reproduce

1. Visit https://github.com/jakehurst/TagStudio/blob/main/CONTRIBUTING.md
2. Scroll to 'Getting Started' section.
3. Click link 'Planned Features'.
4. Page leads to a 404 with this error: "The main branch of TagStudio does not contain the path docs/updates/planned_features.md."

jakehurst:
Noticed that I had a VSCode YAML formatter gone rogue and formatting things that I did not touch. I've disabled that plugin and reverted the lines that I did not want changed.

PR #594, opened 2024-11-18
docs: fix link to the feature roadmap page

jakehurst:
Refs: #593


jakehurst:
I set up verified commits and force-pushed, it's the same diff.

jakehurst:
Noticed that I had a VSCode YAML formatter gone rogue and formatting things that I did not touch. I've disabled that plugin and reverted the lines that I did not want changed.

CyanVoxel (approved):

Issue #325, opened 2024-07-20
[Feature Request]: Improving search


mm12:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

I know this is somewhat on the roadmap, but I thought I would share some specifics of how search should be improved. It is very important that work on search functionality starts early, to make sure a system is developed in a way that supports the future implementation of features (ie, we dont want to be in a situation where implementing a standard search feature would require major changes)

### Solution

End goal: A fully featured search system. This could make use of Elastic and/or Opensearch. Desirable qualities: 
- A Standard search syntax
  - should be similar to Boorus and search engines - no one wants to learn a new syntax
- Grouping of items
  - Many implementations of grouping treat each group as its own search, combining results are it goes up. 
- Boolean operators (AND, OR, NOT) of items and groups
- Reserved characters and names: 
  - there should be reserved characters, such as quotation marks and colons, that mean something in search syntax. To search the character itself, it will need to be escaped.
  - Any assignable field should be searchable. For example with `filename:<query>` - this should able to be used with any given type of field. Of course, by default, it is assumed to be a tag. 
- Searching text fields: 
  - RegEx seems like the easiest way to do this. Not everyone knows it, but it could be a mode. 
- Handling entries with spaces (#112) - there are 2 ways this is usually dealt with: 
  - Make all spaces *actually* underscores: if your tag is `a happy new year` it becomes `a_happy_new_year`, but the underscores do not get shown to the user. This means that spaces and underscores are effectively the same. 
  - grouping using quotations: `"a happy new year"` - however, this means that items with a quote in them will need to be escaped: a tag named `"wow thats cool"` would need to be searched with something like `"\"wow thats cool\""` instead.

CyanVoxel:
I agree with your suggestions on this, however there's already several open issues and PRs pertaining to these features.
Boolean search is currently being tracked via #225 + #314 and is being implemented in #284/#310 with metadata search being tracked via #272 and being implemented in #284. I believe that grouping, escaping characters, and addressing tags with spaces are all coming with one or more of these PRs as well.

KillyMXI:
Better written issue comes late though, as it often happens...

There were some discussions in Discord as well. Maybe some of that needs to be summarized in GitHub somehow.

[...]

Issue #314, opened 2024-07-11
[Feature Request]: Add exclusive And/Or search options


coolesding:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

Currently, the search can only be set to either **And** (Includes all Tags) and **Or** (Includes any Tag). I just had the problem of trying to find an image that has two tags, but no others, which seemingly I have to do manually searching for all the tags I want and filtering the ones I don't want myself.

### Solution

I think adding the options
*Ex. Or* (Exclusively includes any Tags)
*Ex. And* (Exclusively includes all Tags)
would be great additions to the versatility of Tagstudio!

# EXAMPLE

I have a database of many images from and about the bocchi the rock anime, which among others includes the tags:

Kita
Bocchi
Nijika
Ryou

There are all combinations of images tagged, some with only one, some with multiple, some with all characters tagged.

Searching for "Kita, Bocchi" in Exclusive Or mode would result in all images that only have the kita tag, only have the bocchi tag or only have the bocchi and kita tags, and no others.

Searching for "Nijika, Ryou" in Exclusive And mode would result in all images that only have the Nijika and the Ryou tags, and no others.

### Alternatives

The naming "Ex. Or" and "Ex. And" could probably be improved, but I can't think of a better solution currently (In my defense, i am writing this at 1 in the morning after having a horrible sleep)

KillyMXI:
This is related to #112 
Also useful as a confirmation of the demand for more Boolean operators.

Exclusive OR is commonly written as XOR.
There is no such thing as Exclusive AND, but there are such operators as XNOR, NAND.

Truth tables:

|  A  |  B  | AND | OR  | XOR | XNOR | NAND | NOR |
| --- | --- | --- | --- | --- | ---- | ---- | --- |
| ❌  | ❌ | ❌ |  ❌ |  ❌ |  ✔  |  ✔  |  ✔  |
| ❌  | ✔  | ❌ |  ✔  |  ✔  |  ❌ |  ✔  | ❌  |
|  ✔  | ❌ | ❌ |  ✔  |  ✔  |  ❌ |  ✔  | ❌  |
|  ✔  | ✔  | ✔  |  ✔  |  ❌ |  ✔  |  ❌ | ❌  |

[...]

PR #595, opened 2024-11-19
Begin adding query parsing

python357-1:
For the moment this will be a draft PR to track the status of adding query parsing. When finished it will likely implement most or all of the features in #272, #314, and #325, with the possibility of others being implemented as well.

### DIFF
--- a/tagstudio/src/core/query_tokenizer.py
+++ b/tagstudio/src/core/query_tokenizer.py
@@ -1,0 +1,100 @@
+from enum import Enum
+
+
+class TokenType(Enum):
+    KEYWORD = 1
+    IDENTIFIER = 2
+    ILLEGAL = 4
+    COLON = 5
+    LEFT_PAREN = 6
+    RIGHT_PAREN = 7
+    END_OF_INPUT = 8
+
+
+class Token:
+    def __init__(self, data: str | None, tok_type: TokenType):
+        self.data: str | None = (
+            data  # this doesnt need to have data for token types where it is obvious (i.e. COLON)
+        )
+        self.type: TokenType = tok_type
+
+
+class QueryTokenizer:
+    def __init__(self, query: str):
+        self.query: str = query
+        self.current_pos = 0
+        self.next_pos = 0
+        self.current_character = ""
+        self.next_character = ""
+        self.read_character()
+
+    keywords = ["AND", "OR", "NOT"]
+
+    @staticmethod
+    def is_character(character: str) -> bool:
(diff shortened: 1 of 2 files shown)


Tishj:
This sounds like it could benefit from using a Parsing Expression Grammar (<https://peps.python.org/pep-0617/#background-on-peg-parsers>), as it looks like this PR is kind of implementing a parser/tokenizer.

This could be made much simpler by using an existing library and only defining a grammar based on the wants and needs outlined by the referenced issues

Tishj:
A basic example using `parsimonious` (not sure this is the best but it's one of the options)

```py
from parsimonious.grammar import Grammar

GRAMMAR = r"""
expr        = (entry / emptyline)*
entry       = section pair*

section     = lpar word rpar ws
pair        = key equal value ws?

key         = word+
value       = quoted / unquoted
word        = ~r"[-\w\.]+"  # Allows alphanumeric, hyphens, underscores, and periods
quoted      = ~r'"[^"]*"'   # Matches anything inside double quotes
unquoted    = ~r"[^\n]+"    # Matches everything until the end of the line (allows spaces)
equal       = ws? "=" ws?
lpar        = "["
rpar        = "]"
ws          = ~r"\s*"
emptyline   = ws+
"""

EXAMPLE = """
[General]
app_name = MyApp
version = 1.0.0
author = John Doe

[Database]
host = localhost
port = 5432
user = admin
password = secret

[Features]
enable_feature_x = true
enable_feature_y = fa
"""

def main():
	grammar = Grammar(GRAMMAR)
	res = grammar.parse(EXAMPLE)
	print(res)

if __name__ == '__main__':
	main()
```

(this grammar is a modified version of the example from <https://github.com/erikrose/parsimonious#example-usage>)
the example is generated from asking chatgpt for a basic `.ini` file example

python357-1:
I'm sure using a library like that could be better, but a parser (particularly for the very simple "language" we're planning on it being) is not a hard thing to make, and with that particular library, using a parser generator that readily admits it "may be slow and use lots of RAM" and hasn't been actually worked on for over 2 years isn't a great example of why we should use one.

Tishj:
Sure, that's fair.

As far as I know RAM usage depends on the size of the user input and the grammar, which will be small as you said.
I also think that the required complexity of the language will evolve over time, but perhaps this avenue can be revisited when the complexity outgrows the handrolled parser/tokenizer approach

python357-1:
closed in lieu of #606

PR #596, opened 2024-11-20
feat(parity): backend for aliases and parent tags

DandyDev01:
subtags can be added and removed when creating or editing a Tag
   - parent tags would not be listed before 
   - parent tags could not be removed before

can exclude tags from the tag search panel
   -   it was possible to make a tag parent itself before. Now when selecting a parent tag, the tag being created or edited is excluded from the tag search panel, so it cannot be added as a parent to itself. 
   - before when a tag could parent itself, that tag would be corrupted and users would be unable to edit the tag or add it as a parent to another tag (this would corrupt that tag too)

changed how aliases and parents are managed (they use a flowlayout):
image
NOTE: the ui for aliases is being updated to use a table in https://github.com/TagStudioDev/TagStudio/pull/534 and parents will be reverted to a list.

  
   - aliases can be removed and added
   - aliases can be edited


CyanVoxel:
The backend works seems solid enough to pull! Quick bug I found though, edits to aliases made from the preview panel don't seem to save. Edits made in the tag management window work, and subtag changes work from both panels.

Side note: I like the flow layout for the parent tags!

Issue #602, opened 2024-11-23
[Bug]: Open in explorer does not open the correct folder


KirilBourakov:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Main

### Operating System & Version

Windows 11

### Description

Open in explorer does not open the correct folder, opens desktop instead. 

### Expected Behavior

Tagstudios should open the correct folder

### Steps to Reproduce

1. Open tag studios
2. Right click an image
3. Click 'open in explorer'

KirilBourakov:
This seems to be an issue with posix paths, and the fact that they use / instead of \. Simply not using a posix path when the user is on windows seems to fix this.

PR #603, opened 2024-11-24
Fixed open in explorer does not open the correct folder

KirilBourakov:
Fixed normpath variable using posix, leading to windows not understanding the command given to it (issue #602)


Computerdores (approved):
works for me and looks good

PR #604, opened 2024-11-24
feat(parity): migrate json libraries to sqlite

CyanVoxel:
This PR adds the ability for users to migrate TagStudio **v9.1-v9.4** JSON libraries to the **v9.5** SQLite format.

Attempting to open a library folder with no `ts_library.sqlite` file and a `ts_library.json` file present will initiate the modal prompt for the user to start the migration process. Closing this prompt early stops the migration process and will not open the library. The migration is not completed until the user clicks the "Finish Migration" button, and the `ts_library.json` is never deleted as a backup precaution.

image

### Migration Checklist
- [x] Library Folder
- [x] Entries
	- [x] File Path
	- [x] File Suffix
	- [x] Fields
		- [ ] ~~Field Order~~ *(The SQL DB doesn't seem built for this but it's fine for now)*
		- [x] Text Box Fields
			- [x] Name/Subtype
			- [x] Content
		- [x] Text Line Fields
			- [x] Name/Subtype
			- [x] Content
		- [x] Date Fields
			- [x] Name/Subtype
			- [x] Content
		- [x] Tag Box Fields
			- [x] Name/Subtype
			- [x] Content
- [x] Tags
	- [x] Tag Name
	- [x] Tag Shorthand
	- [x] Tag Aliases
	- [x] Tag Parent Tags
	- [x] Tag Color
- [x] File Extension List
	- [x] List of extension strings
		- [x] Removes leading “.”


python357-1:
Have no complaints with my small library, but we should get some bigger samples.

Cool-Game-Dev:
I ran a test of the code and it appears it put fields on the wrong folders?

Cool-Game-Dev:
Seems to work on my smaller library, but I agree with python on this one.

CyanVoxel:
Updated tests to account for the way the SQL DB loads multiples of the same tag field type on entries. Currently the SQL DB loads the correct number of fields, but merges and mirrors the tag lists across each similar tag field type. This seems to be a limitation of the DB schema.

While I don't have much public documentation on the concept yet, I'm highly considering targeting the "[Tag Categories](https://docs.tagstud.io/library/tag_categories/)" concept for 9.5 instead of 9.6 to not only address this issue, but to also address the currently broken mixed field editing that needs overhauling anyway.

Cool-Game-Dev:
@CyanVoxel I'm not complaining, go for it if you need.

CyanVoxel:
> Pressing "Start and Preview" currently freezes the window until it is done without any feedback (besides the log) that it is still working

This is something I also don't like but unfortunately don't know a way around... I actually ended up creating an entire embedded progress bar that updated the values on screen as it the library processed, but discovered that SQLAlchemy's desire to be on a single thread directly headbutts against Qt's desire for progress bars to be on different threads. I was unable to come up with a solution where I could yield feedback from the library while not ruining the db connection.
 
> After pressing "Finish Migration" the window freezes again and a small, completely white (also frozen) window with the title "Refreshing Directories" shows up. At some point they both vanish and "Running Macros on New Entries" shows up

I've noticed this as well. For me the white window only loads for a split second as it seems to be closed to fast that it never fully loads. There might be somewhere that an `.update()` method can be placed to help with this, but the issue seems to stem from the refresh bar getting created and destroyed too quickly, and/or being created while the main thread is pretty busy.
 
> After first migrating I had quite a lot of input lag (0.5-2s) when selecting an entry or scrolling (this went away after 30s or so)
>       
> something I just noticed: the input lag when selecting an entry is also sometimes present just after opening the library but only for the first time selecting an entry

Hmm, there was a bug from 332 that caused lag like this when selecting entries however it should've been fixed in this PR. Could this be thumbnails rendering in the "background" while you try to click on entries? I usually get some lag until everything on the page is rendered.

Computerdores:
> This is something I also don't like but unfortunately don't know a way around... I actually ended up creating an entire embedded progress bar that updated the values on screen as it the library processed, but discovered that SQLAlchemy's desire to be on a single thread directly headbutts against Qt's desire for progress bars to be on different threads. I was unable to come up with a solution where I could yield feedback from the library while not ruining the db connection.

Yikes that sounds like a pain to get working. Have you thought about instead of having an indicator for progress, just having an indicator that the program is still running (e.g. a spinner)? This way the window wouldn't be frozen and it would still convey that it isn't done yet

> Hmm, there was a bug from 332 that caused lag like this when selecting entries however it should've been fixed in this PR. Could this be thumbnails rendering in the "background" while you try to click on entries? I usually get some lag until everything on the page is rendered.

I was already seeing all the thumbnails at that point (although maybe offscreen ones?)

CyanVoxel:
> > This is something I also don't like but unfortunately don't know a way around... I actually ended up creating an entire embedded progress bar that updated the values on screen as it the library processed, but discovered that SQLAlchemy's desire to be on a single thread directly headbutts against Qt's desire for progress bars to be on different threads. I was unable to come up with a solution where I could yield feedback from the library while not ruining the db connection.
> 
> Yikes that sounds like a pain to get working. Have you thought about instead of having an indicator for progress, just having an indicator that the program is still running (e.g. a spinner)? This way the window wouldn't be frozen and it would still convey that it isn't done yet

That *might* work, assuming I can safely move the entire SQL library opening + conversion + closing process to a single separate thread. I think trying `yield` from the db before was the main issue, but I'm worried that either the GUI will still be blocked or that the DB will refuse to work on any other thread than the main one.

Computerdores:
I think that would be worth trying. Windows freezing sketches me the hell out, I have wasted way to much time waiting on programs only to kill them after a while and conclude that they must be broken (especially since that part can take multiple minutes for larger libraries)

CyanVoxel:
> I think that would be worth trying. Windows freezing sketches me the hell out, I have wasted way to much time waiting on programs only to kill them after a while and conclude that they must be broken (especially since that part can take multiple minutes for larger libraries)

Glad I kept around the progress bar code - I was able to get an indefinite progress bar that updates throughout the major migration events. I've also taken the opportunity to catch and display any errors encountered on the UI itself, as well as display a message box with any data discrepancies if necessary.

Computerdores:
> > I think that would be worth trying. Windows freezing sketches me the hell out, I have wasted way to much time waiting on programs only to kill them after a while and conclude that they must be broken (especially since that part can take multiple minutes for larger libraries)
> 
> Glad I kept around the progress bar code - I was able to get an indefinite progress bar that updates throughout the major migration events. I've also taken the opportunity to catch and display any errors encountered on the UI itself, as well as display a message box with any data discrepancies if necessary.

I've tested this and it looks really good! I think there are probably still improvements that could be made (there are a bunch of log messages from `add_field_to_entry` and they show up over a long timespan so that could probably be shown to the user), but this definitely good enough!

I will try to read over the code and give a review later

tagstudio/src/core/library/alchemy/library.py
```diff
@@ -141,32 +146,118 @@ def close(self):
...
+        # Tags
+        for tag in json_lib.tags:
+            self.add_tag(
+                Tag(
+                    id=tag.id,
+                    name=tag.name,
+                    shorthand=tag.shorthand,
+                    color=TagColor.get_color_from_str(tag.color),
+                )
+            )
```
Computerdores, line 174: This seems a bit weird to me, why are the sub tags and aliases added before the tags that their foreign keys point to? 

I haven't tested this, but I suspect that, because every operation in those loops is its own DB Transaction, should TS crash after the Aliases, but before the Tags have been added, that there would be database inconsistencies stemming from foreign keys in the Aliases and Subtags pointing to tags that don't exist. This might not matter if the migration just restarts after such a crash (or killing by the user), but seems like it is just asking for problems.

tagstudio/src/core/library/alchemy/models.py
```diff
@@ -177,9 +177,12 @@ def __init__(
         path: Path,
         folder: Folder,
         fields: list[BaseField],
+        id: int | None = None,
     ) -> None:
         self.path = path
         self.folder = folder
+        assert not self.id
```
Computerdores, line 184: Why is this necessary? Shouldn't the constructor run before anything can set this?

tagstudio/src/qt/ts_qt.py
```diff
@@ -1170,14 +1174,27 @@ def update_libs_list(self, path: Path | str):
...
         open_message: str = f'Opening Library "{str(path)}"...'
         self.main_window.landing_widget.set_status_label(open_message)
         self.main_window.statusbar.showMessage(open_message, 3)
         self.main_window.repaint()
 
-        open_status = self.lib.open_library(path)
+        open_status: LibraryStatus = self.lib.open_library(path)
+
+        # Migration is required
+        if open_status.message and open_status.message.startswith("[JSON]"):
```
Computerdores, line 1187: I am not sure using this message to figure out whether migration is required is the best idea, maybe a `migration_required` field in `LibraryStatus` would be better? (This way changing what looks like a log message doesn't break functionality)

tagstudio/src/qt/widgets/migration_modal.py
```diff
@@ -0,0 +1,785 @@
...
+    def check_color_parity(self) -> bool:
+        """Check if all JSON tag colors match the new SQL tag colors."""
+        sql_color: str = None
+        json_color: str = None
+
+        for tag in self.sql_lib.tags:
+            tag_id = tag.id  # Tag IDs start at 0
+            sql_color = tag.color.name
+            json_color = (
+                self.json_lib.get_tag(tag_id).color.upper().replace(" ", "_")
```
Computerdores, line 765: `.upper().replace(" ", "_")`

this conversion is duplicated in `TagColor.get_color_from_str` maybe that should be a static method of `TagColor`?

Computerdores (commented):
Looking quite good overall, just a couple comments

tagstudio/src/core/library/alchemy/models.py
CyanVoxel: I went and followed the assertion that was present in the Entry constructor, but yes I don't think this has ever been tripped during debugging and can probably be taken out now

Computerdores (approved):

Issue #600, opened 2024-11-22
[Feature Request]: FINAL search engine features for 9.5


python357-1:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

this issue describes the current consensus on what features the search engine will have for the 9.5 release.
# REQUIRED for 9.5:

new queries - a definition and implementation of a "query language" which can be converted to SQL queries to be made against the database

- Boolean operators: AND, OR, and NOT operators for said queries
- `tag`, `tag_id`, `mediatype`, `filetype`, `path`, and `special` queries: types of queries to be run against entries/tags in the database

| query | description | allowed values | example |
| --- | --- | --- | --- |
| tag| searches by tag name, with an optional "disambiguation" syntax that specifies which tag you are looking for, if there are multiple matches | `string` |`tag: Mario` <br> `tag: Mario[parent=nintendo]` |
| tag_id | searches by the internal ID of a tag | `int` | `tag_id: 1001` |
| mediatype | searches by `name` property of MediaCategories (will eventually use the translated name) | `string` | `mediatype: video` |
| filetype | searches by the file extension |`string` | `filetype: jpg` |
| path | searches by complete path of files. allows globs |`string` | `path: folder/*` |
| special | searches by special metadata of entries | `"untagged"` <br> `"unlinked"` | `special: untagged` |
- field content search: TODO
- sortable results (possibly in a different issue/PR)

# STRETCH GOALS for 9.5
- new search UI (will be in a separate issue/PR)
- HAS query (waiting on composition tags to be implemented)

Computerdores:
I am interested in working on a parser for this. Feedback on my current WIP grammar would be nice:
```ebnf
ANDList        = ORList ( ["AND"] ORList )* ;
ORList         = Term ( "OR", Term )* ;
Term           = Constraint | "(", ANDList, ")" ;

Constraint     = [ConstraintType, ":"], Literal, "[", PropertyList, "]" ;

ConstraintType = "tag" | "mediaType" ; (* not a complete list *)
PropertyList   = Property, (",", Property)* ;
Property       = ULITERAL, "=", Literal ;
Literal        = ULITERAL | QLITERAL ;
```
Notes:
[...]

PR #606, opened 2024-11-27
feat: Implement Query lang

Computerdores:
Implements #600

Current supported Grammar:
```ebnf
ORList         = ANDList ( "OR", ANDList)* ;
ANDList        = Term ( ["AND"] Term )* ;
Term           = Constraint | "(", ORList, ")" | "NOT", Term ;

Constraint     = [ConstraintType, ":"], Literal, "[", PropertyList, "]" ;

ConstraintType = "tag" | "tag_id" | "mediatype" | "filetype" | "path" | "special" ; (* not case sensitive *)
PropertyList   = Property, (",", Property)* ;
Property       = ULITERAL, "=", Literal ;
Literal        = ULITERAL | QLITERAL ;
```
Notes:
- `path:<glob>` searches entries using a UNIX-Style GLOB
- `special:untagged` searches for entries that don't have any tags

Limitations / Future Work:
- Sub-Tags do not work yet
- autocompletion could use an overhaul
- When query is invalid user should be informed what part of it is wrong
  - also warn / fail when searching for tag / tag id that doesn't exist
- Backwards and Forwards buttons don't work

Todo:
- [x] Try to put tag id lookup into subqueries that return tag ids. This should reduce the number of joins in the main query from 4 to ~~2~~ 1. That might drastically improve performance
- [x] let chained NOTs cancel eachother out
- [x] use [relational division queries](https://www.geeksforgeeks.org/sql-division/) on both negated and non negated Tag constraints 
- [x] fix bug that @CyanVoxel found and @python357-1 figured out the root of


python357-1:
NOT operator also seems to not work (`NOT tag_id:1000` returns no results)

Computerdores:
TODO: port `untagged` and `empty` or equivalents

Computerdores:
> NOT operator also seems to not work (`NOT tag_id:1000` returns no results) in a library with entries that dont have tag id 1000

NOT is now implemented

tagstudio/src/core/library/alchemy/visitors.py
```diff
@@ -38,7 +38,18 @@ def visit_constraint(self, node: Constraint) -> ColumnExpressionArgument:
...
 
+        # raise exception if Constraint stays unhandled
         raise NotImplementedError("This type of constraint is not implemented yet")
 
     def visit_property(self, node: Property) -> None:
         return
+
+    def visit_not(self, node: Not) -> ColumnExpressionArgument:
+        return ~Entry.id.in_(
+            # TODO TSQLANG there is technically code duplication from Library.search_library here
```
Computerdores, line 49: Not really sure on whether / how I should deal with this duplication, opinions would be appreciated

CyanVoxel:
From some surface-level testing with my existing libraries it seems that OR and NOT are working, however any form of AND always seems to return zero results for me

Computerdores:
> From some surface-level testing with my existing libraries it seems that OR and NOT are working, however any form of AND always seems to return zero results for me

Small correction: AND seems to break when more than one tag is involved, will try to fix later today

Computerdores:
It should be working now.

I was wondering for like 2h how this ever worked only to git bisect it and find out it didn't. No idea how I didn't notice that

CyanVoxel:
> It should be working now.
> 
> I was wondering for like 2h how this ever worked only to git bisect it and find out it didn't. No idea how I didn't notice that

Haha I've done the exact same thing with this project before, no worries!

Just tested it and it seems to be working now with the same library! The only thing I'm noticing, and this was a bug present for me before this PR, is that often I won't get the full number of results returned in the thumbnail grid. For multi-page results, I may only get 16, (29!?), 32, or 128 results per page instead of the full 500. Maybe this has to do with extension filtering?
I will also note that performance seems to have taken a big hit for me with this, with some searches taking up to 15-20 seconds on a pretty fast SSD.

python357-1:
I will say even one second seems like way more than I would expect, and 20 is just beyond anything reasonable. We should probably make sure it is actually the parser that caused it. I find it somewhat hard to believe that it would be taking that long to do its thing.

Computerdores:
Note to self: Try to put tag id lookup into subqueries that return tag ids. This should reduce the number of joins in the main query from 4 to 2. That might drastically improve performance

CyanVoxel:
Performance seems improved, and the longest query times I've been getting have been cut down to a max of 5-7 seconds down from 15-20.

I think I've also narrowed down the source of the incomplete page numbers to this query:
```
filter=FilterState(page_index=1, page_size=500, ast=ORList(elements=[])) query_full=SELECT entries.id, entries.folder_id, entries.path, entries.suffix 
FROM entries LEFT OUTER JOIN tag_box_fields ON entries.id = tag_box_fields.entry_id LEFT OUTER JOIN (tag_fields AS tag_fields_1 JOIN tags ON tags.id = tag_fields_1.tag_id) ON tag_box_fields.id = tag_fields_1.field_id LEFT OUTER JOIN tag_aliases ON tags.id = tag_aliases.tag_id 
WHERE (entries.suffix NOT IN ('json', 'xmp', 'aae', 'xml', 'txt', 'md', 'lnk', 'ini', 'bat', 'py', 'sqlite', 'fish', 'sh'))
```

That one returns odd numbers of results per page even on an empty search, while the following query that gets run on startup returns the full 500 results per page:
```
filter=FilterState(page_index=0, page_size=500, ast=None) query_full=SELECT entries.id, entries.folder_id, entries.path, entries.suffix 
FROM entries 
WHERE (entries.suffix NOT IN ('json', 'xmp', 'aae', 'xml', 'txt', 'md', 'lnk', 'ini', 'bat', 'py', 'sqlite', 'fish', 'sh'))
 LIMIT 500 OFFSET 0
 ```
 
I also seem to be missing results for certain queries, but I don't know if that's related to the above issue or not. I'll work on putting together a test library to help get some reproducible and sharable results.

**Comment by CyanVoxel** [comment]
I've put together a minimal reproducible example of the page mismatch issue I've been having. This just contains a `ts_library.sqlite` file with no actual image files, but works for searching just as well. It contains **600** entries, **100** of which have the Favorite tag.

Steps to reproduce issue:
1. Open library
2. Select all items on first page, observe count at **500** (Optional: navigate to second page and observe that count at **100**)
3. Enter a blank search or click the search button with the search field blank
4. Select all items on first page, observe count at **200**
5. Navigate to second page, select all items and observe count at **250**

Along with the unfilled pages, a total of **150** entries do not show up in the library after this point

[Page Mismatch Example.zip](https://github.com/user-attachments/files/17997241/Page.Mismatch.Example.zip)

**Comment by Computerdores** [comment]
I have to say I have no idea at this point how the results are going missing. The query only uses left outer joins on the data that contains the Entry ids and that should to my understanding mean that the Joins do not exclude any Entry IDs. Furthermore the WHERE clause is the exact same between both of them so that can't be the issue either.

So from reading the queries I would normally confidently say that nothing is going missing, but alas...

**Comment by python357-1** [comment]
idea: the main query that gets run is basically this:
```
SELECT entries.id, entries.folder_id, entries.path, entries.suffix
FROM entries
LEFT OUTER JOIN tag_box_fields ON entries.id = tag_box_fields.entry_id
LEFT OUTER JOIN (tag_fields AS tag_fields_1 JOIN tags ON tags.id = tag_fields_1.tag_id) ON tag_box_fields.id = tag_fields_1.field_id
LEFT OUTER JOIN tag_aliases ON tags.id = tag_aliases.tag_id 
WHERE (entries.suffix NOT IN ('.json', '.xmp', '.aae'))
LIMIT XXX OFFSET YYY
```

however, if you run this against the database in the zip Cyan provided, you will get results similar to this:
```
id|folder_id|path            |suffix|
--+---------+----------------+------+
 1|        1|mono copy.png   |png   |
 1|        1|mono copy.png   |png   |
 1|        1|mono copy.png   |png   |
 2|        1|mono copy 71.png|png   |
 2|        1|mono copy 71.png|png   |
 2|        1|mono copy 71.png|png   |
 3|        1|mono copy 65.png|png   |
 3|        1|mono copy 65.png|png   |
 3|        1|mono copy 65.png|png   |
 4|        1|mono copy 59.png|png   |
```
i believe what may be happening is the query is getting duplicate values from the database as shown above, and THEN removing duplicate values. most entries are being returned 2 or 3 times, so (assuming an average of 2.5 results per entry) if you got the unique values out of that 500, you would only have 250 items left.

**Comment by python357-1** [comment]
more explanation of what im talking about: in `search_library`, there is the following line:
```
statement = statement.limit(search.limit).offset(search.offset)
```

AFTER that, the query is set to only have unique values (which i believe is happening in-memory because it is not in the query):
```
res = SearchResult(
    total_count=count_all,
    items=list(session.scalars(statement).unique()),
)
```

if we get unique values first, THEN limit and offset, this problem should go away.

**Comment by Computerdores** [comment]
You are right, thank you so much!! I would have wasted hours to find this, I didn't even consider the deduplication

**Comment by Computerdores** [comment]
The bug has been fixed, the number of Joins reduced and NOTs now cancel out in the AST!

My hope is that those last to should have substantially improved the performance of the search, but I don't have a fast way of testing that.
@CyanVoxel could you maybe try it on your library where you had performance issues before?

**Comment by python357-1** [comment]
can confirm bug is fixed

**Comment by Computerdores** [comment]
I have also somewhat modified the autocompletion to improve how well it works with this PR. It isn't perfect (or even good imo), but I think it is better than nothing

**Comment by CyanVoxel** [comment]
Bug is also confirmed fixed on my end, and performance is *much* better at around a second per (typical) query 👍

**Comment by Computerdores** [comment]
I have now implemented the optimization with Relational Division Queries, which, for me, has improved performance by another 90% from ~1s to ~0.1s search time.

**Comment by python357-1** [comment]
i wasnt seeing 1s (or even 0.1s) search times to begin with but its definitely not slower so it looks good to me

**Comment by CyanVoxel** [comment]
I'm seeing about a 50% improvement, which is still very nice

**Comment by Computerdores** [comment]
> I'm seeing about a 50% improvement, which is still very nice

What kind of search times are you at now? For me the thumbnail stuff is now much more significant than the search

**Comment by Computerdores** [comment]
Also, in 9.4 the forward and back buttons next to the search cycle through a history of past queries, while on main now they don't. Is this desired?

**Comment by CyanVoxel** [comment]
> > I'm seeing about a 50% improvement, which is still very nice
> 
> What kind of search times are you at now? For me the thumbnail stuff is now much more significant than the search

On the mac I've been testing with, about half a second average for most queries. The lowest number I've seen is .3 seconds and the highest is about 1.7 seconds, though it seems like re-running queries attributes to this lower number

**Comment by CyanVoxel** [comment]
> Also, in 9.4 the forward and back buttons next to the search cycle through a history of past queries, while on main now they don't. Is this desired?

The history states are still desired, this was just another thing that got broken along the way to sql...

**Comment by Computerdores** [comment]
> On the mac I've been testing with, about half a second average for most queries. The lowest number I've seen is .3 seconds and the highest is about 1.7 seconds, though it seems like re-running queries attributes to this lower number

Are you happy with this (at least for now)?

There are probably still places where I could eek out some performance, but it would delay getting this merged and be extra work.

**Comment by CyanVoxel** [comment]
> > On the mac I've been testing with, about half a second average for most queries. The lowest number I've seen is .3 seconds and the highest is about 1.7 seconds, though it seems like re-running queries attributes to this lower number
> 
> Are you happy with this (at least for now)?
> 
> There are probably still places where I could eek out some performance, but it would delay getting this merged and be extra work.

I think the performance is pretty dang good for now, and surely is better than 9.4 👍

**Comment by Computerdores** [comment]
Okay then I am calling a self imposed stop on the feature creep in this PR. I will obviously address any comments, but I don't want to infinitely delay this by adding more and more.

Subtags and properties are still needed, but I think it would be easier to review and organise if those are not done in this PR.

Therefore unless you disagree, I think this PR is ready for final review.

**Comment by CyanVoxel** [comment]
> Okay then I am calling a self imposed stop on the feature creep in this PR. I will obviously address any comments, but I don't want to infinitely delay this by adding more and more.
> 
> Subtags and properties are still needed, but I think it would be easier to review and organize if those are not done in this PR.
> 
> Therefore, unless you disagree, I think this PR is ready for final review.

Works for me! This set of features will be super valuable to have in the meantime. I also agree with Python, some tests in here would be great to have before we merge. In the meantime, I'll do my best to break this 😁

I also wanted to ask, are parenthesis supported in this PR? That's the impression I got from the grammar but I may be mistaken. In the event that they are, I ask because I always get syntax errors when trying to use them.

**Comment by Computerdores** [comment]
> Works for me! This set of features will be super valuable to have in the meantime. I also agree with Python, some tests in here would be great to have before we merge. In the meantime, I'll do my best to break this 😁

Is there a bigger library for tests around than the fixture with like 3 entries? If not I'll just migrate the json library that is in there and use that for testing the search.

> I also wanted to ask, are parenthesis supported in this PR? That's the impression I got from the grammar but I may be mistaken. In the event that they are, I ask because I always get syntax errors when trying to use them.

Parentheses should be fixed now!

**Comment by python357-1** [comment]
Yeah, having the bigger test library for this would be helpful, but that sounds like it would take quite a while to get done. Am I wrong @CyanVoxel ?

**Comment by CyanVoxel** [comment]
> Yeah, having the bigger test library for this would be helpful, but that sounds like it would take quite a while to get done. Am I wrong @CyanVoxel ?

For a *big* library, yes that'll probably take a while. But I've started working towards this, and already have something that should help out with search test cases - especially when parent tag searching comes down the line:
[Search Example.zip](https://github.com/user-attachments/files/18013129/Search.Example.zip)

**Comment by Computerdores** [comment]
I have now written every test I could think of and then some.
Test Coverage of `query_lang/*` and `visitors.py` is at 92%, I think this should be good enough.

(Going through the code you might see `# pragma: nocover`, this marks lines as excluded from the coverage calculation. I used this to exclude liens that will only ever be run during development / debugging)

File: tagstudio/tests/fixtures/search_library/.TagStudio/ts_library.json [hunk 2]
```diff
@@ -0,0 +1 @@
...:"inherit colors shapes","fields":[{"6":[1011]}]},{"id":22,"filename":"yellow_ellipse.png","path":"inherit colors shapes","fields":[{"6":[1042,1013]}]},{"id":23,"filename":"ellipse.png","path":"inherit colors shapes","fields":[{"6":[1042]}]},{"id":24,"filename":"red_ellipse.png","path":"inherit colors shapes","fields":[{"6":[1042,1009]}]},{"id":25,"filename":"blue_ellipse.png","path":"inherit colors shapes","fields":[{"6":[1021,1042]}]},{"id":26,"filename":"green_ellipse.png","path":"inherit colors shapes","fields":[{"6":[1042,1017]}]},{"id":27,"filename":"orange_ellipse.png","path":"inherit colors shapes","fields":[{"6":[1042,1011]}]},{"id":30,"filename":"r_circle_b_square.png","path":"comp colors shapes","fields":[{"6":[1021,1041,1009,1038]}]},{"id":31,"filename":"r_circle_g_square.png","path":"comp colors shapes","fields":[{"6":[1041,1017,1009,1038]}]},{"id":32,"filename":"r_circle_y_square.png","path":"comp colors shapes","fields":[{"6":[1041,1009,1038,1013]}]},{"id":33,"filename":"r_circle_o_square.png","path":"comp colors shapes","fields":[{"6":[1041,1011,1009,1038]}]},{"id":34,"filename":"r_circle_r_square.png","path":"comp colors shapes","fields":[{"6":[1041,1009,1038]}]}]}
```
CyanVoxel, line 1: Note to self: The JSON file for this test library shouldn't be necessary to include, but I'll allow it here as I'm still using 9.4 to create libraries for testing due to alias and field bugs on main.

tagstudio/tests/test_library.py
```diff
@@ -194,8 +190,8 @@ def test_add_field_tag(library, entry_full, generate_tag):
     library.add_field_tag(entry_full, tag, tag_field.type_key)
 
     # Then
-    results = library.search_library(FilterState(id=entry_full.id))
-    tag_field = results[0].tag_box_fields[0]
+    results = library.get_entry_full(entry_full.id)
+    tag_field = results.tag_box_fields[0]
```
CyanVoxel, line 194: Nit: match singular result name from other `get_entry_full()` calls for consistency
```suggestion
    result = library.get_entry_full(entry_full.id)
    tag_field = result.tag_box_fields[0]
```

CyanVoxel (approved):
Looks fantastic! Works great on my end, and I don't even have much in terms of review feedback left at this point.

Thank you so much for your amazing work on this!!

tagstudio/tests/fixtures/search_library/.TagStudio/ts_library.json
Computerdores: Yeah I pretty much figured as long as it could still be useful we might as well keep it in the repo instead of just as a link in some random comment on some PR

Computerdores:
No idea what is going with ruff and mypy, 9d357b0fba2732719dfe9f52c8676a2e7a1212b0 passed both and on 4265fde7ffad3af479be68800c36dca2e222ef18 they suddenly fail in files that weren't even modified

CyanVoxel:
> No idea what is going with ruff and mypy, [9d357b0](https://github.com/TagStudioDev/TagStudio/commit/9d357b0fba2732719dfe9f52c8676a2e7a1212b0) passed both and on [4265fde](https://github.com/TagStudioDev/TagStudio/commit/4265fde7ffad3af479be68800c36dca2e222ef18) they suddenly fail in files that weren't even modified

Hmm, ruff was updated on main recently but I don't see how the mypy error is occurring suddenly

Computerdores:
Let me guess the stuff that mypy/ruff criticize here has been fixed on main? I would guess that the check probably compares the errors of my branch to the errors that happen on the branch I want to merge onto and sees that they aren't there on main so points them out

VasigaranAndAngel:
> No idea what is going with ruff and mypy, [9d357b0](https://github.com/TagStudioDev/TagStudio/commit/9d357b0fba2732719dfe9f52c8676a2e7a1212b0) passed both and on [4265fde](https://github.com/TagStudioDev/TagStudio/commit/4265fde7ffad3af479be68800c36dca2e222ef18) they suddenly fail in files that weren't even modified

it's strange. i didn’t encounter any errors on my system (windows and ubuntu 22.04) using the same `ruff` and `mypy` versions as specified in the workflow. this could be due to:

```python
from src.core.library.alchemy import Library
from src.core.library.alchemy.library import Entry, LibraryStatus
```
`Entry` and `LibraryStatus` are imported directly by importing library and not using the `__init__.py` _?_
in `src/core/library/alchemy/`, there’s an `__init__.py` file. (contains `__all__` attribute also):

```python
from .enums import ItemType
from .library import Library
from .models import Entry, Tag

__all__ = ["Entry", "Library", "Tag", "ItemType"]
```

isn’t that odd? but it’s just a weird assumption for a weird issue. nvm.

> ruff was updated on main recently  

the `ruff` and `mypy` versions are hard coded in the workflows, so they aren’t loaded from `requirements.txt`,

https://github.com/TagStudioDev/TagStudio/blob/4265fde7ffad3af479be68800c36dca2e222ef18/.github/workflows/ruff.yaml#L10-L20
https://github.com/TagStudioDev/TagStudio/blob/4265fde7ffad3af479be68800c36dca2e222ef18/.github/workflows/mypy.yaml#L26-L29

PR #608, opened 2024-11-28
fix: ui/ux parity fixes for thumbnails and files

CyanVoxel:
This PR adds various fixes and improvements for increased parity of thumbnail and file entry behavior between v9.4 and v9.5. This includes:
- Fix: Thumbnail loading icons are properly displayed before the main render jobs start
- Fix: Tag color names now show correct display names
- Fix: Preview panel fields no longer leave previous UI elements around when browsing
- Parity: Optimized file refreshing by removing string manipulation operations and re-introducing a cache of known files to skip when re-scanning
- Feat: Add new folder names to always ignore when scanning
- Fix/Parity/Feat: Automatic library refreshing on start is now only enabled for libraries with <10,000 previously known files

### DIFF
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -1203,7 +1203,8 @@ class QtDriver(DriverMixin, QObject):
         self.filter.page_size = self.lib.prefs(LibraryPrefs.PAGE_SIZE)
 
         # TODO - make this call optional
-        self.add_new_files_callback()
+        if self.lib.entries_count < 10000:
+            self.add_new_files_callback()
 
         self.update_libs_list(path)
         title_text = f"{self.base_title} - Library '{self.lib.library_dir}'"
--- a/tagstudio/src/qt/ts_qt.py
+++ b/tagstudio/src/qt/ts_qt.py
@@ -1055,6 +1055,9 @@ class QtDriver(DriverMixin, QObject):
         for idx, (entry, item_thumb) in enumerate(
             zip_longest(self.frame_content, self.item_thumbs)
         ):
+            if not entry:
+                continue
+
             filepath = self.lib.library_dir / entry.path
             is_loading = False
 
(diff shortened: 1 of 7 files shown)


tagstudio/src/qt/ts_qt.py
```diff
@@ -1030,25 +1030,41 @@ def update_thumbs(self):
...
 
-            filepath = self.lib.library_dir / entry.path
-            item_thumb = self.item_thumbs[idx]
             item_thumb.set_mode(ItemType.ENTRY)
             item_thumb.set_item_id(entry)
 
             # TODO - show after item is rendered
             item_thumb.show()
 
+            is_loading = True
```
Computerdores, line 1046: could this not be hardcoded? As far as I can tell this is never changed and only accessed once
(same in the next loop)

tagstudio/src/qt/ts_qt.py
```diff
@@ -1187,7 +1203,8 @@ def open_library(self, path: Path) -> LibraryStatus:
         self.filter.page_size = self.lib.prefs(LibraryPrefs.PAGE_SIZE)
 
         # TODO - make this call optional
-        self.add_new_files_callback()
+        if self.lib.entries_count < 10000:
+            self.add_new_files_callback()
```
Computerdores, line 1207: Is this change what the todo refers to? If so can it be removed?

Computerdores (approved):
Looks good overall, just a couple nits

tagstudio/src/qt/ts_qt.py
CyanVoxel: It *can* be hard-coded, but the reason I split it off into a variable is due to the way the arguments are being passed where I'm unable to use named parameters. It took some time to sort out which of the existing unnamed boolean arguments were doing what here, so I opted to store them in named variables for clarity.

tagstudio/src/qt/ts_qt.py
CyanVoxel: I left the `TODO` here since while it originally referred to the persistent call, it still works as a `TODO` to add the option to customize the new threshold here.

Computerdores (approved):

PR #609, opened 2024-11-28
fix: shut up mypy

Computerdores:
Just four extremly short fixes in order to stop mypy from complaining constantly. Specifically:
- two type ignore comments threw unused-ignore on windows (but didn't on linux), the type ignores now also ignore unused-ignore
- a call in ts_qt was missing a trivial argument
- two variables in file_opener had the same name but different type and mypy wasn't happy about that, I renamed one appropriately


CyanVoxel:
The `preview_panel.py` one I finally snuffed out in #608, but these should be some much welcomed fixes 👍

CyanVoxel (approved):

Issue #610, opened 2024-11-28
[Bug]: _match_missing_file walks library for every missing file


Toby222:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Alpha 9.4.0+

### Operating System & Version

NixOS unstable

### Description

the Search & Relink feature for missing files is incredibly slow for larger libraries, since for every missing entry it walks the library directory again. Extremely noticeable on larger library.

(I noticed this when reorganizing my photos folder from `YYYY/` folders to `YYYY/MM/DD/` folders, so there were ~11000 moved files)

### Expected Behavior

The list of files in the library should be cached between calls to _match_missing_file to speed up unlinked entry Relinking.
In a bodge solution I added a simple member `_cache` to the Library class, and in _match_missing_file initialize it with the result of os.walk, if not already set

### Steps to Reproduce

1. Create a library with a lot of entries
2. Move multiple files to a new directory
3. Search & Relink the now-missing entries
4. Observe that between every entry a second or more can pass

CyanVoxel:
I would call this more of a request/need for optimization than a bug since the code is working as intended, but I do agree that this area of the code could *really* use some optimizing.

Toby222:
Between feature request and bug, I felt like "doesn't work as it should" fit better than "doesn't work like how I want"

dominic248:
Hi @CyanVoxel for auto-relink of files on rename or moved files, we can have...
1) a prefix in the filename like id of file in DB (drawbacks - if user removes the prefix, then manual linking is needed)
2) hash value of file (drawbacks - if file is edited then hash value will change and manual linking is needed)
3) combination of both prefix id & hash value for auto-relink, incase user mistakenly deleted the prefix , then the program can rename the file & add prefix, based on the hash match (drawbacks - file with same hash/duplicate files)

PR #611, opened 2024-11-28
Improve performance of relinking missing entries

Toby222:
Basic fix for #610
Haven't looked at other things that could be improved, nor is this thoroughly tested yet

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -141,6 +141,9 @@ venv.bak/
 # Rope project settings
 .ropeproject
 
+# macoS
+*.DS_Store
+
 # mkdocs documentation
 /site
 


CyanVoxel:
I see this is targeting main (9.5) when the code is changing the older JSON (9.4.x) library code. I would suggest either targeting this towards the [`Alpha-v9.4`](https://github.com/TagStudioDev/TagStudio/tree/Alpha-v9.4) branch in time for the upcoming 9.4.2 patch, or preferably implement the code in the [SQLite portion](https://github.com/TagStudioDev/TagStudio/blob/main/tagstudio/src/core/library/alchemy/library.py) of the code for use on TagStudio versions going forward.

Toby222:
Right, I did that in 9.4 then rebased onto main without thinking when updating my fork; will do tomorrow

Toby222:
I forgot github does this when force pushing..
I think I did that last time too

PR #613, opened 2024-11-29
Small fix: Remove unnecessary warning in console

Thesacraft:
Hi,
I removed the Warning that is shown in the console that occurs when Tagstudio is opened and a thumbnail is clicked for the first time.

Bug:
grafik

Fix:
Check if provided callback function is actually a Callable and preventing it from trying to disconnect when the callback is None.

### DIFF
--- a/tagstudio/src/qt/widgets/fields.py
+++ b/tagstudio/src/qt/widgets/fields.py
@@ -113,7 +113,7 @@ class FieldContainer(QWidget):
 
         self.copy_callback = callback
         self.copy_button.clicked.connect(callback)
-        self.copy_button.is_connected = isinstance(callback, Callable)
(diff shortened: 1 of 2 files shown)


CyanVoxel (commented):
Looks pretty good! The only suggestion I have is that inside `QPushButtonWrapper`, `self.remove_button.is_connected` is still set to `False` by default despite it no longer being used as a boolean value. If this value is now dependent on it containing a `callable` or not, I would suggest changing this line to `self.is_connected: callable | None = None`

Thesacraft:
Hi, callable is an inbuilt function that should return True if it's a callable that's passed in and otherwise returns false from what I know. So it should still be a boolean.

PR #615, opened 2024-11-30
Rating fields

Nginearing:
- Adds widget for rating field type
- Adds field type "Rating Box" and field "Rating"

I need some help to get it editable without an edit menu, this code has been moved over from 9.4 and I'm not currently able to test on the main branch, so this code may not immediately work. Any help would be appreciated.


Cool-Game-Dev:
Let me test this one out

PR #622, opened 2024-12-02
fix: clear all setting values when opening a library

VasigaranAndAngel:
Replaced `self.settings.clear()` with `self.settings.remove("")` in the `update_libs_list()` method. the original implementation was clearing all settings. this bug prevented the app from reopening the last library also. and surprisingly no one noticed or reported it. this change ensures only the intended group of settings is cleared without affecting other values.


Computerdores (approved):
>  this bug prevented the app from reopening the last library also.

It didn't for me.

All in all seems to work and looks good!

CyanVoxel:
I didn't seem to encounter the bug on macOS, but I *do* seem to encounter it on Windows, and this seems to address it!
Thank you for the fix!

PR #623, opened 2024-12-02
feat: user defined colors

### LINKED ISSUE #264: [Feature Request] Hex/RGB Color Values for Tags

python357-1:
This PR implements user-defined tag colors (discussed  on the roadmap and would close #264)


python357-1:
**Initial ideas for user-defined colors**

This implementation would add a new table for user defined colors and pre-populate it with the default colors. This would allow for the colors to be ID based, as mentioned on the roadmap, and allow them to have custom names and hex values. The roadmap also mentions having the base colors be non-removable, which is what the following trigger does:
```
CREATE TRIGGER delete_color BEFORE DELETE ON user_defined_colors
WHEN OLD.user_defined = FALSE
BEGIN
    SELECT RAISE(ABORT, 'Cannot delete program-defined colors');
END;
```
User-defined colors can still have their names and hex values updated to whatever the user wants, but they cannot be removed from the database - any attempts to do so will result in a SQL error.

python357-1:
@CyanVoxel Do these user-defined colors need support for all the ColorTypes like in the current `palette.py`? ex:
```
TagColor.DEFAULT: {
        ColorType.PRIMARY: "#1e1e1e",
        ColorType.TEXT: ColorType.LIGHT_ACCENT,
        ColorType.BORDER: "#333333",
        ColorType.LIGHT_ACCENT: "#FFFFFF",
        ColorType.DARK_ACCENT: "#222222",
    },
```

CyanVoxel:
> @CyanVoxel Do these user-defined colors need support for all the ColorTypes like in the current `palette.py`? ex:
> 
> ```
> TagColor.DEFAULT: {
>         ColorType.PRIMARY: "#1e1e1e",
>         ColorType.TEXT: ColorType.LIGHT_ACCENT,
>         ColorType.BORDER: "#333333",
>         ColorType.LIGHT_ACCENT: "#FFFFFF",
>         ColorType.DARK_ACCENT: "#222222",
>     },
> ```

Thank you for asking! The plan is to get rid of this system, and have all colors (built-in + user) be based off of a single hex value per color. From there, we can extrapolate any accent colors and detect whether it needs dark or light text programmatically. That function would likely need a lot of back-and-forth tweaking so I wouldn't worry about nailing it, just having something that works decently enough in its place.

python357-1:
Were you thinking this function would just take a ColorType and a hex value/Color?

tagstudio/src/core/library/alchemy/enums.py
```diff
@@ -42,12 +43,19 @@ class TagColor(enum.IntEnum):
     COOL_GRAY = 36
     OLIVE = 37
 
-    @staticmethod
-    def get_color_from_str(color_name: str) -> "TagColor":
-        for color in TagColor:
-            if color.name == color_name.upper().replace(" ", "_"):
-                return color
-        return TagColor.DEFAULT
+#class TSStandardColors(enum.IntEnum):
```
python357-1, line 46: I know you said you aren't a fan of enums, but I think it would be good to have an enum that just has the IDs of colors in the default tagstudio color pack (whatever we will call it).

tagstudio/src/core/library/alchemy/library.py
```diff
@@ -91,6 +91,51 @@ def get_default_tags() -> tuple[Tag, ...]:
     return archive_tag, favorite_tag
 
 
+def get_default_colors(namespace: ColorNamespace) -> list[Color]:
+    colors: list[Color] = [
+        #TODO: reduce this to limited tagstudio standard color set
```
python357-1, line 96: I don't know if you have settled on a final list of colors you do want in the default pack, but the sooner I can have that the better

tagstudio/src/core/utils/color.py
```diff
@@ -0,0 +1,15 @@
+from src.core.library.alchemy.models import Color, ColorNamespace
+from src.core.palette import ColorType
+
+def get_hex_color(color: Color, color_type: ColorType) -> str:
+    #TODO: everything
```
python357-1, line 5: this is obviously just a placeholder, I don't even know if this will be the final location for it, but for right now anything you want to do with the color function can be done here

python357-1 (commented):
@CyanVoxel i made some comments for stuff i wanted your input on

tagstudio/src/core/library/alchemy/enums.py
CyanVoxel: I'll concede to this ;)

tagstudio/src/core/library/alchemy/library.py
CyanVoxel: No problem, I'll get that fleshed out for you

tagstudio/src/core/utils/color.py
CyanVoxel: Assuming the `Color` object will take care of the color ID + namespace identification, I think this should take care of it!

PR #633, opened 2024-12-10
feat(ui): show filenames in thumbnail grid (Closes #85)

### LINKED ISSUE #85: [Suggestion] File name displayed in grid view
Maybe displaying the filename next to the file type would be helpfull. Because currently their is no real difference between diffrent files when no preview is available. So this would make it a lot easier to distinguish files without previews.

CyanVoxel:
This PR adds the ability display an entry's filename underneath its thumbnail in the grid view. This feature can be toggled on/off through the View -> "Show Filenames in Grid" menu option. I've also organized some of the existing menu options to make more sense, notably removing the "Window" menu.

image


tagstudio/tests/fixtures/search_library/.TagStudio/ts_library.sqlite
Computerdores, line 1: Are there intentional changes to this? If not should probably not merge this for fear of overwriting other intentional changes (Note: Maybe we should add this file to the `.gitignore` so it doesn't get accidentally included in commits?)

tagstudio/src/qt/ts_qt.py
```diff
@@ -322,6 +314,17 @@ def start(self) -> None:
...
+
+        open_on_start_action = QAction("Open Library on Start", self)
+        open_on_start_action.setCheckable(True)
+        open_on_start_action.setChecked(
+            bool(self.settings.value(SettingItems.START_LOAD_LAST, defaultValue=True, type=bool))
+        )
+        open_on_start_action.triggered.connect(
+            lambda checked: self.settings.setValue(SettingItems.START_LOAD_LAST, checked)
+        )
+        file_menu.addAction(open_on_start_action)
```
Computerdores, line 327: The "Open Library on Start" option exists twice, once in View and once in File, and (un)checking one of them doesn't update the visual state of the other

Computerdores (commented):
Should the "Open Library on Start" / "Show Recent Libraries" Stuff even be included here?

tagstudio/tests/fixtures/search_library/.TagStudio/ts_library.sqlite
CyanVoxel: Shoot, I didn't realize something happened to this file when committing (nor *what*) - and since it's part of a test that we can change I don't think we can get away with adding it to the `.gitignore`...

CyanVoxel:
> Should the "Open Library on Start" / "Show Recent Libraries" Stuff even be included here?

The incorrect placements and names of these actions were getting in my way when adding the new action, so I went ahead and moved them to more accurate places, both in the code and in the menus themselves.

Computerdores:
During the start I get a really weird issue where a small window will rapidly open and then close again:
image
(you can see how in the task bar there is two little bars, those are from instances of that little window in the middle)
I think this rapidly opening and closing window might have something to do with the splash, however the splash is still shown.

First bad commit is: a93809df63223db3d4abeecb635ad2ab51bb15ca.

Computerdores:
> > Should the "Open Library on Start" / "Show Recent Libraries" Stuff even be included here?
> 
> The incorrect placements and names of these actions were getting in my way when adding the new action, so I went ahead and moved them to more accurate places, both in the code and in the menus themselves.

Maybe reference that in the PR description / title, but sounds good

CyanVoxel:
> During the start I get a really weird issue where a small window will rapidly open and then close again: image (you can see how in the task bar there is two little bars, those are from instances of that little window in the middle) I think this rapidly opening and closing window might have something to do with the splash, however the splash is still shown.
> 
> First bad commit is: [a93809d](https://github.com/TagStudioDev/TagStudio/commit/a93809df63223db3d4abeecb635ad2ab51bb15ca).

I got this error yesterday on Windows as well, but it should've been fixed by 901ec7cc42cb423f05cf90784efc77b77991872c? 👀

CyanVoxel:
> > > Should the "Open Library on Start" / "Show Recent Libraries" Stuff even be included here?
> > 
> > 
> > The incorrect placements and names of these actions were getting in my way when adding the new action, so I went ahead and moved them to more accurate places, both in the code and in the menus themselves.
> 
> Maybe reference that in the PR description / title, but sounds good

This is what I was referring to in the description with "I've also organized some of the existing menu options to make more sense, notably removing the "Window" menu", but maybe I could've been more clear about it

tagstudio/tests/fixtures/search_library/.TagStudio/ts_library.sqlite
Computerdores: maybe we could add something to the test harness that automatically reverts any changes to it when cleaning up, unless it had already been changed during init of the harness?
(not sure about the details; reverting changes should be easy, but checking whether it has been changed, no idea)

Computerdores: Further note: There is a tool called sqldiff made by SQLite that checks two sqlite files for differences, maybe that could be used.
It only checks whether data in the tables has changed, while git would also report metadata changing. So if sqldiff finds no differences but git does it could be reset?

CyanVoxel: Having the tests clean it up could work as a temporary solution or as a good solution for tests that require modifying a fixture database, however this database should *not* be being modified here and signals that there's a deeper issue.

We also went in depth with sqldiff, sqlite cli, git text diffs, even a hex editor yesterday and narrowed down the problem a bit. It seems to be the internal [file change counter](https://www.sqlite.org/fileformat.html) value being updated, and I've opened up a dedicated issue for it here: #644

Computerdores (approved):
Everything was addressed and I wasn't able to find any other issues!

PR #635, opened 2024-12-11
feat: add VERSION_NUMERIC constant

### LINKED ISSUE #634: [Feature Request]: Easy test of current version of TagStudio
### Checklist

### Description

[Forward features for plugins]

Add a easy way to test the current version of TagStudio running (a string is not a handy format)

### Solution

Adding a numeric tuple version 

```py
VERSION_NUMERIC: tuple[int, int, int] = (9, 5, 0) # Major, Minor, Patch

if VERSION_NUMERIC >= (9, 5, 0):
    # is SQL database
else:
    # is JSON database
```

### Alternatives

_No response_

un-pogaz:
Adding a numeric tuple version to easily test the current version of TagStudio running. Close #634.

### DIFF
--- a/tagstudio/src/core/constants.py
+++ b/tagstudio/src/core/constants.py
@@ -1,4 +1,5 @@
-VERSION: str = "9.5.0"  # Major.Minor.Patch
+VERSION_NUMERIC: tuple[int, int, int] = (9, 5, 0)  # Major, Minor, Patch
+VERSION: str = ".".join(map(str, VERSION_NUMERIC))
 VERSION_BRANCH: str = "EXPERIMENTAL"  # Usually "" or "Pre-Release"
 
 # The folder & file names where TagStudio keeps its data relative to a library.


CyanVoxel:
While this is a useful change for testing the version number, I don't feel that it's immediately necessary and is just making a change based off of future assumptions (as #634 mentions, "[Forward features for plugins]").

The version number is something I've liked to store somewhere else for quite some time now, so that it's available to other parts of the program + project, most notably the build workflows. This will probably involve keeping it inside the pyproject.toml instead of this constants file. If I were to pull this, it would mean either breaking that future use case or just creating more refactor work for us in the future with little to no benefit in the meantime, as nothing currently uses this tuple check.

If the version number was kept in the pyproject.toml and then loaded into the project as a string and/or tuple (old JSON code checks the string iirc) then I think that would have the best of both worlds here.

PR #637, opened 2024-12-11
feat: Adding Qt Wrapper

### LINKED ISSUE #636: [Feature Request]: Improve migrating of Qt version
### Checklist

### Description

[Forward feature for plugins]

Add a implementation to prevent plugin to break when changing of Qt version.

So, if TagStudio change of Qt version, migrate the import of the main app is normaly not a issue, but things become incredibly complex if you decide to implement a third-party plugin system, because third-party plugins are not updated at the same time as TagStudio.

If TagStudio change of Qt version, all the third party plugins will immediately stop working when the user will update TagStudio, as they will not be able to import the previous version of Qt implementation.

<br>

So, Yes, the plugins are far from being here, but I think that sooner we implement this feature, sooner we'll have solved a huge problem before *when* it will happen.

### Solution

Their is many way to solve this, but we can learn for our elder: [Calibre](https://github.com/kovidgoyal/calibre) (by Kovid Goyal) is a very honorable Python/Qt open sources project who lived across 3 Qt version (Qt4 -> Qt5 -> Qt6).

I recomend to implement the same solution as Calibre: a Qt Wrapper that is used instead of the default Qt import. See #637.

### Alternatives

_No response_

un-pogaz:
> Flattering all the Qt elements into one central module.
> Created by Kovid Goyal, for calibre https://github.com/kovidgoyal/calibre
> BREAKING CHANGE: All previous PySide6 import has been replaced by src.qt.core

Close #636

Adding a Qt Wrapper that allow a more easy import of the Qt elements accross versions.

The wrapper has 2 goals:

The first is to "flattering" all the Qt elements into one central module. In original Qt implementation, you always need to know in wich submodule are the elements you want to import (QtCore? QtWidgets? The root module?). With the wrapper, this question won't bother you anymore: All the frequently use elements are  in `qt.core`.

(Note that it's possible to group together others Qt submodules if they have a specific use, as I've done by create `qt.ui_tools` and `qt.renderer`)

But the main one is to migration. With this wrapper, it is possible to change version of PySide in a very easy and transparent way by changing only 1 variable in 1 file. And that is lot more important that it's sound like.

So, if migrate the import of the main app is normaly not a issue, but things become incredibly complex if you decide to implement a third-party plugin system, because third-party plugins are not updated at the same time as TagStudio.

If TagStudio change of Qt version, all the third party plugins will immediately stop working on TagStudio updates, as it will not be able to import the previous version of Qt implementation.
Their is many way to solve this, but we can learn for our elder: [Calibre](https://github.com/kovidgoyal/calibre) (by Kovid Goyal) is a very honorable Python/Qt open sources project who lived across 3 Qt version (Qt4 -> Qt5 -> Qt6), and this one a create this proposed wrapper to make a suitable and seamlessly solution to import the Qt elements across multiple versions of the main app, regardless the Qt versions.

<br>

So, Yes, the plugins are far from being here, and Yes this do a breaking change by remplacing all PySide6 import by `qt.core` which will be rather inconvenient for the various pull requests currently on hold, **but** I think that sooner we implement this feature, sooner we'll have solved a huge problem before *when* it will happen.

Plus, is realy more ieasy to use that the default implementation.


CyanVoxel:
Thank you for your work on this, however I have concerns with the nature of this PR. Fundamentally, this refactor is preemptively trying to prevent an assumed future scenario. While it's true that we'd like to add a [plugin system](https://docs.tagstud.io/updates/roadmap/#post-100) in the future, this is the *far* future - after the program is considered "feature complete". This system has not been designed, and any assumptions made here about TagStudio using future Qt versions years down the road, or plugins using Python and PySide directly, or TagStudio still being based on Python and Qt at all, are unfortunately just shots in the dark. 

While I appreciate the concerns here and the desire to prevent issues early, unfortunately I feel that this refactor is just going to cause us more harm than good. A refactor like this is going to be highly disruptive to the project, not just in the short term but in the longer term as we would have to actively maintain this wrapper and work around it instead of making direct imports from known Qt module locations. To summarize, there's guaranteed headaches and setbacks in the short to medium term with this PR with a low chance that it would be worth it in the long term - and if we did decide that this was a necessary change for whatever future plugin system we design, then this change would be better off being made closer to that future decision than in the present.

PR #639, opened 2024-12-12
chore: update pytest workflow to 24.04

python357-1:
This PR updates the pytest workflow to Ubuntu 24.04.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -26,7 +26,7 @@ jobs:
             x11-utils \
             libyaml-dev \
             libgl1 \
-            libglx-mesa0 \
+            libegl1 \
             libxcb-icccm4 \
             libxcb-image0 \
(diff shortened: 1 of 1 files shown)


python357-1:
@CyanVoxel it appears the only thing stopping pytest from working on ubuntu 24.04 is apt-update not being run, but it also doesn't seem like apt-update takes too long right now. In the future it might be a problem, but for now this fixes the pytest workflow and updates it to 24.04.

CyanVoxel:
> @CyanVoxel it appears the only thing stopping pytest from working on ubuntu 24.04 is apt-update not being run, but it also doesn't seem like apt-update takes too long right now. In the future it might be a problem, but for now this fixes the pytest workflow and updates it to 24.04.

Thank you for the fix! We can monitor the run times as we go - a slow workflow is better than a broken workflow...

PR #641, opened 2024-12-13
 fix: preview panel aliases not staying up to date with database

DandyDev01:
When changes were made to a tag using the tagDatabasePanel, those changes would not be reflected in the previewPanel.

When changes where made to a tag using the previewPanel, those changes would not be reflected in the database.


CyanVoxel (approved):
Everything seems to be working now, thank you so much for the fix!

PR #646, opened 2024-12-14
fix: Pytest modifies and updates search_library sqlite file

### LINKED ISSUE #644: [Bug]: Pytest modifies and updates search_library sqlite file
### Checklist

### TagStudio Version

Alpha v9.5.0 (main branch)

Windows 10

### Description

Running Pytest (specifically one or more of the search_library tests) results in the test database being modified, resulting in git wanting to commit the changes.
The internal change seems to be a result of the [file change counter](https://www.sqlite.org/fileformat.html) value being updated.

### Expected Behavior

The test database should not be changed on disk, especially when no changes are being made to it. It should also not be included in the .gitignore because the file in question can receive genuine updates and changes.

### Steps to Reproduce

1. Clone and set up a TagStudio development environment per the [CONTRIBUTING.md](https://github.com/TagStudioDev/TagStudio/blob/main/CONTRIBUTING.md) instructions
2. Run Pytest with `pytest tagstudio/tests/`
3. Observe modified file at `\TagStudio\tagstudio\tests\fixtures\search_library\.TagStudio\ts_library.sqlite`

### Logs

_No response_

Computerdores:
Copies the search library to a temporary directory before opening it, therefore avoiding changes to the `.sqlite` file in the repo.
Close #644.

### DIFF
--- a/tagstudio/tests/conftest.py
+++ b/tagstudio/tests/conftest.py
@@ -1,13 +1,16 @@
-import pathlib
 import sys
+from pathlib import Path
 from tempfile import TemporaryDirectory
+from typing import Generator
 from unittest.mock import Mock, patch
 
 import pytest
 
-CWD = pathlib.Path(__file__).parent
+CWD = Path(__file__).parent
 # this needs to be above `src` imports
 sys.path.insert(0, str(CWD.parent))
+
+import shutil
 
 from src.core.library import Entry, Library, Tag
 from src.core.library import alchemy as backend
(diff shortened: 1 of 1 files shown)


Tishj:
Not very familiar with the tests or the importance of `search_library`, so take this with a grain of salt;
Is it not a better practice to open `search_library.sqlite` in read-only mode? explicitly making a temporary copy for the tests that do require making modifications?

That way it should prevent bugs where methods unexpectedly make changes to the search_library, when they are perceived to be read-only instructions

And also prevent tests making changes to the search_library that cause the behavior of other tests to change, depending on the order the tests are ran in? (not sure how your current solution deals with lifetimes like this, is the copied `search_library` cleaned up at the end of the test?)

CyanVoxel:
> Not very familiar with the tests or the importance of `search_library`, so take this with a grain of salt; Is it not a better practice to open `search_library.sqlite` in read-only mode? explicitly making a temporary copy for the tests that do require making modifications?
> 
> That way it should prevent bugs where methods unexpectedly make changes to the search_library, when they are perceived to be read-only instructions
> 
> And also prevent tests making changes to the search_library that cause the behavior of other tests to change, depending on the order the tests are ran in? (not sure how your current solution deals with lifetimes like this, is the copied `search_library` cleaned up at the end of the test?)

This was something I was thinking of as well when initially trying to fix #644. I'm not sure if there's a different way, but trying to add the readonly options `create_engine([...].execution_options(postgresql_readonly=True))` or `create_engine([...], execution_options={"postgresql_readonly": True})` didn't seem to make any difference and still let the databases be written to. Admittedly I'm learning this ORM as we go, so there's a good chance I'm just not using the proper methods to open the database in readonly mode here.

Computerdores:
Closed in favor of #648

PR #648, opened 2024-12-17
fix: stop sqlite db from being updated while running tests

### LINKED ISSUE #644: [Bug]: Pytest modifies and updates search_library sqlite file

python357-1:
This PR fixes #644 by having `make_tables` check the current autoincrement value (what the next inserted row would have as its id) on the `tags` table. Previously, an `INSERT` and a `DELETE` were always being run when a library was opened, causing the file change counter to increase

close #644 
close #646

### DIFF
--- a/tagstudio/src/core/library/alchemy/db.py
+++ b/tagstudio/src/core/library/alchemy/db.py
@@ -2,7 +2,7 @@ from pathlib import Path
 
 import structlog
 from sqlalchemy import Dialect, Engine, String, TypeDecorator, create_engine, text
-from sqlalchemy.orm import DeclarativeBase, sessionmaker
+from sqlalchemy.orm import DeclarativeBase
 
 logger = structlog.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


python357-1:
I don't know how to check if the schema is the same to do this - even the sqldiff tool didn't detect this. The reason we are doing the insert and delete is to make sure the next auto increment ID is at least 1000. This implementation allows it to skip updating the database if the next ID is at least 1000.

Computerdores (approved):
This fixes the issue for me on Win10 and I don't see anything wrong with the code either (no idea whether using `sessionmaker` is better here, but I see nothing wrong with it)

python357-1:
yeah i actually just fixed that, realized there was no need for that to be there lol

CyanVoxel (approved):
Looks like this fixes the modification issues without breaking the auto-increment values 👍

PR #649, opened 2024-12-17
feat: implement search equivalence of "jpg" and "jpeg" filetypes

Computerdores:
Implements feature from the roadmap:
> Optional consolidation of extension synonyms (i.e. ".jpg" can equal ".jpeg")

More equivalencies can easily be added through modifying a constant in `src/core/library/alchemy/visitors.py`.

(No issue because this is a short PR and because I don't think anyone else is touching the search rn)

### DIFF
--- a/tagstudio/src/core/library/alchemy/visitors.py
+++ b/tagstudio/src/core/library/alchemy/visitors.py
@@ -16,6 +16,15 @@ if TYPE_CHECKING:
 else:
     Library = None  # don't import .library because of circular imports
 
+FILETYPE_EQUIVALENTS = [set(["jpg", "jpeg"])]
+
+
+def get_filetype_equivalency_list(item: str) -> list[str] | set[str]:
+    for s in FILETYPE_EQUIVALENTS:
+        if item in s:
+            return s
+    return [item]
+
 
 class SQLBoolExpressionBuilder(BaseVisitor[ColumnExpressionArgument]):
     def __init__(self, lib: Library) -> None:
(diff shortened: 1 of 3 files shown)


tagstudio/src/core/library/alchemy/visitors.py
```diff
@@ -16,6 +16,15 @@
 else:
     Library = None  # don't import .library because of circular imports
 
+FILETYPE_EQUIVALENTS = [set(["jpg", "jpeg"])]
```
CyanVoxel, line 19: For the JPG set I have a couple of suggested additions: `jpeg_large`/`jpg_large` and `jfif`. As far as I *know*, `jfif` should be the same as `jpeg`.

It might also be easier going forward to break out these sets into something like the `media_types.py` so they'll be easier to find, modify, and use elsewhere if needed.

CyanVoxel (commented):
Seems to work great! And thank you for updating the roadmap, too!
Approved when suggestions are addressed 👍

tagstudio/src/core/library/alchemy/visitors.py
Computerdores: What do you think about adding this to `core/constants.py`? Since this is not specifically SQL related it should go into core anyways so it would fit

CyanVoxel: `core/media_types.py` is not directly related to SQL and in fact predates the SQL migration - It's just a general core library for all file type / extension / media type information.

At first I considered `core/constants.py` but thought that `core/media_types.py` would be a better fit since it's also the file that handles all related media type "constants" and definitions.

Computerdores: I don't know how I missed that `media_types.py` already existed lol
Should be fixed now!
(This was a 3AM commit though, maybe read it over :) )

CyanVoxel (approved):
Thank you again for your work on this!

PR #651, opened 2024-12-21
Fix: [bug] Enter/Return not adding top result tag

### LINKED ISSUE #645: [Bug]: Pressing Enter/Return no longer adds searched tag to entry
### Checklist

### TagStudio Version

Alpha v9.5.0 (main branch)

Window 10

### Description

When in a tag search panel, pressing the enter/return key no longer adds that tag to an entry before closing the panel.

### Expected Behavior

In previous versions, the top tag result in a tag search panel would be added to an entry upon pressing the enter/return key.

### Steps to Reproduce

1. On an entry with a tag field, press the "+" button to open the tag search panel
2. Type in the full or partial name of a tag to add
3. Press the enter/return key
4. Observe how the panel closes, yet does not add the tag as it would in previous versions (guaranteed 9.4.x)

### Logs

_No response_

SkeleyM:
fixes bug mentioned in #645

close #645


CyanVoxel (approved):

PR #653, opened 2024-12-21
Fix: [Bug] "Fix Unlinked Entries" not functioning, only returns None count

### LINKED ISSUE #652: [Bug]: "Fix Unlinked Entries" not functioning, only returns `None` count
### Checklist

### TagStudio Version

`main` branch (v9.5.0)

macOS 15.1.1

### Description

Using the "Fix Unlined Entries" tool only ever returns "None" as the count value, even if there are unlinked entries present in the library.

This seems to stem from [this](https://github.com/TagStudioDev/TagStudio/blob/934c98a4d486e36257054022c44b3901eae12596/tagstudio/src/qt/modals/fix_unlinked.py#L129) change made to `fix_unlinked.py` in #528. From what I can tell it seems that `set_missing_count()` is being passed as a callable function to `from_iterable_function` and is being called when the `done` signal inside its CustomRunnable `r` is emitted - however *without* any count argument passing to `set_missing_count()`, resulting in the method's default behavior of setting the count to `None`.

### Expected Behavior

Refreshing the unlinked entries method should return the count of the unlinked entries, and a 0 if none are found. Note that the "None" text currently being shown is not a result of there being no unlinked entries, but rather is an errant `None` value being displayed as the count.

### Steps to Reproduce

1. Open a TagStudio library
2. Move or delete a present in the library on disk to create an "unlinked entry"
3. Go to "Tools -> Fix Unlinked Entries" and click "Refresh All"
4. Note how the count now always says "Unlinked Entries: None"

5. Bonus: Repeat the previous steps while omitting step 2, noting the same result

### Logs

_No response_

SkeleyM:
fixes use of count instead of self.missing_count

close #652

### DIFF
--- a/tagstudio/src/qt/modals/fix_unlinked.py
+++ b/tagstudio/src/qt/modals/fix_unlinked.py
@@ -144,4 +144,4 @@ class FixUnlinkedEntriesModal(QWidget):
             # disable buttons if there are no files to fix
             self.search_button.setDisabled(self.missing_count == 0)
             self.delete_button.setDisabled(self.missing_count == 0)
-            self.missing_count_label.setText(f"Unlinked Entries: {count}")
+            self.missing_count_label.setText(f"Unlinked Entries: {self.missing_count}")


Computerdores:
This does fix the bug outlined in #652, however relinking is still broken even with this fix. Would you be interested in fixing that as well in this PR?

SkeleyM:
@Computerdores 
so pressing "Refresh All" is supposed to relink an entry if it is replaced in its original location?
if im not mistaken it's an easy fix.

Refresh all seems to relink the entry just fine just not refreshing the thumbnail. id be happy to fix it.

Computerdores:
> @Computerdores so pressing "Refresh All" is supposed to relink an entry if it is replaced in its original location? if im not mistaken it's an easy fix.

"Refresh All" is supposed to search for unlinked entries and is working as it is supposed to now afaict. "Search & Relink" is supposed to also relink them, but this is currently broken for me. So if I for example rename a file "Refresh All" now correctly reports 1 unlinked entry, however pressing "Search & Relink" doesn't relink the entry to the renamed file.

Since #652 was only reporting about "Refresh All" being broken, fixing this could be considered out of scope, however it could also make sense do fix that here as well, hence my question of whether you would like to work on that here as well.

Hope that clears up any confusion!

SkeleyM:
thanks for clearing that up, i'm sorry but working on that is something i'd prefer not to do

Computerdores (approved):

CyanVoxel:
Thank you so much for the fix! Now that this reveals a deeper issue with the relinking functionality, I'll open up a separate issue for that.

PR #656, opened 2024-12-22
fix(#655): search now works again

Computerdores:
fixes the search for #655


CyanVoxel (approved):
Thank you so much for this! 👍

Issue #664, opened 2024-12-23
[Bug]: Ui: duplicate subtag parents in tag editor


SkeleyM:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Alpha v9.5.0 (main branch)

### Operating System & Version

Windows 10 22H2

### Description

In the tag editor, when creating or editing an existing tag's parent tags, pressing the '+' button multiple times on the same tag results in duplicate tags being added to the parent tags.

this is only a ui issue, saving the tag and reopening it in the editor shows only 1 tag
![image](https://github.com/user-attachments/assets/2aa00bd7-e340-4b30-9fb7-6a38d85608fa)

### Expected Behavior

Only 1 unique tag should be in the parent tags

### Steps to Reproduce

1.  create a new tag
2.  edit parent tags
3.  press the plus button multiple times on the same tag

### Logs

2024-12-23 12:30:07 [info     ] add_subtag_callback            tag_id=1002
2024-12-23 12:30:08 [info     ] add_subtag_callback            tag_id=1002
2024-12-23 12:30:08 [info     ] add_subtag_callback            tag_id=1002

SkeleyM:
i have already patched this in my local branch by replacing the list with a set, ill submit a pull request in a minute

PR #665, opened 2024-12-23
[Bug] Ui: Stop duplicate tag parents

SkeleyM:
fix for issue #664 

replaces subtag_ids list with a set to prevent duplicate tag parents


CyanVoxel (approved):

Issue #350, opened 2024-08-14
[Bug]: Can't open an image if its path or name includes non-latin symbols


Aosneg:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Alpha 9.3.2

### Operating System & Version

Windows 11

### Description

Tried with both Russian and Japanese letters in path and names, but only English letters work for images. At the same time, opening txt files works, but encoding is broken for preview images of txt that contains non-latin letters.

### Expected Behavior

I'm clicking on image and it opens.

### Steps to Reproduce

Here is video demonstration, if needed: https://gofile.io/d/TICXz2

python357-1:
I apologize that nobody has responded to this so far. Would you be willing to re-upload the video?

SkeleyM:
i did some investigating on this, im not sure if this is just me, it seems to open photos.exe as a process but it never creates a window and runs in the background indefinitely (persisting even after tag studio is closed)

![image](https://github.com/user-attachments/assets/5c28266e-e369-4dea-b260-4d44890ed41b)


this also happens with some other characters too, such as #, +, £ (likely more)

Aosneg:
> i did some investigating on this, im not sure if this is just me, it seems to open photos.exe as a process but it never creates a window and runs in the background indefinitely (persisting even after tag studio is closed)
> 
[...]

PR #667, opened 2024-12-24
[Bug] Allow files with non latin paths to be opened from preview

### LINKED ISSUE #666: [Bug]: Opening file with an '&' fails
### Checklist

### TagStudio Version

Alpha v9.5.0 (main branch)

Windows 10 22H2

### Description

Clicking on the file preview to open the file fails when the path contains an ampersand ('&')

The '&' gets interpreted as a shell command
image

### Expected Behavior

The file should be opened regardless of if the path contains an & or not

### Steps to Reproduce

1. Rename a file to contain an '&' or have one in a folder with an '&' in the name
2. Click the preview of the file in tag studio

### Logs

2024-12-23 14:44:47 [info     ] Opening file                   path=WindowsPath('C:/Users/user/Pictures/path&path/waw.PNG')
The system cannot find the file C:\Users\user\Pictures\path.

SkeleyM:
this change fixes #666 and #350.

entering files full path into cmd opens it in its associated program, and using that instead of "start" fixes the issues.

im assuming it has something to do with "start" being almost 30 years old at this point but this seems to fix it for me

### DIFF
--- a/tagstudio/src/qt/helpers/file_opener.py
+++ b/tagstudio/src/qt/helpers/file_opener.py
@@ -46,11 +46,9 @@ def open_file(path: str | Path, file_manager: bool = False):
                     | subprocess.CREATE_BREAKAWAY_FROM_JOB,
                 )
             else:
-                command_name = "start"
-                # first parameter is for title, NOT filepath
-                command_args = ["", normpath]
+                command = f'"{normpath}"'
                 subprocess.Popen(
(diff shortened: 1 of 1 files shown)


CyanVoxel:
This works for me, however I was never able to reproduce any of the original issues on my system (Windows 10) so I'd appreciate it if someone else experiencing these issues could help verify this fix.

PR #672, opened 2024-12-24
Fix: [Bug] Accent colors show as Gray on windows

### LINKED ISSUE #668: [Bug]: Accent Colors show as Gray on Windows with PySide 6.8.0.1
### Checklist

### TagStudio Version

`main` branch (v9.5.0)

Windows 10 22H2

### Description

System accent colors grabbed by QPalette show as gray on Windows as of updating to PySide 6.8.0.1. The following snippet is from `thumb_button.py`:
```py
self.select_color: QColor = QPalette.color(
            self.palette(),
            QPalette.ColorGroup.Active,
            QPalette.ColorRole.Accent,
        )
```

PySide 6.7.1 (aaea0b1475206edf093f32191359a5f0aa0d6187)
image

PySide 6.8.0.1 (latest)
image

Note that I've only seen this occur on Windows while it seems to be functioning on macOS. Other native Qt widgets such as loading bars also seem to be grabbing the accent color correctly.

### Expected Behavior

Accent/selection colors on all platforms should function as they used to before PySide 6.8.0.1.

### Steps to Reproduce

1. Open a TagStudio library on Windows (confirmed on 10, not sure about 11)
2. Hover over an item thumbnail and observe the incorrect accent color

### Logs

_No response_

SkeleyM:
I'm not too sure why the behaviour changed between versions of PySide but this seems to fix it on the newest version.
the colour limit was set too high so if you had a bright accent colour it would become closer to white/gray due to being too bright.

i cant speak for how this change looks on other platforms though so it should be tested there but on windows at least it brings back the colour and still creates an obvious select effect

close #668

### DIFF
--- a/tagstudio/src/qt/widgets/thumb_button.py
+++ b/tagstudio/src/qt/widgets/thumb_button.py
@@ -47,7 +47,7 @@ class ThumbButton(QPushButtonWrapper):
         self.hover_color.setHsl(
             self.hover_color.hslHue(),
             self.hover_color.hslSaturation(),
-            min(self.hover_color.lightness() + 80, 255),
+            min(self.hover_color.lightness() + 80, 200),
             self.hover_color.alpha(),
         )
 


CyanVoxel:
Unfortunately this doesn't resolve it on my machine, in fact skipping the HSL steps entirely still doesn't get me the original accent color and is still always some middle shade of gray...

SkeleyM:
Weird, i just tried with the same colour used in the issue and got the same result.

this issue only seems to be effecting really light colours since darker colours work just fine, ill try do some more looking around

SkeleyM:
I'm going to close this as i havent been able to make any progress with this issue, ill likely ask on the qt forums

PR #673, opened 2024-12-28
feat: implement parent tag search

### LINKED ISSUE #661: [Parity/Feature Request]: Parent Tag Searching on SQL
### Checklist

### Description

Request to implement v9.4 style parent tag searching on the current v9.5 SQL search system first introduced in #606.

### Solution

Allow tags to be substituted in searches by their child tags, e.g. any tags that mark them as a parent tag.

Example:
- Tag "Mario" as the tag "Character" as a parent tag.
- Searching for the tag "Mario" should bring up all entries tagged with "Mario" (currently operational)
- Searching for the tag "Character" should *also* bring up all entries tagged with "Mario" (v9.4 behavior, currently not operational in v9.5)

### Alternatives

N/A

Computerdores:
Searching for a tag now includes entries that are only tagged with child tags of the tag. A search for `Nintendo` would therefore also include entries that only have the tag `Mario` or `Luigi` (assuming that `Nintendo` is a parent tag of `Mario` and `Luigi`).

Close #661.


tagstudio/src/core/library/alchemy/visitors.py
```diff
@@ -98,16 +113,28 @@ def visit_property(self, node: Property) -> None:
...
+            tag_ids = list(
                 session.scalars(
                     select(Tag.id)
                     .where(or_(Tag.name.ilike(tag_name), Tag.shorthand.ilike(tag_name)))
                     .union(select(TagAlias.tag_id).where(TagAlias.name.ilike(tag_name)))
                 )
             )
+            if len(tag_ids) > 1:
+                logger.debug(
+                    f'Tag Constraint "{tag_name}" is ambiguos, {len(tag_ids)} matching tags found',
```
CyanVoxel, line 128:
```suggestion
                    f'Tag Constraint "{tag_name}" is ambiguous, {len(tag_ids)} matching tags found',
```

CyanVoxel (commented):
Seems to function as it's supposed to! The query times can get quite high for some larger libraries and tag sets (16+ seconds for me), but I suppose that's the nature of a non-indexed recursive search like this?

python357-1 (approved):
I think this is good for merging but we should probably take a look at the performance when we can, preferably before release. Cyan is probably at the higher end in terms of tags in a library but if a user gets search times near 16 seconds, they will probably think something is wrong.

Issue #393, opened 2024-08-25
[Bug]: Unable to run on fresh install Ubuntu 24.04 using X11


faellacurcio:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Alpha v9.3.2

### Operating System & Version

Ubuntu 24.04

### Description

I have found that the latest release of TagStudio won't open on my freshly installed (one month) Ubuntu 24.04 operating system.
When I try to open I get the following:

```
❯ ./tagstudio
[ResourceManager] 4 resources registered
[QT DRIVER] Config File not specified, defaulting to /home/rafael/.config/TagStudio/TagStudio.ini
qt.qpa.plugin: From 6.5.0, xcb-cursor0 or libxcb-cursor0 is needed to load the Qt xcb platform plugin.
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "/home/rafael/Downloads/tagstudio_linux_x86_64/tagstudio/_internal/cv2/qt/plugins" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: xcb, wayland, minimalegl, eglfs, vnc, offscreen, wayland-egl, vkkhrdisplay, linuxfb, minimal.

[1]    10938 IOT instruction (core dumped)  ./tagstudio
```

After googling a bit and tweaking around, I found that if I do login using "Ubuntu Wayland".

### Expected Behavior

Open the program normally.

### Steps to Reproduce

1. Install Ubuntu 24.04
2. Update/upgrade
3. Download release from TagStudio
4. Open Terminal and run ./tagstudio

### Logs

❯ ./tagstudio
[ResourceManager] 4 resources registered
[QT DRIVER] Config File not specified, defaulting to /home/rafael/.config/TagStudio/TagStudio.ini
qt.qpa.plugin: From 6.5.0, xcb-cursor0 or libxcb-cursor0 is needed to load the Qt xcb platform plugin.
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "/home/rafael/Downloads/tagstudio_linux_x86_64/tagstudio/_internal/cv2/qt/plugins" even though it was found.
[...]

PR #689, opened 2025-01-06
docs: add a linux install notice

Qronikarz:
Adds a notice about possible missing packages on Linux:
https://github.com/TagStudioDev/TagStudio/discussions/182
https://github.com/TagStudioDev/TagStudio/issues/393


Computerdores (approved):
should probably be squashed together with #690 though

PR #618, opened 2024-11-30
feat: Improve performance of Delete Missing Entries function

Toby222:
Fixes #617 by deleting all missing entries simultaneously instead of one at a time.

PR #696, opened 2025-01-11
feat: Improve performance of Delete Missing Entries function

### LINKED ISSUE #617: [Feature Request]: Use sql to delete all unlinked entries simultaneously
### Checklist

### Description

As per [this TODO](https://github.com/TagStudioDev/TagStudio/blob/b7e652ad8d8055cd49bf37138413568d058e8a7b/tagstudio/src/core/utils/missing_files.py#L65):
Currently when deleting unlinked entries, they are deleted one-by-one, only so they can update the progress bar.
I believe it would be better to let the progress bar be indeterminate and speed up the deletion by deleting all items in parallel.
In brief testing the speedup was significant enough that the progress bar was no longer visible long enough for the lack of visible progress to matter :^)

### Solution

Instead of calling `remove_entries` with one id at a time, pass all relevant IDs simultaneously.

### Alternatives

_No response_

Computerdores:
Continuation of #618. Improves speed of deletion from 3min for 550 unlinked entries to 4s for 250k unlinked entries.
Fix #617.
Close #618.


tagstudio/src/qt/modals/delete_unlinked.py
```diff
@@ -95,8 +96,17 @@ def displayed_text(x):
         pw = ProgressWidget(
             cancel_button_text=None,
             minimum=0,
-            maximum=self.tracker.missing_files_count,
+            maximum=0,
         )
```
CyanVoxel, line 100: Seems to work great! My only nitpick (and this comes from #618) is that after removing the iterator the final label text `pw.update_label(f"Deleting {idx}/{files_count} Unlinked Entries")` was removed as well. This causes the window to show a text-less progress bar, even if it is a pretty brief operation. I would suggest using the same translation used for the window title and applying that to the pw label as well.

Other than that, this is approved! Thank you both for this!

CyanVoxel (approved):

tagstudio/src/qt/modals/delete_unlinked.py
Computerdores: should be fixed!

PR #647, opened 2024-12-15
feat: begin adding settings menu

python357-1:
This PR addresses the "Settings Menu" item in the tagstudio roadmap

PR #703, opened 2025-01-15
feat(ui): add configurable splash screens

CyanVoxel:
### Summary
This PR introduces a new configurable splash screen system which allows for the selection of multiple preset splash screens, as well as the ability to set up future ones in code without needing to bake in information such as copyright notices or version numbers.

As part of this PR I've reconfigured the the existing splash screen to work with this system, added a remastered version of the original splash screen, and finally added a configuration for an upcoming v9.5 splash screen.

image

***

**NOTE:** This PR is currently awaiting the merging of #647 in order to add the splash screen selection to the new settings config.


Computerdores (commented):
Source code looks good just two notes from testing:
- SPLASH_95 seems broken
- on SPLASH_GOO the cyan text is partially on cyan background and therefore hard to read

CyanVoxel:
> Source code looks good just two notes from testing:
> 
> * SPLASH_95 seems broken

I've omitted the image for `SPLASH_95` in this PR and plan on committing it at a later date, so presently it should fall back to a black image

> * on SPLASH_GOO the cyan text is partially on cyan background and therefore hard to read

There shouldn't be any cyan text or overlapping text on visually disruptive parts of the image, could you send a screenshot of what this looks like for you? The reference of how this splash should appear is in the PR description

Computerdores:
Had the colors mixed up, but this is what it looks like:
image

CyanVoxel:
> Had the colors mixed up, but this is what it looks like:

Thanks for the image, looks like I need to make some adjustments so that everything shows up the same (or similar enough) on different machines.

If you could, could you let me know the OS you tested this on, your monitor resolution, and your display scaling percentage (probably just 100% / 1.0 unless you're on a high resolution display)?

Computerdores:
Win 10 22H2
1920x1080
100%

CyanVoxel:
It seems like it's directly related to the different default fonts Qt uses on different systems. Testing on Windows with different screen scaling settings all got me the same proportions as your screenshot.

I added some code that changes the point size of the font if it's detected to be Segoe UI so that it roughly matches the size used when originally designing this on macOS. There's probably going to be one or more extra cases that arise on Linux, but I may just need to fix those as case(s) come in since I'm unable to develop on that myself at the moment.

Computerdores (approved):

CyanVoxel:
Thank you @Computerdores for the review!

PR #707, opened 2025-01-16
fix: restore environment before launching external programs

mashed5894:
### Issue
By default, any subprocesses will inherit environment variables from its parent. Pyinstaller will specify a custom ```LD_LIBRARY_PATH```, which causes issues when the subprocess attempts to load its own dynamically linked libraries.
https://web.archive.org/web/20240824170055/https://github.com/TagStudioDev/TagStudio/issues/210

### Solution
1. Implemented suggested solutions for linux and windows from https://pyinstaller.org/en/stable/common-issues-and-pitfalls.html#launching-external-programs-from-the-frozen-application. Doesn't really seem like a solution for MacOS is necessary based on the docs.

2. Renamed and slightly reworked promptless_Popen into silent_Popen (better name, and congruent with the actual file name) so future features can take advantage of the same solution.

Tested on linux and briefly on windows (but could not figure out a way of getting hold of an outdated dll). At least it shouldn't be worse.

### DIFF
--- a/tagstudio/src/qt/helpers/file_opener.py
+++ b/tagstudio/src/qt/helpers/file_opener.py
@@ -12,6 +12,8 @@ import structlog
 from PySide6.QtCore import Qt
 from PySide6.QtWidgets import QLabel
 
+from src.qt.helpers.silent_popen import silent_Popen 
+
 logger = structlog.get_logger(__name__)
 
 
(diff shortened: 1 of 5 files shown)


Computerdores (approved):
Don't really have a way to test this, but it seems to follow the documentation well and I see no other issues.

CyanVoxel:
PyInstaller builds tested and working on Windows 10 22H2 (Normal + Portable) and macOS 15 👍

simjeehza:
PyInstaller build tested and working on Ubuntu 24.04.1 👍

CyanVoxel:
Thank you so much for this fix! And thank you to everyone who helped test! 🎉

PR #710, opened 2025-01-17
Fix: [Bug] Can't Drag Out Multiple Files

### LINKED ISSUE #706: [Bug]: Can't Drag Out Multiple Files
### Checklist

### TagStudio Version

Alpha 9.5.0 (main branch)

macOS 15.1.1

### Description

When selecting multiple files to drag outside of the program, only the last file clicked is dragged out since the selection updates when the mouse is pressed down rather than when released. This seems to be a direct result of the changes proposed in #555 and implemented in #556.

While "select on down" still seems to be the default behavior of most if not all visual file explorers, there is additional logic needed to determine when a drag is taking place that TagStudio lacks as of the change made in #556.

### Expected Behavior

When multiple files are selected, clicking and dragging the selection to outside of the program should "share" all of the files in the same way that this currently works for single files.

### Steps to Reproduce

1. Select more than one file in a TagStudio library
2. Click and drag the group of selected files outside of the program (i.e. to another app, a system folder, etc.)
3. Observe how the selection updates to item that was clicked, and how that's now the only file being dragged

### Logs

_No response_

SkeleyM:
Fixes #706 

Let me know if this is alright

### DIFF
--- a/tagstudio/src/qt/widgets/item_thumb.py
+++ b/tagstudio/src/qt/widgets/item_thumb.py
@@ -441,7 +441,9 @@ class ItemThumb(FlowWidget):
         if clickable:
             with catch_warnings(record=True):
                 self.thumb_button.pressed.disconnect()
-            self.thumb_button.pressed.connect(clickable)
+            self.thumb_button.pressed.connect(
+                lambda: clickable() if not self.thumb_button.selected else None
+            )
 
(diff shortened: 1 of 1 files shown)


CyanVoxel (approved):
This seems to fix it! Tested on macOS 15 and Windows 10.

Issue #687, opened 2025-01-05
[Bug]: Regression: 'Add Tags' search result is no longer ordered by exactness


Tishj:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Alpha 9.5

### Operating System & Version

Windows 10

### Description

Previously, when searching for a tag, results would be ordered (probably as a fortunate coincidence) by how exact the match is

If we for example take these words:
Hand
Handlebar
Handle

Previously a search for `h` up to `hand` would get `Hand`, `handl` and `handle` result in `Handle` and `handleb` up to `handlebar` would result in `Handlebar`

Now it seems to be random, or based on the order the tags were created in, a search for `hand` could have `Handlebar` as the first result.
The order in which the results are given is relevant when 'enter' is used to select the tag, which I would think is the preferred way of adding tags

CyanVoxel:
I've also noticed this, thank you for opening up an issue for it!

The current tag search order present in v9.4 came from #329, which doesn't appear to have been ported over to the `main` branch yet. That one also leveraged tag display names when it could, which also still aren't currently on `main` yet.

PR #721, opened 2025-01-23
fix: sort tag results

mashed5894:
fix for #687. added logic to sort the results of a tag search. currently prioritizing results that start with the correct string and sorting them by length, and then sorting all other results alphabetically. saw the todo comment, but i could not figure out a simple way of reusing the widgets, theyd probably have to be stored in a dict between updates or something.


CyanVoxel (approved):
Fantastic work on this, thank you so much for the fix!

Issue #660, opened 2024-12-22
[Parity]: Port Tag/Field Copy and Pasting from v9.4 to v9.5


CyanVoxel:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

Port the copy and pasting functionality of tags and fields (#79) from v9.4 over to the main branch (v9.5)

### Solution

Port the features of #79 to the main branch, while accounting for differences in the database schema - most notably those introduced in #655. It would also be nice to finally have the hover icons for copying (similar to the "pencil" edit icon) function as a way to copy tag and field data with a single left-click rather than always needing to use the right-click context menu.

### Alternatives

Rebuild the copy pasting system, not being influenced by #79. While there's been some significant changes, #79 at least provides a skeleton of the features needed.

mashed5894:
Working on this one!

PR #722, opened 2025-01-23
feat: copy/paste fields and tags

mashed5894:
implementing #660 using stuff from #79. copying and pasting fields and tags with ctrl+c & ctrl+v from one or several entries (copying only works with a single entry). i got the preview to update properly, but im having some issues updating the badge when pasting to several entries at once. the system for doing so seems pretty convoluted and might need some refactoring to implement properly.


tagstudio/resources/translations/en.json
```diff
@@ -182,6 +182,8 @@
     "preview.no_selection": "No Items Selected",
     "select.all": "Select All",
     "select.clear": "Clear Selection",
+    "copypaste.copy_fields": "Copy Fields",
+    "copypaste.paste_fields": "Paste Fields",
```
CyanVoxel, line 186: I know this is still a draft, I just wanted to drive-by nitpick and suggest renaming the term`copypaste` to `edit` for clarity and consistency

CyanVoxel:
There seems to be an issue where pasting a group of tags to an entry which already contains one or more of the tags stops the rest of the tags from pasting

tagstudio/src/qt/ts_qt.py
```diff
@@ -1077,11 +1128,22 @@ def toggle_item_selection(self, item_id: int, append: bool, bridge: bool):
                 it.thumb_button.set_selected(False)
 
         self.set_macro_menu_viability()
+        self.set_copypaste_menu_viability()
         self.preview_panel.update_widgets()
 
     def set_macro_menu_viability(self):
         self.autofill_action.setDisabled(not self.selected)
 
+    def set_copypaste_menu_viability(self):
```
CyanVoxel, line 1137: Similar to the change in the translation keys, I suggest renaming `copypaste` here to something like `clipboard`

tagstudio/src/qt/ts_qt.py
```diff
@@ -1007,6 +1037,27 @@ def _init_thumb_grid(self):
...
+            entry = self.lib.get_entry_full(self.selected[0])
+            if entry:
+                self.copy_buffer["fields"] = entry.fields
+                self.copy_buffer["tags"] = [tag.id for tag in entry.tags]
+        self.set_copypaste_menu_viability()
+
+    def paste_fields_action_callback(self):
+        for id in self.selected:
+            for field in self.copy_buffer["fields"]:
+                self.lib.add_field_to_entry(id, field_id=field.type_key, value=field.value)
```
CyanVoxel, line 1051: > is it intended that mirror_entry_fields doesnt add a second field if the entry already has a field with the same key, and also doesnt mirror tags?

When answering you in the discord I missed one oversight about how the 9.4 version worked (if not the 9.0 version, which predates the initial commit on GitHub) - if a field has the same type key but a different *value* then it should be pasted. If the field has both the same type key *and* the same value, then I feel it's better for it to *not* be pasted, as duplicate fields with the same content most likely isn't ever the desired outcome.

CyanVoxel (commented):
Approved once all comments are addressed, and the database insert fix discussed in the discord is implemented 👍 

tagstudio/src/core/library/alchemy/library.py
```diff
@@ -928,19 +929,18 @@ def add_tag(
...
-                    session.add(TagEntry(tag_id=tag_id, entry_id=entry_id))
-                    session.flush()
+                session.execute(stmt)
                 session.commit()
-                return True
             except IntegrityError as e:
                 logger.warning("[add_tags_to_entry]", warning=e)
                 session.rollback()
                 return False
+            return True
```
CyanVoxel, line 943: If these changes causing the failing migration tests are reverted, then I'd be happy to pull this!

CyanVoxel (changes requested):

CyanVoxel (approved):
Thank you for all your work on this!

PR #735, opened 2025-01-26
fix: replace calls to `translate_qobject` with `translate_formatted` inside `TagWidget` (Fix #733)

### LINKED ISSUE #733: [Bug]: Adding (or searching for) tags gets increasingly slower in the same session (running process)
### Checklist

### TagStudio Version

main

Windows 10

### Description

Updating a library, adding tags to new entries gets increasingly slower over time.

Some details in hopes they help:
Library has around 300~ tags, 1200~ existing entries
Slowness becomes pretty much unbearable (tag search takes >5 seconds) after like 5-6 new entries, which each receive maybe 8~ tags.
Closing the library (keeping the application open) and reopening the library does not help.
Even clearing the search_field input, causing it to search for *all tags* takes seconds after use, whereas this is near instant on a fresh start of the process.
Closing the process and reopening, then performing the exact same search is 10x faster

### Attempted fixes / debugging information

I've noticed search gets run for every keypress in the search_field, which I've made a local change for using QTimer as a debounce timer but that didn't make noticeable improvements (I can possibly PR this if that seems interesting to you).
Essentially it waits 225ms after a keypress to perform the search, getting canceled by every subsequent keypress, so hopefully when the name is already known by the user, we can perform a single tag search, at the end.

I suspect there's a memory leak somewhere, or an infinitely growing list that is making every subsequent iteration take increasingly longer, but I'm not  familiar enough with the project to find this.

Also the logs didn't seem to provide much useful information to debug this.

### Reproduction

I made a reproduction, all tags are obfuscated (replaced with numbers) because the data is personal.
All entries are replaced with empty `.txt` files

https://drive.google.com/file/d/1QZ86CO6JzLVMQlm8kZRk9675_-l2q98o/view?usp=sharing

Note: I think the size of the tags or the size of the search matters in reproducing the issue, the real tags are about 8-12 characters wide.

1. Open the library (I noticed the `folders` contains an absolute path, so I'm not sure the repro is sharable without some manual editing ?)
2. Add these tags to each entry separately:
```
12
23
34
45
56
78
89
90
```
*using the keyboard, pressing enter inbetween, no cursor is involved other than pressing "Add Tags"
3. By the 8th~ entry you will notice it takes increasingly longer to complete all these steps for each subsequent entry

CyanVoxel:
This PR aims to fix #733 by replacing calls to `translate_qobject` with `translate_formatted` inside the `TagWidget` class, as for an unknown reason this seems to cause the application to progressively slow down as more TagWidgets are created and removed over time, and also seemingly causes a memory leak.

The `translate_qobject` issue should be investigated further, however for now this fix should hopefully address the specific issue raised in #733. I've also included a couple small fixes and tweaks to nearby code that I discovered while tracking this down.

### DIFF
--- a/tagstudio/src/qt/modals/tag_search.py
+++ b/tagstudio/src/qt/modals/tag_search.py
@@ -52,7 +52,7 @@ class TagSearchPanel(PanelWidget):
         self.search_field.setObjectName("searchField")
         self.search_field.setMinimumSize(QSize(0, 32))
         Translations.translate_with_setter(self.search_field.setPlaceholderText, "home.search_tags")
-        self.search_field.textEdited.connect(lambda: self.update_tags(self.search_field.text()))
+        self.search_field.textEdited.connect(lambda text: self.update_tags(text))
         self.search_field.returnPressed.connect(lambda: self.on_return(self.search_field.text()))
 
         self.scroll_contents = QWidget()
@@ -119,7 +119,7 @@ class TagSearchPanel(PanelWidget):
             row.addWidget(add_button)
         return container
 
-    def construct_tag_button(self, query: str | None):
+    def build_create_tag_button(self, query: str | None):
         """Constructs a Create Tag Button."""
         container = QWidget()
         row = QHBoxLayout(container)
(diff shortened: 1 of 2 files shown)


Tishj:
This indeed fixes #733 ! 🥳

Computerdores:
it also breaks being able to change the language without restarting...

CyanVoxel:
> it also breaks being able to change the language without restarting...

That will need to be a separate issue then, since the resource leak caused by `translate_qobject` here was severely degrading the usability of the program

PR #741, opened 2025-01-28
feat: show audio/video file lengths

SkeleyM:
this pr gets and sets the file lengths for video and audio, upon hovering over a thumbnail, this can work as a temporary fix until I can get it to work without a performance hit (thanks @Nginearing for the idea)

image
image

this doesnt block the main thread and i havent noticed any performance hit (my audio libraries arent huge, 30 at best).

### DIFF
--- a/tagstudio/src/qt/widgets/item_thumb.py
+++ b/tagstudio/src/qt/widgets/item_thumb.py
@@ -10,8 +10,9 @@ from typing import TYPE_CHECKING
 from warnings import catch_warnings
 
 import structlog
+import subprocess
 from PIL import Image, ImageQt
-from PySide6.QtCore import QEvent, QMimeData, QSize, Qt, QUrl
+from PySide6.QtCore import QEvent, QMimeData, QSize, QThreadPool, Qt, QUrl
 from PySide6.QtGui import QAction, QDrag, QEnterEvent, QPixmap
 from PySide6.QtWidgets import (
     QBoxLayout,
@@ -33,6 +34,8 @@ from src.qt.platform_strings import open_file_str, trash_term
 from src.qt.translations import Translations
 from src.qt.widgets.thumb_button import ThumbButton
 from src.qt.widgets.thumb_renderer import ThumbRenderer
+from src.qt.helpers.vendored.ffmpeg import FFPROBE_CMD
+from src.qt.helpers.custom_runnable import CustomRunnable
 
 if TYPE_CHECKING:
     from src.qt.ts_qt import QtDriver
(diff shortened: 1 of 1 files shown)


CyanVoxel:
Could you update this from main? This will be easier to review and benchmark with the thumbnail caching system and newest internal library version (DB_VERSION 5)

CyanVoxel:
While this does provide duration labels for video and audio files, unfortunately in my testing it indeed appears to lock up the program for several seconds during each operation. Loading pages with video and audio files that would otherwise be instant now takes up to 40 seconds to become responsive.

SkeleyM:
i think the slowdowns are due to my approach. of course reading every single file for each page is extremely slow.

maybe only doing that process once could be beneficial i.e. caching the timestamps in an external file (of course then changes wont be reflected) or having a background thread update the timestamps as it goes to avoid unresponsiveness. or maybe some combination of the two. 
i'm going to change this to a draft since this wont be done overnight

Nginearing:
> i think the slowdowns are due to my approach. of course reading every single file for each page is extremely slow.
> 
> maybe only doing that process once could be beneficial i.e. caching the timestamps in an external file (of course then changes wont be reflected) or having a background thread update the timestamps as it goes to avoid unresponsiveness. or maybe some combination of the two. i'm going to change this to a draft since this wont be done overnight

Perhaps, you could only show the length on hover, or if it is being used to find a video or audio file of a certain length, maybe it is better suited to be in the form of a search query.

SkeleyM:
showing length on hover could work as a temporary way of having the lengths be previewed, so long as it doesnt increase time to interact. might try it and see if it could be suitable

CyanVoxel:
The duration on hover isn't a bad holdover solution, but I'm thinking it may be worth just waiting until there's a longer-term solution. I've had it in mind to eventually cache information like the video duration inside the database but haven't settled on schema to do so yet. I hastily added the `date_added` and `date_modified` file attribute columns to the Entry table in anticipation of caching those, but my thought now is to create a dedicated table for this sort of information. This would also let us feel free to add several new columns for things like the word count or line count of text files without worrying about bloating the Entry table. Or maybe that was only *my* worry... 😅 

Assuming I decide on and put in a table with a column to cache the video and/or audio duration, I think taking a similar approach to how the preview panel refactor works where we can return certain stats when rendering a thumbnail would be beneficial for speeding up the initial generation of this data. For example in the preview panel I have image dimensions and video durations returned in a "stats" dict that gets used by other elements in the preview panel without the need to expensively re-access the file.

Let me know what you think of this. I'll workshop the the cached data table schema some more. I've already got an idea for how detecting updates may work as well (just comparing the date_modified column of the file).

SkeleyM:
I agree creating an exclusive table for information relating to entries would be a very nice and organized solution for things like  this. I think ill close this pr and revisit this in the future when the schema for this stuff is more concrete :)

PR #743, opened 2025-01-28
feat: add smartcase and globless path searches

CyanVoxel:
### Summary
This PR adds "smartcase" style path searches and allows for non-glob searches to still return results. Smartcase searches will act as case-insensitive when the entire query is lowercase, but will turn case-sensitive if one or more capital letters are present in the query. Meanwhile the glob detection is now based on whether or not a query starts or ends with a "*" character.

This PR also limits the number of results in the path autocomplete list to 100 items to avoid sluggish search typing that scales with the size of the library.

### Smartcase Search (New)
`path: temp`
- Returns all paths that have "temp" **(Case insensitive)** somewhere in the name. 

`path: Temp`
- Returns all paths that have "Temp" **(Case sensitive)** somewhere in the name.

#### Glob Patterns (Expanded)
`path: *temp*`
- Returns all paths that have "temp" **(Case insensitive)** somewhere in the name.
 
`path: *Temp*`
- Returns all paths that have "Temp" **(Case sensitive)** somewhere in the name.

`path: temp*`
- Returns all paths that start with "temp" **(Case insensitive)** somewhere in the name. 

`path: Temp*`
- Returns all paths that start with "Temp" **(Case sensitive)** somewhere in the name.

`path: *temp`
- Returns all paths that end with "temp" **(Case insensitive)** somewhere in the name. 

`path: *TEmP`
- Returns all paths that end with "TEmP" **(Case sensitive)** somewhere in the name.

### DIFF
--- a/tagstudio/src/core/library/alchemy/library.py
+++ b/tagstudio/src/core/library/alchemy/library.py
@@ -697,8 +697,6 @@ class Library:
         with Session(self.engine) as session:
             paths = session.scalars(select(Entry.path)).unique()
             path_strings = list(map(lambda x: x.as_posix(), paths))
-
-            path_strings: list[str] = list(map(lambda x: x.as_posix(), paths))
             return path_strings
 
     def search_library(
--- a/tagstudio/src/core/library/alchemy/library.py
+++ b/tagstudio/src/core/library/alchemy/library.py
@@ -692,10 +692,13 @@ class Library:
         with Session(self.engine) as session:
             return session.query(exists().where(Entry.path == path)).scalar()
 
-    def get_paths(self, glob: str | None = None) -> list[str]:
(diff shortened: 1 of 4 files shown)


Tishj:
The glob patterns examples in the description are wrong, no?

I imagine `*temp` would find tags that end in `temp` (insensitive)
`*Temp` is the sensitive variant

`temp*` finds tags that start with `temp` (insensitive)
`Temp*` is the sensitive variant

CyanVoxel:
> The glob patterns examples in the description are wrong, no?
> 
> I imagine `*temp` would find tags that end in `temp` (insensitive) `*Temp` is the sensitive variant
> 
> `temp*` finds tags that start with `temp` (insensitive) `Temp*` is the sensitive variant

Thank you, should be fixed now

PR #754, opened 2025-01-31
dont autoplay media with media player

mashed5894:
alternatively autoplay muted, which is still default for mp4, etc.


CyanVoxel:
Is this PR trying to fix a bug or is it just changing the default behavior of video playback? If you don't wish to have videos autoplay on start, there's a right-click option that disables it.

PR #760, opened 2025-02-01
Correct Cagegory to Category in docs/usage.md

pinheadtf2:
I just saw this while reading up on categories, figured I'd submit a change so it doesn't go unnoticed


CyanVoxel:
Nice catch! Are there any other typos nearby that you know of?

pinheadtf2:
Not immediately! I'll make further commits as I see them, but so far everything else seems good.

PR #775, opened 2025-02-02
fix(docs): fix image not rendering

SkeleyM:
image wasn't showing in the docs

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -9,8 +9,10 @@ title: Home
 TagStudio is a photo & file organization application with an underlying tag-based system that focuses on giving freedom and flexibility to the user. No proprietary programs or formats, no sea of sidecar files, and no complete upheaval of your filesystem structure.
 
 <figure width="60%" markdown="span">
+
   ![TagStudio screenshot](assets/screenshot.png)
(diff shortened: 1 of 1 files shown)


CyanVoxel:
Interesting, the images already show up fine for me? 🤔 

image

SkeleyM:
how odd, I noticed this not working when checking the documentation on GitHub, maybe its due to my browser? however this also happens for me when rendering it in visual studio code
image

CyanVoxel:
> how odd, I noticed this not working when checking the documentation on GitHub, maybe its due to my browser? however this also happens for me when rendering it in visual studio code image

VSCode also doesn't render it properly for me, though that's likely due to MKDocs using such a weird syntax here which combines markdown and HTML. Although now that I try applying the change you made, it renders? Which kind of makes sense but also not really?

Formatting with Prettier also wants it to add an extra space in-between the markdown embed and the figcaption tag:
```html
<figure width="60%" markdown="span">

![TagStudio screenshot](assets/screenshot.png)

<figcaption>TagStudio Alpha v9.5.0 running on macOS Sequoia.</figcaption>

</figure>
```

SkeleyM:
combining html and markdown causes all kinds of weird quirks sometimes, ill add the extra space in-between the figcaption tag

Issue #771, opened 2025-02-02
Tag is no longer organized by color.


Daxslays:
Hello, I do not know if this was intentional or bug, but now the tabs are organize capitalization and then alphabetically rather than by color, capitalization, and then being alphabetized, thus it made searching and tagging items a bit of a pain. in 9.4.2, I would use parent tags to maintain the order, acting's as a make shift "categories" but now its all jumbled as what would have been:

(red) (Parent) Physique
(red) average (physique)
(red) petite (physique)
(red) heavy-set (physique)
(green) (Parent) Class
(green) warrior (Class)
(green) mage (Class)
(green) rouge (Class)
(auburn) (Parent) Race
(auburn) human (Race)
(auburn) orc (Race)
(auburn) elf (Race)

Is now

(green) (Parent) Class
(red) (Parent) Physique
(auburn) (Parent) Race
(red) average (physique)
(auburn) elf (Race)
(red) heavy-set (physique)
(auburn) human (Race)
(green) mage (Class)
(auburn) orc (Race)
(red) petite (physique)
(green) rouge (Class)
(green) warrior (Class)

Which absolutely destroyed my ability to go through the 50+ tags I have already made. IF this is intentional I can work around I would just like to know if this will reverted or repaired, if its a bug, before I overhaul the mess I have created.

CyanVoxel:
This is partially a bug with the tag search, partially intentional changes that have come as a result from the new color system. The separate capital/lowercase sorting is definitely a bug and is being fixed in #784, along with several other search issues.

Sorting by color, however, will unfortunately be gone in 9.5. The color system in 9.5 has changed from a small predetermined set of known colors to a variable set of built-in and user-defined colors that make it much harder to sort by compared to 9.4. With that being said though, an expanded tag management window is planned for 9.6 and I'd like to have color sorting and grouping make a comeback along with that. I understand that it's an important part of some people's tagging workflows, so I'll make sure to incorporate this feedback into the next tag manager rework!

Daxslays:
Thank you for the insight and consideration, never the less I am in love with this project and am ecstatic to see what comes next. :)

PR #784, opened 2025-02-03
fix(ui): improve "Add Tag" ux 

CyanVoxel:
### Summary
This PR makes several tweaks and changes to improve the tagging user experience. Since these were fixing lots of small bugs and issues, I decided not to open up any additional bug reports as it would take less time to fix these than it would be to open up reports for all of them.

### Fixes & Changes
- Reset tag search box and focus each time a tag search panel is opened
- Include tag parents in tag search results (v9.4 parity)
- Lowercase tag names now get properly sorted with uppercase ones (bug mentioned in #771)
- The "use for disambiguation" button has been moved to the right-hand side of parent tags in order to prevent accidental clicks involving the left-hand "remove tag" button
- Don't include tag display names in "closeness" factor when searching
- Add "Create & Add" button to the bottom of all non-whitespace searches, even if they return some tags
- Escape "&" characters inside tag names so Qt doesn't treat them as mnemonics
- (This was initially relevant to this PR) Make imports of `datetime` consistent
- The awkward "+" button next to tags in the "Add Tags" panel has been removed in favor of clicking on tags themselves
- Highlighting, keyboard focusing, and clicking tags has improved visual feedback
- The clickable area of the "-" button on tags has been increased and has visual feedback when you hover and click it
- Set minimum tag width
- You can now tab into the tag search list and add tags with a spacebar press (previously possible but very janky)
- In tag search panels, pressing the <kbd>Esc</kbd> key will return your focus to the search bar and highlight your previous query. If the search box is already highlighted, pressing <kbd>Esc</kbd> will close the modal
- In modals such as the "Add Tag" and "Edit Tag" panels, pressing <kbd>Esc</kbd> will cancel the operation and close the modal
- Fix "Add Tags" panel missing its window title when accessing from the keyboard shortcut
- Add <kbd>Ctrl</kbd>+<kbd>M</kbd> shortcut to open up the "Tag Manager"


Tishj:
> Include tag parents in tag search results (v9.4 parity)

This messes with the result order, child tags could show up before the parent tag, even if the tag is fully spelled out in the search

CyanVoxel:
> > Include tag parents in tag search results (v9.4 parity)

> This messes with the result order, child tags could show up before the parent tag, even if the tag is fully spelled out in the search

Don't worry, I noticed that too with a few tags and drafted this. My current plan is to split the results up by "priority" on the `search_tags` side of `library.py` instead of having the UI try and untangle a single results set. This should be slightly more performant and give a lot more granular control over how results are returned, too

Tishj:
Looks great!

It looks like you missed one though, I added this to the PanelWidget to make it respond to `Esc`
```patch
diff --git a/tagstudio/src/qt/widgets/panel.py b/tagstudio/src/qt/widgets/panel.py
index 141849b..2e22fb0 100755
--- a/tagstudio/src/qt/widgets/panel.py
+++ b/tagstudio/src/qt/widgets/panel.py
@@ -95,6 +95,11 @@ class PanelModal(QWidget):
         self.root_layout.addWidget(self.button_container)
         widget.parent_post_init()

+    def keyPressEvent(self, event: QtGui.QKeyEvent) -> None:  # noqa N802
+        if event.key() == QtCore.Qt.Key.Key_Escape:
+            self.close()
+        return super().keyPressEvent(event)
+
     def closeEvent(self, event):  # noqa: N802
         self.done_button.click()
         event.accept()
```

CyanVoxel:
> Looks great!
> 
> It looks like you missed one though, I added this to the PanelWidget to make it respond to `Esc`
> 
> ```diff
> diff --git a/tagstudio/src/qt/widgets/panel.py b/tagstudio/src/qt/widgets/panel.py
> index 141849b..2e22fb0 100755
> --- a/tagstudio/src/qt/widgets/panel.py
> +++ b/tagstudio/src/qt/widgets/panel.py
> @@ -95,6 +95,11 @@ class PanelModal(QWidget):
>          self.root_layout.addWidget(self.button_container)
>          widget.parent_post_init()
> 
> +    def keyPressEvent(self, event: QtGui.QKeyEvent) -> None:  # noqa N802
> +        if event.key() == QtCore.Qt.Key.Key_Escape:
> +            self.close()
> +        return super().keyPressEvent(event)
> +
>      def closeEvent(self, event):  # noqa: N802
>          self.done_button.click()
>          event.accept()
> ```

Thank you! However this patch overrides and breaks the following change in this PR:
> - In tag search panels, pressing the <kbd>Esc</kbd> key will return your focus to the search bar and highlight your previous query. If the search box is already highlighted, pressing <kbd>Esc</kbd> will close the modal

Tishj:
That is true, however currently Esc does nothing there, because the buttons take precedence and no behavior for Esc is set in its key press event handler

CyanVoxel:
#793: I've added the event handler to every extra missing modal as well as tweaked the TagSearchPanel's handler to selectively defer to PanelWidget's handler rather than completely override it

PR #791, opened 2025-02-04
Added references to alternative POSIX shells, as well as pyenv to CONTRIBUTING.md

ChloeZamorano:
Now `CONTRIBUTING.md` is gonna be helpful for when a user encounters problems with non-matching Python version (Which is a problem Arch Linux users specifically are sure to encounter for example), as it now references the official Python website to download the correct version, but also pyenv as viable solutions.

`CONTRIBUTING.md` also now elaborates on running the Python virtual environment on Linux and macOS, by providing a handy table of shells/scripts supported by Python's virtual environment system, taken directly from their [documentation](https://docs.python.org/3.12/library/venv.html).

The `.gitignore` file now also includes the `.python-version` file which is generated by pyenv to keep track of the version selected for the current folder, and should not be pushed to the remote repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,7 +88,9 @@ profile_default/
 ipython_config.py
 
 # pyenv
-.python-version
+#   For a library or package, you might want to ignore these files since the code is
+#   intended to run in multiple environments; otherwise, check them in:
+# .python-version
 
 # pipenv
 #   According to pypa/pipenv#598, it is recommended to include Pipfile.lock in version control.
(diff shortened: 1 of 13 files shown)


CyanVoxel:
Did you mean to include the `setWindowFlag()` changes in here?

ChloeZamorano:
I did not, I forked off the wrong branch on accident.

CyanVoxel (approved):
Thank you for this!

PR #829, opened 2025-03-04
fix: KeyError when formatting translation

### LINKED ISSUE #827: [Bug]: Switching translations causes error on program start
### Checklist

### TagStudio Version

Alpha 9.5.0

Windows 11 Home 23H2

### Description

I was testing languages for reporting a different bug, but when I switched to one of them (German, I think. It's near the top of the list, anyhow), the program refuses to open and throws this error.
Image
I also tried deleting TagStudio and re-extracting it to attempt to reproduce this, but the error seems to persist across installs.

### Expected Behavior

The program starts normally (and bugs to not persist between installs).

### Steps to Reproduce

Either:
1. Switch to affected language (again, I believe it's German)
2. Restart TagStudio
or rapidly switch languages, restarting the program each time, and it'll happen at some point. 

For obvious reasons, I can't test this at the moment. If I can somehow get TagStudio working again, I'll update this issue with more info (and make the other report I mentioned).

### Logs

_No response_

Computerdores:
Fix #827. I implemented a string subclass that handles the formatting differently. I did it that way because it keeps the Translator API simpler and when formatting the translation one can just use the normal way without having to look for a specific method, as opposed to creating a custom formatting method for `Translator`.

The formatting also formats all keys that a value is present for as opposed to the previous implementation (removed in #817) that didn't substitute any placeholders if the value for one of them wasn't specified.

## Tasks Completed

- Tested on:
  - [x] Windows x86
  - [ ] Windows ARM
  - [ ] macOS x86
  - [ ] macOS ARM
  - [ ] Linux x86
  - [ ] Linux ARM
  
- Tested:
  - [x] Basic functionality
  - [ ] PyInstaller executable

### DIFF
--- a/tagstudio/src/qt/translations.py
+++ b/tagstudio/src/qt/translations.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from pathlib import Path
 
 import structlog
@@ -6,6 +7,29 @@ import ujson
 logger = structlog.get_logger(__name__)
 
 DEFAULT_TRANSLATION = "en"
+
+
+class ErrorlessFormatString(str):
+    __translator: "Translator"
+
+    def __new__(cls, translator: "Translator", object: object = ""):
+        obj = super().__new__(cls, object)
+        obj.__translator = translator
+        return obj
+
+    def format(self, *args: object, **kwargs: object) -> str:
+        try:
+            return super().format(*args, **kwargs)
+        except KeyError:
+            logger.error(
+                "[Translations] Error while formatting translation.",
+                text=self,
+                language=self.__translator._lang,
+                kwargs=kwargs,
+            )
+            params: defaultdict = defaultdict(lambda: "{missing_key}")
+            params.update(kwargs)
+            return super().format_map(params)
 
 
 class Translator:
@@ -28,7 +52,9 @@ class Translator:
         self._strings = self.__get_translation_dict(lang)
 
     def __getitem__(self, key: str) -> str:
-        return self._strings.get(key) or self._default_strings.get(key) or f"[{key}]"
+        return ErrorlessFormatString(
+            self, self._strings.get(key) or self._default_strings.get(key) or f"[{key}]"
+        )
 
 
 Translations = Translator()


tagstudio/src/qt/translations.py
```diff
@@ -28,7 +52,9 @@ def change_language(self, lang: str):
         self._strings = self.__get_translation_dict(lang)
 
     def __getitem__(self, key: str) -> str:
-        return self._strings.get(key) or self._default_strings.get(key) or f"[{key}]"
+        return ErrorlessFormatString(
+            self, self._strings.get(key) or self._default_strings.get(key) or f"[{key}]"
+        )
```
CyanVoxel, line 57: While I like the ability to use strings that properly handle KeyErrors raised from the `.format()` call, I feel that subclassing `str` creates more type messiness and ambiguity wherever Translation strings are used. It also creates some incongruity when it comes to which `.format()` calls need to be error handled, where `str.format()` calls still need to be error handled everywhere else but get a free pass here.

This incongruent behavior can also be seen if we store the wrapped string somewhere or process it through a function where it is cast as a proper `str`, and *then* we attempt to use the same `.format()` method. It's seemingly  the same `format()` method used on the same text returned from the Translations class, but now it has the ability to raise errors in a way that may not be immediately apparent without knowledge of the wrapper sublcass.

tagstudio/src/qt/translations.py
```diff
@@ -8,6 +9,29 @@
...
+        try:
+            return super().format(*args, **kwargs)
+        except KeyError:
+            logger.error(
+                "[Translations] Error while formatting translation.",
+                text=self,
+                language=self.__translator._lang,
+                kwargs=kwargs,
+            )
+            params: defaultdict = defaultdict(lambda: "{missing_key}")
```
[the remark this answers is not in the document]
Computerdores: Not replacing the key with the missing value while still replacing the ones with a value would definitely better, but I am not sure how to do that elegantly which is why I haven't done so already

tagstudio/src/qt/translations.py
Computerdores: > It's seemingly the same format() method used on the same text returned from the Translations class, but now it has the ability to raise errors in a way that may not be immediately apparent without knowledge of the wrapper sublcass.

would you feel differently if the `__getitem__` returned `ErrorlessFormatString` instead of `str`? Because in that case the format method would explicitly be called on an `ErrorlessFormatString` not a `str`

CyanVoxel: > would you feel differently if the __getitem__ returned ErrorlessFormatString instead of str? Because in that case the format method would explicitly be called on an ErrorlessFormatString not a str

Did you mean that the other way around? `__getitem__` currently returns an `ErrorlessFormatString` and not a `str`. If the `ErrorlessFormatString` could somehow never escape containment of the Translations class and not bleed into the rest of the codebase then I would have no issues with it. In other words, if it was only ever a private object while the return type of any translation strings given (from `__getitem__`, `.format()`, etc.) are true Python `str` types with zero behavioral ambiguity, then that would side step my issue with wrapping a built-in type. Although I'm skeptical if that's possible due to the chaining of .format(), and feel that a formatting method native to the Translations class like the one you originally created would be the better way to go, which is what I brought back in #830.

Computerdores: What I mean is that currently the return type of the method is `str` meaning that from the outside it isn't apparent that a subclass of str is actually what is being returned. My question is whether changing that return type, thus making it obvious that the returned value *isn't* a normal string, would alleviate your concerns about the ambiguity of `format` due to being more clear than it is currently

CyanVoxel: My concerns don't come from the return type hint, but the actual consequences of being handed a wrapped built-in type that sometimes has the behavior of a `str` and sometimes not.

If `__getitem__` gives me a `str`-like object which I can use in a way such as `Translations["title"].format(invalid_key="")` with no consequences, then I'm going to assume that my format key was probably valid and that doing something like `Translations["title"].upper().format(invalid_key="")` would also work - but now that call will unknowingly crash the program. This is true for any case where an `ErrorlessFormatString` object is stored as, is cast to, or is processed as a native `str` before the  `.format()` method, outwardly identical to the one from `str`, is called upon it.

Computerdores: Good point, I suppose one way to fix this would making it a string wrapper instead of a subclass. At that point I am not sure though if I like that better or your approach

CyanVoxel: After the refactor in my approach that renamed "translate_formatted" to just "format", I believe there's a lot less incentive to find an alternative way of calling a format method. The differences in calls now look like this:
```py
Translations["trash.context.singular"].format(trash_term=trash_term()) #829
Translations.format("trash.context.singular", trash_term=trash_term()) #830
```
with the later even being more friendly to wrapping. All error handing is done within the Translations class in #830, and the `.format()` method itself is distinct in both its location, description, and signature than the one from `str`.

I think you had it right originally when having the format method as part of the Translations class, and with the name refactor it's already about as equally easy to call as this other syntax.

Computerdores:
Closing in favor of #830.

PR #830, opened 2025-03-04
fix: restore `translate_formatted()` method

### LINKED ISSUE #827: [Bug]: Switching translations causes error on program start

CyanVoxel:
### Summary
This PR restores the use of the `translate_formatted()` method (now refactored as just `format()`) that was removed in #817 in order to catch KeyErrors raised from invalid formatting keys being added in the translation files and passed to `.format()` calls.

### Tasks Completed

- Tested on:
  - [ ] Windows x86
  - [ ] Windows ARM
  - [ ] macOS x86
  - [x] macOS ARM
  - [ ] Linux x86
  - [ ] Linux ARM
  
- Tested for:
  - [x] Basic functionality
  - [ ] PyInstaller executable
  
  ***
  
  Closes #827, along with the fixed translations in #826.

### DIFF
--- a/tagstudio/src/qt/translations.py
+++ b/tagstudio/src/qt/translations.py
@@ -27,6 +27,21 @@ class Translator:
         self._lang = lang
         self._strings = self.__get_translation_dict(lang)
 
+    def __format(self, text: str, **kwargs) -> str:
+        try:
+            return text.format(**kwargs)
+        except (KeyError, ValueError):
+            logger.error(
+                "[Translations] Error while formatting translation.",
+                text=text,
+                kwargs=kwargs,
+                language=self._lang,
+            )
+            return text
+
+    def translate_formatted(self, key: str, **kwargs) -> str:
+        return self.__format(self[key], **kwargs)
+
     def __getitem__(self, key: str) -> str:
         return self._strings.get(key) or self._default_strings.get(key) or f"[{key}]"
 
--- a/tagstudio/src/qt/translations.py
+++ b/tagstudio/src/qt/translations.py
@@ -39,7 +39,7 @@ class Translator:
             )
             return text
 
-    def translate_formatted(self, key: str, **kwargs) -> str:
+    def format(self, key: str, **kwargs) -> str:
         return self.__format(self[key], **kwargs)
 
     def __getitem__(self, key: str) -> str:
--- a/tagstudio/src/qt/translations.py
+++ b/tagstudio/src/qt/translations.py
@@ -1,4 +1,6 @@
+from collections import defaultdict
 from pathlib import Path
+from typing import Any
 
 import structlog
 import ujson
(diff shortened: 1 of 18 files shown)


Computerdores:
If you end up choosing this PR over #829, I would suggest porting the formatting improvements and renaming `translate_formatted` to `formatted` as `Translations.translate_formatted` seems a bit redundant (I had changed this in #817 as well before removing the method altogether)

tagstudio/src/qt/translations.py
```diff
@@ -27,6 +27,21 @@ def change_language(self, lang: str):
...
+        except (KeyError, ValueError):
+            logger.error(
+                "[Translations] Error while formatting translation.",
+                text=text,
+                kwargs=kwargs,
+                language=self._lang,
+            )
+            return text
+
+    def translate_formatted(self, key: str, **kwargs) -> str:
```
Computerdores, line 42: I would suggest renaming this to `formatted` as outlined in the comment

Computerdores, line 42: I would suggest renaming this to formatted as outlined in the comment

tagstudio/src/qt/translations.py
```diff
@@ -27,6 +27,21 @@ def change_language(self, lang: str):
         self._lang = lang
         self._strings = self.__get_translation_dict(lang)
 
+    def __format(self, text: str, **kwargs) -> str:
```
Computerdores, line 30: I would suggest porting this improvements from #829

Computerdores (commented):
No issues besides what was mentioned before

tagstudio/src/qt/translations.py
CyanVoxel: I was looking forward to doing this as well

tagstudio/src/qt/translations.py
Computerdores: I would change it to this (plus an import for `defaultdict` at the top):
```python
    def __format(self, text: str, *args, **kwargs) -> str:
        try:
            return text.format(*args, **kwargs)
        except (KeyError, ValueError):
            logger.error(
                "[Translations] Error while formatting translation.",
                text=text,
                kwargs=kwargs,
                language=self._lang,
            )
            params: defaultdict = defaultdict(lambda: "{missing_key}")
            params.update(kwargs)
            return super().format_map(params)
```

CyanVoxel: Addressed in 499b548c09b6cbbe762b3fb74269f66877df58f2

PR #833, opened 2025-03-05
feat: add tests for translations

Computerdores:
### Summary
Test for the following problems:
- translation keys that are present in en.json not being present in other translations
- translation keys that are present other translations not being present in en.json
- translations using format keys that aren't used by the equivalent in en.json
-  translations not using format keys that *are* used by the equivalent in en.json

(Note that these tests currently fail and that that is intentional as the translations are incorrect at the moment)

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/tagstudio/tests/conftest.py
+++ b/tagstudio/tests/conftest.py
@@ -1,11 +1,11 @@
-import pathlib
 import sys
+from pathlib import Path
 from tempfile import TemporaryDirectory
 from unittest.mock import Mock, patch
 
 import pytest
 
-CWD = pathlib.Path(__file__).parent
+CWD = Path(__file__).parent
 # this needs to be above `src` imports
 sys.path.insert(0, str(CWD.parent))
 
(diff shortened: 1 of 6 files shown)


Computerdores:
this is now ready for review

CyanVoxel:
(Very funny how the failing test is a good thing in this case 😁)

One issue I have by looking at this though is that the test should not fail in the case of a full translation key being present in en.json but missing in other translations. This is expected and anticipated behavior for incomplete translations, and would mean that pytest would always fail unless each language was 100% translated. But the other 3 test scenarios listed are great things to test for

CyanVoxel (approved):
Thank you for this! Looks all good to me - Normally I'd recommend using implicit concatenation for the f-strings to prevent the E501, but in this case it would look pretty odd and I'll give it a pass since it's just a test file.

I'll mark this with "blocked" while I fix the remaining issues discovered from the test here on the Weblate side, and then pull the changes as a part of #826

CyanVoxel:
Actually, I'm noticing that these tests will only report a single issue per translation file at a time. Would it be straightforward to show all detected errors at once upfront?

CyanVoxel:
> Actually, I'm noticing that these tests will only report a single issue per translation file at a time. Would it be straightforward to show all detected errors at once upfront?

I'll merge this in the meantime since it's critical for v9.5.1, but it's something to consider for the future

PR #840, opened 2025-03-07
fix(ffmpeg): remove redundant log statement

Computerdores:
### Summary
A string for a log message was missing the `f` to make it a format string.

### Tasks Completed
-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/qt/modals/ffmpeg_checker.py
+++ b/src/tagstudio/qt/modals/ffmpeg_checker.py
@@ -45,7 +45,6 @@ class FfmpegChecker(QMessageBox):
         if which(FFPROBE_CMD):
             self.ffprobe = True
 
-        logger.info("FFmpeg found: {self.ffmpeg}, FFprobe found: {self.ffprobe}")
         return self.ffmpeg and self.ffprobe
 
     def version(self):


CyanVoxel:
I was curious why this wasn't caught before because I *swear* that I see valid FFmpeg location logs each time on startup, and then I noticed why - this log statement is redundant and doesn't even need to be here in the first place.

Two log statements in `ffmpeg.py` already take care of this job, so the log message here can even just be straight up removed 👀 

```yaml
2025-03-07 11:43:18 [info     ] [FFMPEG] Using FFprobe location: ffprobe
2025-03-07 11:43:18 [info     ] [FFMPEG] Using FFmpeg location: ffmpeg
```

Computerdores:
fair point, is gone now

CyanVoxel:
I apologize for the inconvenience, #844 has now been pulled and a rebase targeting `main` will be required here now.
Alternatively, I'm looking at fixing some larger issues with this file as it causes a regression with several CMD popups in Windows builds since it circumvents our vendored ffmpeg code. If you wish to rebase this then I'll pull it, otherwise I can take care of this in a future PR - whichever you prefer.

Thank you for your work on this, and sorry again for the disruption!

Computerdores:
rebase is done

xarvex:
Actually, are we sure the log is redundant? In testing an environment without FFmpeg, I still get these two log statements, I do not think they actually say FFmpeg is present, but rather that it is being tested:
```
2025-03-10 10:45:17 [info     ] [FFMPEG] Using FFprobe location: ffprobe
2025-03-10 10:45:17 [info     ] [FFMPEG] Using FFmpeg location: ffmpeg
```

Later on, I suppose after the check is done, there is this statement:
```
2025-03-10 10:45:20 [info     ] FFmpeg found: False, FFprobe found: False
```

CyanVoxel:
> Actually, are we sure the log is redundant? In testing an environment without FFmpeg, I still get these two log statements, I do not think they actually say FFmpeg is present, but rather that it is being tested:

You're right that the original log statement doesn't indicate whether or not it was found or not. However, the underlying code is capable of it and the ffmpeg_checker should be using that backend code (or expanding upon it) instead of being a combination of frontend and backend code that also circumvents fixes done in #436. Overall this is unfortunately just a "tip of the iceberg" issue with some more explanation now in #854

CyanVoxel (approved):
I'll be merging this into a new branch to use as the basis for a greater set of fixes and refactors involving this corner of the codebase. Thank you for the initial find and work on this!

PR #845, opened 2025-03-07
refactor: split translation keys for about screen

CyanVoxel:
### Summary
This PR refactors the "About TagStudio" window by reorganizing its widgets and splitting the translation keys out from the single monolithic `"about.content"` key to several smaller keys. This was prompted due to several hyperlinks being present in the base string which should ***NOT*** be made accessible for translators to modify.

#### Changes
- Refactored the UI widgets and layouts of `about.py`.
- Fixed DPI of logo for hi-DPI screens
- Split the `"about.content"` key into `"about.description"`, `"about.config_path"`, `"about.documentation"`, `"about.license"`, and more.
- Added translation keys for the "Found" and "Missing" strings used in FFmpeg module checks.
- Stopped proper names from being translated (e.g. "GitHub", "Discord", "FFmpeg").
- Removed HTML tags from the about screen translation strings.
- Removed the "help.visit_github" key as it's use was replaced by the current about screen keys.
- Removed an errant `<b>` tag in the translation key `"trash.dialog.permanent_delete_warning"`.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality


Computerdores:
I noticed that you split things like `Konfigurations-Pfad: {config_path}<br>` out into just `Konfigurations-Pfad` without the colon and placeholder. I believe we had in a previous case decided to include colons and such in translations because other languages may not use them. Is this just an oversight?

CyanVoxel:
> I noticed that you split things like `Konfigurations-Pfad: {config_path}<br>` out into just `Konfigurations-Pfad` without the colon and placeholder. I believe we had in a previous case decided to include colons and such in translations because other languages may not use them. Is this just an oversight?

You're right - in this case I was trying to avoid a scenario where I needed to give several additional new translations strings with the colons and placeholders to translate, which would introduce several new entry points for placeholder errors. The maintenance overhead of and fixing these broken placeholders has started to really build up for me, and it's not where I want to be spending my time for this project.

I've now addressed this inconsistency by adjusting the styling of the system info section to no longer relying on colons. I did try for quite a while to get the config path field's background to use the minimum amount of space needed, however I guess that's a quirk of Qt and word wrapping that I've been unable to discover a solution for.
image

CyanVoxel:
Merging this now as it's needed to prevent conflicts with incoming translations, and is needed as a base before #854 can be addressed

Issue #351, opened 2024-08-21
[Feature Request]: Render STL thumbnails


CJones-Optics:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

It would be awesome eventually to get thumbnails for STL files. The tagging system works great for a 3D printing workflow. (And I imagine other CAD workflows)

### Solution

I found a github repo for rendering STL file thumbnails for use in Windows or Gnome like DE: [REPO](https://github.com/unlimitedbacon/stl-thumb/tree/master)

Some kind of fork of this could be used. It is all written in rust which I am no expert in. 

### Alternatives

Issue #97 may provide an equivalent method of getting STL files. I am not sure which implmenetation would be easier. 

I don't know any Rust, and am not very good at software in general lmao. I might have a play with the repo and see if I can implement it, but opening the issue in case anyone with more experiance wants to tackle it.,

thibmaek:
Could act similar to [manyfold](https://github.com/manyfold3d/manyfold) for example. TagStudio would be a great alternative to manage a large library of 3D printing files (STL, OBJ, 3MF, SCAD, F3D) without the complexity of selfhosting or resource heavy usage of manyfold.

macOS renders STL thumbs natively, is there no way to tap into QuickLook (what's used for the thumbnail) here?
![image](https://github.com/user-attachments/assets/273d7261-befc-4d96-ad64-850bbd95451b)

CyanVoxel:
> I have been looking into rendering it myself using OpenGL support in QT like this [repo](https://github.com/alonrubintec/3DViewer) but that requires the usage of `QWindow` instead of `QWidget` for the `PreviewPanel`
> 
> I don't know QT enough to know the ramifications of such a change, and if it's even possible.

I'm also unfortunately not familiar enough with Qt to know the consequences off hand, but I feel it would either be a quick "well this doesn't work it's a separate window" or "okay this'll work I just need to shuffle around some object types". It's up to you if you wanna take a stab at it or just wait for open3d to roll around to 3.12. I couldn't exactly tell you the pros and cons of either implementation, but I feel open3d would be easier to implement for and expand upon.

JonatanNevo:
I'll take a stab at it, 
[...]

PR #693, opened 2025-01-08
WIP: Add 3d object rendering using open3d

JonatanNevo:

PR #860, opened 2025-03-13
feat: add model thumbnail rendering

Lunarient:
### Summary

This pull request adds 3D model thumbnail support for .obj, .gltf, .stl, .glb, and .fbx formats, using Open3D 0.19.0 for rendering. It supports textures and can color texture-less models. In the future, I plan to add a setting for users to choose their default model color.

This supersedes https://github.com/TagStudioDev/TagStudio/pull/693

This closes https://github.com/TagStudioDev/TagStudio/issues/351

#### Remaining Issue
The remaining issue with the pull request is that Open3D's OffscreenRenderer requires a dedicated thread, and it crashes if no thread adopts the renderer. I'm unsure how to integrate this with QThreads, so as a workaround, I set daemon=True for the rendering thread to prevent the software from hanging on exit. While this stops the hanging issue, it feels like a hacky solution.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [X] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [ ] Basic functionality

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ dependencies = [
     "typing_extensions>=3.10.0.0,<4.11.0",
     "ujson>=5.8.0,<5.9.0",
     "vtf2img==0.1.0",
+    "open3d==0.19.0"
 ]
 
 [project.optional-dependencies]
--- a/src/tagstudio/qt/helpers/model_thumbnailer.py
+++ b/src/tagstudio/qt/helpers/model_thumbnailer.py
@@ -1,0 +1,147 @@
+import threading
+from dataclasses import dataclass
+from pathlib import Path
+from queue import Empty, Queue
+
+import numpy as np
+import structlog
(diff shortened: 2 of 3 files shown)


Lunarient:
Okay I'm no longer certain it's the model renderer that's causing the seg fault. I disabled the renderer and it still seg faults when adding new files to a library, but otherwise I don't know what causes it. If I someone else can test this, then it should be good to merge

Lunarient:
Looks like the offscreenrenderer that comes with open3d only works on linux. So I need to try a different way to render off screen. I think there are other ways, they will just take more time. Back to the drawing board

CyanVoxel:
Closing, see discussion post #1231

Issue #847, opened 2025-03-08
[Feature Request]: Add .desktop file


Cool-Game-Dev:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

Currently when installed, TagStudio does not create a .desktop file, which causes issues with things such as icon packs and application launchers. 

### Solution

Create a .desktop file and add it during the install process.

zierf:
As already briefly mentioned in my [comment for PR #368](https://github.com/TagStudioDev/TagStudio/pull/368#issuecomment-2715712767), the recently updated Nix Flake is missing the entry for generating a `.desktop` file.

---

For my originally intended Flake with _Poetry2Nix_ I solved it as follows.

<details>
<summary>File flake.nix</summary>

__Filepath [/flake.nix](https://github.com/TagStudioDev/TagStudio/blob/d80f04ac6674cff882d999d21a4674a47a3d4bd1/flake.nix#L74-L94)__

Copy the TagStudio icon to the appropriate system path and define a `.desktop` file for the current system architecture (I didn't use a systems helper). Of course, this should also be compatible with the new metadata and the "pyproject.toml" in the end.

```SH
preInstall = ''
  mkdir -p $out/share/icons/hicolor/512x512/apps
  cp -rv ${projectDir}/tagstudio/resources/icon.png $out/share/icons/hicolor/512x512/apps/${pname}.png
'';

desktopItems = [
  (pkgs.${system}.makeDesktopItem {
    name = "${pname}";
    desktopName = "TagStudio";
    genericName = "Tagging Document Management System";
    comment = "A User-Focused Document Management System";
    categories = [
      "AudioVideo"
      "Qt"
    ];
    exec = "${pname}";
    icon = "${pname}";
    terminal = false;
    type = "Application";
  })
];
```
</details>

<details>
<summary>File ts_qt.py</summary>

__Filepath [/tagstudio/src/qt/ts_qt.py](https://github.com/TagStudioDev/TagStudio/blob/d80f04ac6674cff882d999d21a4674a47a3d4bd1/tagstudio/src/qt/ts_qt.py#L296-L298)__
(now different file path)

[...]

Issue #422, opened 2024-08-30
[Bug]: Can't pin Tagstudio to panel on Linux


Bamowen:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Alpha 9.3.2

### Operating System & Version

Linux Mint 22 Cinnamon [Cinnamon 6.2.9]

### Description

When right clicking the icon in the panel (a.k.a taskbar) there is no option to pin Tagstudio

### Expected Behavior

There is an option to pin the Tagstudio to the panel.
When selected, the program stays on the panel even when closed.

### Steps to Reproduce

1. Open Tagstudio
2. Right click on the panel icon
3. Observe no option to pin
![Screenshot from 2024-08-30 17-32-13](https://github.com/user-attachments/assets/7ea2d3b3-a765-4337-9c05-e22365f9cce4)

zierf:
I have a suspicion that this is a similar problem like the missing icon in #415, but this time it affects both Wayland and XWayland. It happens under KDE for me.

<details>
<summary>What it looks like now</summary>

![Screenshot_20240830_222230](https://github.com/user-attachments/assets/f3f2ffa1-2cf0-4974-b587-d8f42c4a0a56)
</details>

<details>
<summary>What it should look like</summary>

![Screenshot_20240830_223039](https://github.com/user-attachments/assets/7ec566f2-3abf-4cf1-bce3-ff84148fe88d)

</details>

[KDE Bug #436465](https://bugs.kde.org/show_bug.cgi?id=436465#c6) associates it with a missing _.desktop_ file.
"_This will happen if the application cannot be matched to the .desktop file used for launching that application._"

Issue #415, opened 2024-08-29
[Bug]: Wayland missing menu, application icon and title


zierf:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

main #cb4798b AND Alpha v9.4

### Operating System & Version

NixOS 24.11 (Vicuna), 24.11.20240818.8a33541 | nixpkgs-unstable#8a33541 | KDE

### Description

TagStudio does not display the global menu, appropriate icon or title under Wayland.

The entire menu is also missing if the desktop environment uses global menus (also called appmenu or dbus-menu). The menu is then neither displayed in the window itself nor in the global menu.

This means that the application is not usable at all.

Apart from NixOS, I also verified this on a Debian virtual machine with the pre-built `Alpha v9.3.2` release, and it behaves the same there.

![Screenshot_20240829_200922](https://github.com/user-attachments/assets/c06294db-3e2e-499d-bca1-d9bb592fa7f6)

---

Apparently Wayland needs a suitable `.desktop` file to display the correct title and icon. Just like someone referenced here with [app.setDesktopFileName("myapp")](https://stackoverflow.com/a/78355448) to set a system icon.

See also [this C++ example](https://github.com/mgba-emu/mgba/commit/333483a69e90d558830d6a6d96e72ac4dde00cca). You could possibly use something similar in the `tagstudio/src/qt/ts_qt.py` file.

```Python
if sys.platform != "darwin":
    icon = QIcon()
    icon.addFile(str(icon_path))
    app.setWindowIcon(icon)

if sys.platform == "linux":
    # app icon must be defined in a .desktop file to be displayed under Wayland
    app.setDesktopFileName("TagStudio");
```

Just for fun, I tried to use the icon for  _KDE-Discover_, which has a `.desktop` file on the system. This actually made the icon and title match.

```SH
if sys.platform == "linux":
    # app icon must be defined in a .desktop file to be displayed under Wayland
    app.setDesktopFileName("org.kde.discover");
```

![Screenshot_20240829_201108](https://github.com/user-attachments/assets/18abe66b-89b9-46d1-abbc-55d5c512f628)

---

[...]

PR #870, opened 2025-03-16
feat(resources): provide desktop file

xarvex:
### Summary

Provides a desktop file, and exposes it and the relevant icon for the Nix package.

Fixes: #415
Fixes: #422
Fixes: #847

Note that I cannot test pinning functionality as I do not use a desktop environment. Would appreciate help with testing that. Otherwise, I will create a virtual machine.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86 (NixOS)
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality
    
    -   [x] Nix package building

### DIFF
--- a/nix/package/default.nix
+++ b/nix/package/default.nix
@@ -58,6 +58,14 @@ buildPythonApplication {
     syrupy
   ];
 
+  # TODO: Install more icon resolutions when available.
+  preInstall = ''
+    mkdir -p $out/share/applications $out/share/icons/hicolor/512x512/apps
+
+    cp $src/src/tagstudio/resources/tagstudio.desktop $out/share/applications
+    cp $src/src/tagstudio/resources/icon.png $out/share/icons/hicolor/512x512/apps/tagstudio.png
+  '';
+
   makeWrapperArgs =
     [ "--prefix PATH : ${lib.makeBinPath [ ffmpeg-headless ]}" ]
     ++ lib.optional stdenv.hostPlatform.isLinux "--prefix LD_LIBRARY_PATH : ${
--- a/src/tagstudio/qt/ts_qt.py
+++ b/src/tagstudio/qt/ts_qt.py
@@ -297,10 +297,13 @@ class QtDriver(DriverMixin, QObject):
             appid = "cyanvoxel.tagstudio.9"
             ctypes.windll.shell32.SetCurrentProcessExplicitAppUserModelID(appid)  # type: ignore[attr-defined,unused-ignore]
 
-        if sys.platform != "darwin":
+        if platform.system() != "Darwin":
             icon = QIcon()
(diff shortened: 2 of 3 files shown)


zierf:
Looks pretty good. I would say it now works as intended.

```SH
nixpkgs: github:NixOS/nixpkgs/c80f6a7e10b39afcc1894e02ef785b1ad0b0d7e5?narHash=sha256-C7jVfohcGzdZRF6DO%2BybyG/sqpo1h6bZi9T56sxLy%2Bk%3D (2025-03-15 20:13:08)
```

```SH
$> nix-shell -p nix-info --run "nix-info -m"

 - system: `"x86_64-linux"`
 - host os: `Linux 6.13.6-xanmod1, NixOS, 25.05 (Warbler), 25.05.20250315.c80f6a7`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.24.12`
 - nixpkgs: `/nix/store/alzxn3hjisc84hrlv44x6hni48crww26-source`
```

---

I've temporarily switched my system flake's input to the commit in PR #870.
```Nix
tagstudio = {
  url = "github:TagStudioDev/TagStudio/77a1ac0d18b1ae9ffa8629f2cf92e352466bf941";
  inputs.nixpkgs.follows = "nixpkgs";
};
```

After that, I have a suitable entry in my Application Launcher.
Launcher

The AppMenu/GlobalMenu is still there and now has the appropriate icon and application title.
Global Menu

I was also able to pin the window to my dock (just a floating KDE bar using built-in functionality).
PinWindow Open
(Application open)

PinWindow Closed
(Application closed and still pinned)

---

If I don't install TagStudio via my system flake and run it directly from GitHub, it no longer has an icon or title (almost to be expected).

```Nix
nix run github:TagStudioDev/TagStudio/77a1ac0d18b1ae9ffa8629f2cf92e352466bf941#tagstudio-jxl
```

AppMenu, run from GitHub

Dock, run from GitHub

For a quick test, that should be enough, and I don't know if it would be possible to fix it, since it doesn't create a desktop file without being installed on the system.

I think this should be sufficient for a quick test for new users. I don't know if it would even be possible to create a desktop file when running directly, since the application isn't installed on the system.

xarvex:
I initially thought about prefixing `XDG_DATA_DIRS` to provide the desktop file and icon in that way, however that would mean stripping the user's ability to customize. I went ahead and provided default values instead for the application name and icon.

This won't address things like pinning, as there is no desktop entry, but that makes sense to me when the application is not installed.

PR #876, opened 2025-03-18
[Refactor] type fixes and minor improvements to file_opener.py

VasigaranAndAngel:
### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/qt/helpers/file_opener.py
+++ b/src/tagstudio/qt/helpers/file_opener.py
@@ -136,7 +136,7 @@ class FileOpenerLabel(QLabel):
         self.filepath = filepath
 
     @override
-    def mousePressEvent(self, ev: QMouseEvent) -> None:  # noqa: N802
+    def mousePressEvent(self, ev: QMouseEvent) -> None:
         """Handle mouse press events.
 
         On a left click, open the file in the default file explorer.
--- a/src/tagstudio/qt/helpers/file_opener.py
+++ b/src/tagstudio/qt/helpers/file_opener.py
@@ -7,10 +7,12 @@ import subprocess
 import sys
 import traceback
 from pathlib import Path
+from typing import override
 
 import structlog
 from PySide6.QtCore import Qt
-from PySide6.QtWidgets import QLabel
+from PySide6.QtGui import QMouseEvent
+from PySide6.QtWidgets import QLabel, QWidget
 
 from tagstudio.qt.helpers.silent_popen import silent_Popen
 
@@ -115,15 +117,17 @@ class FileOpenerHelper:
 
 
 class FileOpenerLabel(QLabel):
-    def __init__(self, parent=None):
+    def __init__(self, parent: QWidget | None = None) -> None:
         """Initialize the FileOpenerLabel.
 
         Args:
             parent (QWidget, optional): The parent widget. Defaults to None.
         """
+        self.filepath: str | Path | None = None
+
         super().__init__(parent)
 
-    def set_file_path(self, filepath):
+    def set_file_path(self, filepath: str | Path) -> None:
         """Set the filepath to open.
 
         Args:
@@ -131,20 +135,22 @@ class FileOpenerLabel(QLabel):
         """
         self.filepath = filepath
 
-    def mousePressEvent(self, event):  # noqa: N802
+    @override
+    def mousePressEvent(self, ev: QMouseEvent) -> None:  # noqa: N802
         """Handle mouse press events.
 
         On a left click, open the file in the default file explorer.
         On a right click, show a context menu.
 
         Args:
-            event (QMouseEvent): The mouse press event.
+            ev (QMouseEvent): The mouse press event.
         """
-        super().mousePressEvent(event)
-
-        if event.button() == Qt.MouseButton.LeftButton:
+        if ev.button() == Qt.MouseButton.LeftButton:
+            assert self.filepath is not None, "File path is not set"
             opener = FileOpenerHelper(self.filepath)
             opener.open_explorer()
-        elif event.button() == Qt.MouseButton.RightButton:
+        elif ev.button() == Qt.MouseButton.RightButton:
             # Show context menu
             pass
+        else:
+            super().mousePressEvent(ev)


src/tagstudio/qt/helpers/file_opener.py
```diff
@@ -115,36 +117,40 @@ def open_explorer(self):
...
+        if ev.button() == Qt.MouseButton.LeftButton:
+            assert self.filepath is not None, "File path is not set"
             opener = FileOpenerHelper(self.filepath)
             opener.open_explorer()
-        elif event.button() == Qt.MouseButton.RightButton:
+        elif ev.button() == Qt.MouseButton.RightButton:
             # Show context menu
             pass
+        else:
+            super().mousePressEvent(ev)
```
CyanVoxel, line 156: Why put the `super()` call in the conditional?

Computerdores, line 156: Nit: What is the reason for renaming the parameter here? (From a practical standpoint, `event` is more verbose and thus, imo, reads better and there really isn't much difference between typing `ev` and typing `ev`+Tab)

Computerdores (commented):
Looks good besides what Cyan pointed out

src/tagstudio/qt/helpers/file_opener.py
CyanVoxel: `ev` is the real base parameter name for this method, so naming it something different creates an invalid override. Qt seems to be wildly inconsistent with their parameter names, with some event ones being `event`, `ev`, `e`, and even just `arg__1`

Computerdores: that is upsetting

VasigaranAndAngel: if `super()` method of an event method called, it passes the event to the next widget (like parent widget, but not just the parent). so, in mouse related events, if there's a widget below this one, the event will go to that widget. if the event is already handled, there is no need to pass it to the next one (unless there is a specific reason to do so).

_it's a bit tricky for me to explain this properly_ 😅. _let me know if this explanation doesn't make sense._

src/tagstudio/qt/helpers/file_opener.py
```diff
@@ -115,36 +117,40 @@ def open_explorer(self):
...
-
-        if event.button() == Qt.MouseButton.LeftButton:
+        if ev.button() == Qt.MouseButton.LeftButton:
+            assert self.filepath is not None, "File path is not set"
             opener = FileOpenerHelper(self.filepath)
             opener.open_explorer()
-        elif event.button() == Qt.MouseButton.RightButton:
+        elif ev.button() == Qt.MouseButton.RightButton:
             # Show context menu
             pass
```
VasigaranAndAngel, line 154: what kind of context menu it should show?

src/tagstudio/qt/helpers/file_opener.py
CyanVoxel: I'm aware of what the `super()` call does, I was just curious if you noticed some unintended behavior and moved it inside the conditional because of that. Otherwise we normally call `super()` regardless if the event is handled or not so that it's able to propagate out to the parent class. In this case I don't think it makes a difference either way since QLabel doesn't really have a mouse press event to begin with, but yeah - I was just wondering if there was a specific motivation that prompted this change.

src/tagstudio/qt/helpers/file_opener.py
CyanVoxel: This is an old code comment not left by any of us, but I believe it's referring to showing a right-click context menu that would contain options for opening the file or file location, similar to file entries or the preview panel.

VasigaranAndAngel: is that need to be removed?

CyanVoxel: Might as well keep it since there really *should* be a context menu there eventually

src/tagstudio/qt/helpers/file_opener.py
VasigaranAndAngel: passing the event to the next widget essentially makes that widget act kind of transparent. ideally, none of the widgets in an app should be transparent. so i think the event should only be passed to the next widget if it’s really needed.

CyanVoxel: The click wouldn't be passing though the label as if it were transparent, FileOpenerLabel would use the event in the overridden `mousePressEvent` method and then pass the event to the base class, which in this case is QLabel. And since QLabels don't really do anything on click this doesn't make a difference. But if this were a different class that had `mousePressEvent` handling of it's own then that would be quietly overridden here. Ultimately it's not a big thing, I just wanted to see if this change was due to an active issue uncovered.

VasigaranAndAngel: yes it will pass it to the base class which is `QLabel`. then the `QLabel `will pass it to the next widget. which makes that label act like transparent.

mousepress3

in our case, `FileOpenerLabel` is a clickable widget. so i think it shouldn't be transparent.

there are no any issues with this behavior. but i feel like that's not the ideal way. that's why i changed it.

CyanVoxel: > yes it will pass it to the base class which is `QLabel`. then the `QLabel `will pass it to the next widget. which makes that label act like transparent.
> in our case, `FileOpenerLabel` is a clickable widget. so i think it shouldn't be transparent.
> 
> there are no any issues with this behavior. but i feel like that's not the ideal way. that's why i changed it.

Interesting, I wonder if that's due to how QLabel is handing (or rather *not* handling) the `mousePressEvent` from the underlying QWidget class. In either case, I think you're right here 👍

CyanVoxel (approved):
Thank you for these changes!

src/tagstudio/qt/helpers/file_opener.py
VasigaranAndAngel: `QEvent` has two methods. `ignore` and `accept`. when `ignore` method is called, Qt sends that event to the next widget. for example in the case of mouse related events, it will send the event to the widget underneath. i'm not sure how other events are propagated. if the `accept` method of `QEvent` is called, or if `ignore` isn't called at all, it won't propagate further.

when we call `super().mousePressEvent()`, it performs the default action. for `QLabel` the default action is calling `ignore`. (`QLabel` inherits from `QWidget`. so i think it’s just calling the `mousePressEvent` of `QWidget`. and `QWidget`'s default action is calling `ignore`.). when `ignore` is called, it will pass the event to the next widget. if the widget is button widget, the event is handled and `accept` is called. so the event stops there.

this is my understanding.

PR #880, opened 2025-03-21
[Refactor] Type fixes and minor improvements to flowlayout.py

VasigaranAndAngel:
### Summary

- Added `@override` decorators and removed `# noqa: N802` comments.
- Changed parameter names in override methods to match the base methods.
- Used `setProperty` to assign the "ignore_size" property to `FlowWidget`.
- Fixed `HeightForWidth` to return `int` instead of `float` since the base method is annotated to return `int`.
- Simplified the if statement in line 127.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [ ] Basic functionality

### DIFF
--- a/src/tagstudio/qt/flowlayout.py
+++ b/src/tagstudio/qt/flowlayout.py
@@ -50,9 +50,9 @@ class FlowLayout(QLayout):
         return None
 
     @override
-    def takeAt(self, arg__1: int) -> QLayoutItem | None:
-        if 0 <= arg__1 < len(self._item_list):
-            return self._item_list.pop(arg__1)
+    def takeAt(self, index: int) -> QLayoutItem | None:  # pyright: ignore[reportIncompatibleMethodOverride]
+        if 0 <= index< len(self._item_list):
+            return self._item_list.pop(index)
 
         return None
 
--- a/src/tagstudio/qt/flowlayout.py
+++ b/src/tagstudio/qt/flowlayout.py
@@ -43,9 +43,9 @@ class FlowLayout(QLayout):
         return len(self._item_list)
 
     @override
-    def itemAt(self, arg__1: int) -> QLayoutItem | None:
-        if 0 <= arg__1 < len(self._item_list):
-            return self._item_list[arg__1]
+    def itemAt(self, index: int) -> QLayoutItem | None:  # pyright: ignore[reportIncompatibleMethodOverride]
+        if 0 <= index< len(self._item_list):
+            return self._item_list[index]
 
         return None
 
--- a/src/tagstudio/qt/flowlayout.py
+++ b/src/tagstudio/qt/flowlayout.py
@@ -5,24 +5,28 @@
 
 """PySide6 port of the widgets/layouts/flowlayout example from Qt v6.x."""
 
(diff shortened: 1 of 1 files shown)


src/tagstudio/qt/flowlayout.py
```diff
@@ -5,71 +5,85 @@
...
 
-    def takeAt(self, index):  # noqa: N802
-        if 0 <= index < len(self._item_list):
-            return self._item_list.pop(index)
+    @override
+    def takeAt(self, arg__1: int) -> QLayoutItem | None:
+        if 0 <= arg__1 < len(self._item_list):
+            return self._item_list.pop(arg__1)
 
         return None
```
VasigaranAndAngel, line 57: I don't know how to fix this. PySide itself returns None if the index is out of range. but not type annotated.

src/tagstudio/qt/flowlayout.py
```diff
@@ -5,71 +5,85 @@
...
 
-    def takeAt(self, index):  # noqa: N802
-        if 0 <= index < len(self._item_list):
-            return self._item_list.pop(index)
+    @override
+    def takeAt(self, arg__1: int) -> QLayoutItem | None:
+        if 0 <= arg__1 < len(self._item_list):
+            return self._item_list.pop(arg__1)
```
Computerdores, line 55: Same as with the other method: According to pyright the parent method's argument is called `index` and the pyright comment should take care of the complaint.
```suggestion
    @override
    def takeAt(self, index: int) -> QLayoutItem | None:  # pyright: ignore[reportIncompatibleMethodOverride]
        if 0 <= index< len(self._item_list):
            return self._item_list.pop(index)
```

src/tagstudio/qt/flowlayout.py
```diff
@@ -5,71 +5,85 @@
...
+    def count(self) -> int:
         return len(self._item_list)
 
-    def itemAt(self, index):  # noqa: N802
-        if 0 <= index < len(self._item_list):
-            return self._item_list[index]
+    @override
+    def itemAt(self, arg__1: int) -> QLayoutItem | None:
+        if 0 <= arg__1 < len(self._item_list):
+            return self._item_list[arg__1]
```
Computerdores, line 48: According to pyright the parent method's argument is called `index` and the pyright comment should take care of the complaint.
```suggestion
    @override
    def itemAt(self, index: int) -> QLayoutItem | None:  # pyright: ignore[reportIncompatibleMethodOverride]
        if 0 <= index< len(self._item_list):
            return self._item_list[index]
```

Computerdores (approved):
All in all a good improvement, I see no issues besides those two nits.

src/tagstudio/qt/flowlayout.py
VasigaranAndAngel: ah my bad. let me fix that.

Computerdores (approved):

Issue #810, opened 2025-02-19
[Bug]: Error when accessing hidden folder causing program to crash


spyingcreeper09:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Alpha 9.4.2

### Operating System & Version

Windows 10

### Description

There is a folder I want hidden on my computer but I also want to use this tool. When I want to use the tool I have no issue having to manually unhide the folder to use TagStudios but what I dont like is how it gives this long error message and just crashes. I have to unhide the folder first in otder to even get the program to open so I can use it (this includes if i want to open a different library)

### Expected Behavior

I think that it sould provide a very basic error message within the center of the page (Where is says "Opening library 'folder path here'") that says "Cannot access previous library"

![What happens](https://github.com/user-attachments/assets/502ad95a-246d-4cd7-9ad4-821a37b70874)
![What I want to happen](https://github.com/user-attachments/assets/08de2de4-64f8-459c-ab47-4f1205a8d6d1)
![What heppens (without error box)](https://github.com/user-attachments/assets/7275388e-a3bc-4617-b968-2c97e22e0508)

### Steps to Reproduce

1. Create a new library in a folder
2. Close tag studios
3. Go into the folder properties and check "Hidden" then apply
4. Open tag studios
5. Wait for error

### Logs

Failed to execute script 'tag_studio' due to unhandled exception: input(): lost sys.stdin

Traceback (most recent call last):
  File "tag_studio.py", line 75, in main
  File "src\qt\ts_qt.py", line 638, in start
    self.open_library(Path(lib))
  File "src\qt\ts_qt.py", line 1836, in open_library
    return_code = self.lib.open_library(path)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src\core\library.py", line 748, in open_library
    self.save_library_backup_to_disk(is_auto=True)
  File "src\core\library.py", line 856, in save_library_backup_to_disk
    with open(
         ^^^^^
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\SC09\\OneDrive\\Desktop\\⠀\\.TagStudio\\backups\\ts_library_backup_auto.json'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #881, opened 2025-03-21
ui: recent libraries list improvements

CyanVoxel:
### Summary

This PR makes a couple tweaks to the Recent Libraries menu item:
1. Add a more specific error message when trying to open a library location that no longer exists *(while keeping in mind that the full library path may not always be shown anymore, and in the future library files may not be in the same location as the folders they use)*
2. Increase the recent library list limit from 5 to 10

image

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/qt/ts_qt.py
+++ b/src/tagstudio/qt/ts_qt.py
@@ -1857,8 +1857,18 @@ class QtDriver(DriverMixin, QObject):
         open_status: LibraryStatus | None = None
         try:
             open_status = self.lib.open_library(path)
+        except ValueError as e:
+            logger.warning(e)
+            open_status = LibraryStatus(
+                success=False,
+                library_path=path,
+                message=Translations["menu.file.missing_library.title"],
+                msg_description=Translations.format(
+                    "menu.file.missing_library.message", library=library_dir_display
+                ),
+            )
         except Exception as e:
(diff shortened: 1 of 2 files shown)


kitsumed:
I think this pull will also close : https://github.com/TagStudioDev/TagStudio/issues/810.
But I'm not sure 100%.

CyanVoxel:
> I think this pull will also close : #810. But I'm not sure 100%.

#810 seems unrelated as it's a different error in a different part of the codebase

Computerdores:
Wouldn't it be better if it always showed  the full path in the "Library Not Found" message? After all, knowing that a directory called "XYZ" wasn't found *somewhere* on the drive isn't all that useful (and not everyone is on top of their file organisation enough to immediately know where they had the library)

Computerdores (approved):
Besides me uncovering an unrelated bug (#889) while testing, this seems work perfectly!
(I still think that it would be good to have a setting in the future whether to show the full path in the popup, but this seems good as is)

PR #885, opened 2025-03-22
fix: remove unescaped ampersand from "about.description"

CyanVoxel:
### Summary
Remove an unescaped ampersand from the "about.description" translation key which caused the text to render incorrectly. This could have been addressed by escaping the ampersand (e.g. "&&") but I've opted to replace it with the word "and" instead.

Before:
image
After:
image

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality


Computerdores (approved):
Sieht gut aus!

PR #887, opened 2025-03-22
Preview display improvement

Gawidev:
### Summary

Simple useful improvement to previews in the inspector. Previews are only loaded if the widget is open. For now, allows for a significant increase in responsiveness for users looking to simply manage their library while dev efforts are still underway.
Also split off the few ruff/pyright/mypy error fixes in the affected files into their own commit just in case.

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality


Computerdores (commented):
This seems to work as intended, however I think it would be better to instead decouple the loading of the image and the changing of the UI, so that the preview panel can update while the image is still loading. And I would then further recommend applying this uncoupled loading to the thumbnails in the grid as well.
I would personally not want to merge this, as it would mostly be dead weight once the loading is decoupled from the UI. However I would be interested in @CyanVoxel's take on this, especially since this does have meaningful value in the meantime.

VasigaranAndAngel:
> This seems to work as intended, however I think it would be better to instead decouple the loading of the image and the changing of the UI, so that the preview panel can update while the image is still loading. And I would then further recommend applying this uncoupled loading to the thumbnails in the grid as well. I would personally not want to merge this, as it would mostly be dead weight once the loading is decoupled from the UI. However I would be interested in @CyanVoxel's take on this, especially since this does have meaningful value in the interrim.

I've been thinking about "thumbnail decoupling" for a while too. we should discuss about this somewhere else. so someone will try to implement it. it is a must needed optimization.

Gawidev:
If the efforts towards asynchronous thumbnail/preview loading have begun, then it would indeed be pretty pointless. I created it now only since I've found no mention of it being worked on or planned for. If it isnt merged by 9.5.2 then yeah I guess it will have been pointless 🤷

PR #895, opened 2025-03-30
build: Minor improvements to PyInstaller.

Leonard2:
### Summary
Minor changes to pathex and datas

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/tagstudio.spec
+++ b/tagstudio.spec
@@ -25,7 +25,7 @@ elif system == "Darwin":
 
 a = Analysis(
     ["src/tagstudio/main.py"],
-    pathex=[],
+    pathex=["src"],
     binaries=[],
     datas=[("src/tagstudio", "tagstudio")],
     hiddenimports=[],
--- a/tagstudio.spec
+++ b/tagstudio.spec
@@ -23,11 +23,17 @@ elif system == "Darwin":
     icon = "src/tagstudio/resources/icon.icns"
 
 
+datafiles = [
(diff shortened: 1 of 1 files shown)


CyanVoxel:
Could you explain why you made these changes and what they improve upon?

Leonard2:
As far as I could gather, when PyInstaller processes an import within a script passed to it, it will look directly into your sys.paths for the dependency (including TagStudio itself).
This means that if you didn't "pip install .", PyInstaller will just throw a bunch of errors about being unable to find the "TagStudio" module.
As such the correct way to handle this is to add the source itself to the sys.path while compiling (that being the pathext field's purpose).
[This can even be seen in the documentation example](https://pyinstaller.org/en/stable/spec-files.html#spec-file-operation):
vivaldi_re1uD0KfvQ

Following this logic, we also do not need to add the raw source py files to the "datas" field as the TagStudio module is simply processed like the rest of the dependencies and compiled to bytecode into the executable.

CyanVoxel (approved):
Looks good, the changes seem to be follow the PyInstaller docs nicely. Thank you for this!

*(Tested in builds on macOS 15 and Windows 10)*

PR #903, opened 2025-04-01
fix(test): Fix tests to pass on windows without disrupting other platforms

zfbx:
### Summary

Replaced `dev/null` path with a universal `temp_dir` function that persists through the test session and cleans itself up afterwards to function identically but runs on windows as well as others? I don't have another platform to test this on so it will need to be tested.

<sub>Sorry, it was bugging me I wanted tests to pass xD<sub>

image

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -60,9 +60,7 @@ def temp_dir():
 
     test_dir.mkdir(parents=True, exist_ok=True)
 
-    yield test_dir
-
-    shutil.rmtree(test_dir, ignore_errors=True)
+    return test_dir
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,4 @@
 import sys
-import tempfile
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from unittest.mock import Mock, patch
@@ -10,6 +9,7 @@ CWD = Path(__file__).parent
 # this needs to be above `src` imports
 sys.path.insert(0, str(CWD.parent))
 
+from tagstudio.core.constants import THUMB_CACHE_NAME, TS_FOLDER_NAME
 from tagstudio.core.library.alchemy.library import Library
 from tagstudio.core.library.alchemy.models import Entry, Tag
 from tagstudio.qt.ts_qt import QtDriver
@@ -52,28 +52,29 @@ def file_mediatypes_library():
 
 
 @pytest.fixture(scope="session")
-def temp_dir():
-    """Creates a shared library path for tests, and cleans it up after the session."""
-    with tempfile.TemporaryDirectory() as tmp_dir_name:
-        temp_root = Path(tmp_dir_name)
-        yield temp_root
+def library_dir():
+    """Creates a shared library path for tests, that cleans up after the session."""
+    with TemporaryDirectory() as tmp_dir_name:
+        library_path = Path(tmp_dir_name)
+
+        thumbs_path = library_path / TS_FOLDER_NAME / THUMB_CACHE_NAME
+        thumbs_path.mkdir(parents=True, exist_ok=True)
+
+        yield library_path
 
 
 @pytest.fixture
-def library(request, temp_dir: Path):
+def library(request, library_dir: Path):
     # when no param is passed, use the default
-    library_path = str(temp_dir)
+    library_path = library_dir
     if hasattr(request, "param"):
         if isinstance(request.param, TemporaryDirectory):
-            library_path = request.param.name
+            library_path = Path(request.param.name)
         else:
-            library_path = request.param
-
-    thumbs_path = Path(library_path) / ".TagStudio" / "thumbs"
-    thumbs_path.mkdir(parents=True, exist_ok=True)
+            library_path = Path(request.param)
 
     lib = Library()
-    status = lib.open_library(Path(library_path), ":memory:")
+    status = lib.open_library(library_path, ":memory:")
     assert status.success
 
     tag = Tag(
@@ -142,34 +143,32 @@ def entry_full(library: Library):
 
 
 @pytest.fixture
-def qt_driver(qtbot, library):
-    with TemporaryDirectory() as tmp_dir:
+def qt_driver(qtbot, library, library_dir: Path):
+    class Args:
+        settings_file = library_dir / "settings.toml"
+        cache_file = library_dir / "tagstudio.ini"
+        open = library_dir
+        ci = True
 
-        class Args:
-            settings_file = Path(tmp_dir) / "settings.toml"
-            cache_file = Path(tmp_dir) / "tagstudio.ini"
-            open = Path(tmp_dir)
-            ci = True
+    with patch("tagstudio.qt.ts_qt.Consumer"), patch("tagstudio.qt.ts_qt.CustomRunnable"):
+        driver = QtDriver(Args())
 
-        with patch("tagstudio.qt.ts_qt.Consumer"), patch("tagstudio.qt.ts_qt.CustomRunnable"):
-            driver = QtDriver(Args())
+        driver.app = Mock()
+        driver.main_window = Mock()
+        driver.preview_panel = Mock()
+        driver.flow_container = Mock()
+        driver.item_thumbs = []
+        driver.autofill_action = Mock()
 
-            driver.app = Mock()
-            driver.main_window = Mock()
-            driver.preview_panel = Mock()
-            driver.flow_container = Mock()
-            driver.item_thumbs = []
-            driver.autofill_action = Mock()
+        driver.copy_buffer = {"fields": [], "tags": []}
+        driver.copy_fields_action = Mock()
+        driver.paste_fields_action = Mock()
 
-            driver.copy_buffer = {"fields": [], "tags": []}
-            driver.copy_fields_action = Mock()
-            driver.paste_fields_action = Mock()
-
-            driver.lib = library
-            # TODO - downsize this method and use it
-            # driver.start()
-            driver.frame_content = list(library.get_entries())
-            yield driver
+        driver.lib = library
+        # TODO - downsize this method and use it
+        # driver.start()
+        driver.frame_content = list(library.get_entries())
+        yield driver
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -65,7 +65,7 @@ def temp_dir():
 @pytest.fixture
 def library(request, temp_dir: Path):
     # when no param is passed, use the default
-    library_path = temp_dir
+    library_path = str(temp_dir)
     if hasattr(request, "param"):
         if isinstance(request.param, TemporaryDirectory):
             library_path = request.param.name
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -50,16 +52,34 @@ def file_mediatypes_library():
...
+    temp_root = Path(tempfile.gettempdir())
+    test_dir = temp_root / "ts-test-temp"
+
+    if test_dir.exists():
+        shutil.rmtree(test_dir)
+    test_dir.mkdir(parents=True)
+
+    yield test_dir
+
+    shutil.rmtree(test_dir, ignore_errors=True)
```
CyanVoxel, line 67: I don't believe it's necessary to yield and remove the `test_dir` here, since the test(s) that rely on this path (in this case the `thumbs_path` and `test_path` code) recreates `test_dir` anyway via the `parents=True` argument in `mkdir`. And since the later code's paths are never cleaned up, the real logic doing the cleanup is always going to be the `if test_dir.exists()` check that precedes this.
I've tested this on Windows, macOS, and Linux and it works as intended. 
```suggestion
    return test_dir
```

tests/conftest.py
```diff
@@ -50,16 +52,34 @@ def file_mediatypes_library():
...
 
+@pytest.fixture(scope="session")
+def temp_dir():
+    """Creates a shared library path for tests, and cleans it up after the session."""
+    temp_root = Path(tempfile.gettempdir())
+    test_dir = temp_root / "ts-test-temp"
+
+    if test_dir.exists():
+        shutil.rmtree(test_dir)
+    test_dir.mkdir(parents=True)
```
CyanVoxel, line 63: Is there a reason why you wouldn't want to use the `exist_ok` parameter here?
(Tests pass with this on at least Windows and macOS)
```suggestion
    test_dir.mkdir(parents=True, exist_ok=True)
```

zfbx: Just to clear it if it already existed from a failed test prior (I had many goes to get this right 😓) to be thorough, but it can be dropped.

tests/conftest.py
CyanVoxel: Removing the yield + rmtree() here should be fine 👍

tests/qt/test_file_path_options.py
```diff
@@ -106,9 +106,14 @@ def test_file_path_display(
         ),
     ],
 )
-def test_title_update(qt_driver: QtDriver, filepath_option: ShowFilepathOption, expected_title):
+def test_title_update(
+    qt_driver: QtDriver, filepath_option: ShowFilepathOption, expected_title, temp_dir
```
Computerdores, line 110: Nit: `temp_dir` is missing a type hint

tests/conftest.py
```diff
@@ -50,16 +51,30 @@ def file_mediatypes_library():
...
+    test_dir = temp_root / "ts-test-temp"
+
+    test_dir.mkdir(parents=True, exist_ok=True)
+
+    return test_dir
+
+
 @pytest.fixture
-def library(request):
+def library(request, temp_dir):
```
Computerdores, line 66: Not: `temp_dir` is missing a type hint

tests/conftest.py
```diff
@@ -50,16 +51,30 @@ def file_mediatypes_library():
...
 
+@pytest.fixture(scope="session")
+def temp_dir():
+    """Creates a shared library path for tests, and cleans it up after the session."""
+    temp_root = Path(tempfile.gettempdir())
+    test_dir = temp_root / "ts-test-temp"
+
+    test_dir.mkdir(parents=True, exist_ok=True)
+
+    return test_dir
```
Computerdores, line 62: Is there any reason you are not using `tempfile.NamedTemporaryDirectory` here? Also, if you change to that, it might make more sense to create a new one in each test in order to isolate the different tests from each other (I would in that case specifically suggest making use of it's ability to be used with `with`)

Computerdores (commented):
Haven't been able to test it yet, but it seems to be good apart from what I commented on.

tests/qt/test_file_path_options.py
```diff
@@ -106,9 +106,14 @@ def test_file_path_display(
...
+):
     base_title = qt_driver.base_title
-    test_path = Path("/dev/null")
+
+    test_path = temp_dir
+    test_path.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
zfbx: yep, code that was left over from my attempts to get this functioning that I missed when cleaning up. removed in commit

tests/conftest.py
zfbx: I can't find any `NamedTemporaryDirectory`, only [NamedTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile). I actually preferred when it would clean up after itself rather than leaving directories behind but I'm not sure what's proper in this sort of case. I've not had much experience in testing, I just wanted to get them to pass on windows 😅 I'm open to changes suggested if you think it would be better

Computerdores: My bad the vlass js just called `TemporaryDirectory` and you can access its Path via the property `name`, see https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

CyanVoxel: > This module creates temporary files and directories. It works on all supported platforms. [TemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile), [NamedTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile), [TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory), and [SpooledTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile) are high-level interfaces which provide automatic cleanup and can be used as [context managers](https://docs.python.org/3/glossary.html#term-context-manager). [mkstemp()](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp) and [mkdtemp()](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp) are lower-level functions which require manual cleanup.

`TemporaryDirectory` seems to be an easy solution that cleans up after itself, and I agree that it might make more sense to use this in each relevant test rather than pass the same value around

zfbx: Updated to use `TemporaryDirectory` and tested working.

tests/conftest.py
```diff
@@ -50,16 +51,27 @@ def file_mediatypes_library():
     return lib
 
 
+@pytest.fixture(scope="session")
```
CyanVoxel, line 54: I'm not sure what the difference between the default scope vs the session scope here in practice, as they both seem to generate different temp directories when this fixture is called.

The only tangible effect I can observe is having `(scope="session")` causes the tests in `test_refresh_new_files()` to fail if using my suggested removal of the strange pre-existing "`hasattr`" code...

CyanVoxel (commented):
I've got a few more suggestions to look into after the most recent changes, some affecting pre-existing code that was seemingly trying to do what you're currently trying to do now.

I also caught an old note of mine inside the `CacheManager` that described special behavior that was in direct response to the "dev/null" assumptions in these tests up until now. This note and check can now be removed (`cache_manager.py`):
```diff
- # NOTE: The /dev/null check is a workaround for current test assumptions.
- if self.last_lib_path and self.last_lib_path != Path("/dev/null"):
+ if self.last_lib_path and self.last_lib_path:
```

tests/conftest.py
zfbx: For this my understanding of it is that when you use the `scope="session"` fixture it runs the code inside it once and then any further calls to it return the same temporary directory path so it can be reused across multiple tests and then when the session is complete it runs cleanup and deletes the directory which is why the yield is used instead of return because return causes the program to end before TempDir does it's cleanup leaving the files in the temp directory.

Which leads to the other comment, Yeah, I did notice you had more uses of TemporaryDirectory in other test but I wasn't sure if there was a reason for it maybe needing fresh directories for different tests. I only built it this way because it seemed test_file_path_options:test_title_update() depended on the path being exactly the same as used with conftest:library() which is probably why you had it the way you did.

I hope any of this made sense cause I'm only like 80% following my own words here myself

tests/conftest.py
```diff
@@ -50,16 +51,27 @@ def file_mediatypes_library():
...
             library_path = request.param.name
         else:
             library_path = request.param
 
+    thumbs_path = Path(library_path) / ".TagStudio" / "thumbs"
+    thumbs_path.mkdir(parents=True, exist_ok=True)
+
     lib = Library()
     status = lib.open_library(Path(library_path), ":memory:")
     assert status.success
```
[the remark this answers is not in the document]
zfbx: I will see what I can do here 😅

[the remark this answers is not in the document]
zfbx: > looked into what this whole `hasattr` block is actually doing. Turns out it's _also_ using `TemporaryDirectory` but in a much more confusing and roundabout way. I've been able to greatly simplify this section of the code by removing it and using the `temp_dir` parameter directly - here I renamed it to `library_dir`:

Okay, I see what's going on, doing this was because of [this](https://github.com/zfbx/TagStudio/blob/1abf77d5845a38c39d73c04131f29ac68c073c0a/tests/macros/test_refresh_dir.py#L13-L20) on `test_refresh_dir:test_refresh_new_files()` needed a unique directory for it's test and it goes through library() to do so.

New changes up and pass on Windows x86 and Mac x86 (technically x64?)

Computerdores (commented):
The changes look good!
What I noticed though is that the json migration test fails for me and I can't properly check why because I keep getting this error (this was already present before the tests broke on windows):
```
d:\Coding\TagStudio\tests\test_json_migration.py::test_json_migration failed with error:  
 The python test process was terminated before it could exit on its own, the process errored with: Code: 3221226505, Signal: null
```
And the only way to circumvent that is to run *all* tests at once, really weird.
Only thing I can tell atm is that `check_field_parity` is what fails

Computerdores:
> The changes look good! What I noticed though is that the json migration test fails for me and I can't properly check why because I keep getting this error (this was already present before the tests broke on windows):
> 
> ```
> d:\Coding\TagStudio\tests\test_json_migration.py::test_json_migration failed with error:  
>  The python test process was terminated before it could exit on its own, the process errored with: Code: 3221226505, Signal: null
> ```
> 
> And the only way to circumvent that is to run _all_ tests at once, really weird. Only thing I can tell atm is that `check_field_parity` is what fails

Ok I also got this error:
image
And running the tests twice in the terminal had the migration tests fail in only one of the runs... I think this might just be weirdness of my setup.

CyanVoxel (approved):
Sorry for the delay on getting back to this - Everything seems to function on my end (macOS + Windows) and it looks like things are ready to pull.

PR #904, opened 2025-04-01
feat: add date_format and hour_format settings

JCC1998:
### Summary
Added date_format and hour_format settings to allow customization on how to visualize dates and hours in the preview_panel
Date format selector:
image

Time format setting:
image

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality


zfbx:
functions perfectly well on my end. Few comments though, nothing major:
- The dropdown can be a little confusing if you open it in the first 12 days of a month, would it be bad to add a date format along side the current date preview like `MM/DD/YYYY - 04/01/2025`?
- would having optional 0 padding like in the win10 example https://github.com/TagStudioDev/TagStudio/pull/898#pullrequestreview-2730122658 be desirable? either by another checkbox or adding it date formats? (maybe just where it's relevant to a standardized international format as to not make the dropdown twice as long if that route was chosen?)
- In the dropdown it seems to be sorted for the most part by the first value with exception to the first which feels out of place and is doubled later on? Is this intentional or by mistake? If intentional for functionality sake could the order be flipped to put the month-first options first to remove the dupe?
Code_GhkLISLLwg

Just thoughts to consider, love the work though! :D I'm very much a YYYY-MM-DD person myself.

JCC1998:
Hi! Thank you so much for your feedback

About adding the format string on the side, I would agree, the result is:
image
If it's too much information, maybe I could use a constant date like 31-01-2025 or in american english 01-31-2025 instead of the current date

About the 0 padding, i did not find the solution on the docs: https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes, but searching on google I found that adding a minus sign should do the job. I can add another option as a checkbox, so the selector does not get even bigger

And about the dupped entries, it was my fault, the first one was the default option, so I started adding under it. Later I tried to do some iterations and forgot about the default, which was dupped. Thanks for the heads-up!

If you think that these changes I mentioned are correct, I will start implementing them

Thanks again!

JCC1998:
Hi!
I just added the zero-padding setting. Needs check in MacOs and Linux, as I don't have any machine with those OS, so i can't test if the zero padding character works. For windows i had to set the symbol as '#'
If there's something wrong or does not work properly, let me know. I also don't know how this PR work if I have to change the status so it can be reviewed again or i don't have to do anything
Thanks!

zfbx:
So on my mac When i change the date format and press save it crashes the program entirely leaving this as it's final log 😓 
```sh
2025-04-06 05:42:14 [warning  ] [PanelModal] add_callback not implemented for SettingsPanel
python(38886,0x7ff84f26a100) malloc: Corruption of free object 0x7fa44f84f1f0: msizes 30/0 disagree
python(38886,0x7ff84f26a100) malloc: *** set a breakpoint in malloc_error_break to debug
```

On Win10 it seems to run fine.

Personally not sure about the naming for the zero padding. It could confuse people and it's not clear it's referring to being related to the date format. And maybe I'm nit picking but the dropdown feels difficult to read through to find a desired option, I wonder if there would be a better way to order it. Just my 2 cents however, not critical by any means.

For your reference this is what the dropdown looks like on mac for me
Screenshot 2025-04-06 at 5 49 44 AM

JCC1998:
About the crash, not really sure what's causing it. Did you use your mac to test the previous branch changes? Maybe the settings file has a wrong format stored. As you can see in my newest changes, i think I havent touched anything that it could break, because it was working fine for you previously, right?

About the naming of zero-padding, I can change the literal to "Date Zero Padding" or something similar, so it's more clear

About the dropdown, i need to understand why it's difficult to read: Is it too much information? It's difficult because your most used options are not the firstly shown? Would you group them by the separators ("/", ".", "-"), instead of d/m, m/d, y-m-d? Having the Year format as the whole year or the last 2 digits is confusing because it's too many options?

If you want to discuss this on discord so I can troubleshoot what's causing the crash, let me know

Thanks again for your feedback!

zfbx:
You're right, it seems to be me.. I might have bigger problems on my side. Ignore the Mac error, someone else will need to verify 😓 

I think maybe it's abbreviated years mixed in with the full years.. and maybe the preview using the current date is busy, though I get it's purpose maybe it would be better to have a fixed date example rather than the MM/DD AND current day? maybe something with a bit of significance like first commit `04/17/2024` or The day it was announced on YouTube to be open sourced `08/21/2024`? 😄 Just thoughts. I'm just a random contributor though, @CyanVoxel 's opinions might differ.. Maybe even has a better date in mind? :)

JCC1998:
Changed the literal of zero-padding setting and used a constant date to make it clearer by uncluttering the selector options text
Thanks for the feedback!

src/tagstudio/qt/modals/settings_panel.py
```diff
@@ -37,6 +38,26 @@
...
+    "%x": f"{DEFAULT_DATE.strftime("%x")}",
+    "%m/%d/%Y": f"{DEFAULT_DATE.strftime("%m/%d/%Y")}",
+    "%m-%d-%y": f"{DEFAULT_DATE.strftime("%m-%d-%y")}",
+    "%m-%d-%Y": f"{DEFAULT_DATE.strftime("%m-%d-%Y")}",
+    "%m.%d.%y": f"{DEFAULT_DATE.strftime("%m.%d.%y")}",
+    "%m.%d.%Y": f"{DEFAULT_DATE.strftime("%m.%d.%Y")}",
+    "%Y/%m/%d": f"{DEFAULT_DATE.strftime("%Y/%m/%d")}",
+    "%Y-%m-%d": f"{DEFAULT_DATE.strftime("%Y-%m-%d")}",
+    "%Y.%m.%d": f"{DEFAULT_DATE.strftime("%Y.%m.%d")}",
+}
```
zfbx, line 59:
```suggestion
DATE_FORMAT_MAP: dict[str, str] = {
    "%d/%m/%y": "21/08/24",
    "%d/%m/%Y": "21/08/2024",
    "%d.%m.%y": "21.08.24",
    "%d.%m.%Y": "21.08.2024",
    "%d-%m-%y": "21-08-24",
    "%d-%m-%Y": "21-08-2024",
    "%x": "08/21/24",
    "%m/%d/%Y": "08/21/2024",
    "%m-%d-%y": "08-21-24",
    "%m-%d-%Y": "08-21-2024",
    "%m.%d.%y": "08.21.24",
    "%m.%d.%Y": "08.21.2024",
    "%Y/%m/%d": "2024/08/21",
    "%Y-%m-%d": "2024-08-21",
    "%Y.%m.%d": "2024.08.21",
}
```

Wouldn't this be a bit simpler? unless we plan on updating the date value or try to update the values to match the zero padding (but that might be overkill aswell) it might be more performant to use static strings?

JCC1998: As you said, keeping the static strings might be the best option. Thanks!

TrigamDev:
Would it be possible to do something similar to how Obsidian handles formatting, where you can specify your own formatting
image

JCC1998:
It could be done, but i'm not sure it would be the most user friendly usage of it. It could be added as a future improvement of this feature, having a new option on the selector to "Custom", which would enable a text input (which is hidden or disabled when option is not selected)
Even so, I would still keep this implementation as it is for now, so most users would be able to change easily the date format without the need to learn how this format works. And for the looks of the documentation, it would be required to have a "parser" or some kind of translator to replace YYYY to the directives used in python '%Y', and YY to '%y' and so on
This development started because it was the most common used on other OS, so it might be easier to use / recognizable for most of the people using the program, but it should not limit the implementation you are asking
Let me know what you think, and thanks for taking a look at this!

PR #909, opened 2025-04-04
feat: add ability to invert selection

zfbx:
### Summary

Add the ability to select unselected thumbnails while unselecting selected thumbnails.. thus inverting your selection. I found that when I'm trying to go through a lot of tags sometimes it's more of a Boolean operation where I select a bunch to tag with one thing but want to select all the opposite ones to tag them with the other tag. I used what was considered a fairly common hotkey for invert selection (at least it's the one that comes up when you google "hotkey to select inverse")

Sorry I didn't open an issue first, It seemed small enough of a thing I didn't mind to throw it together as I didn't see any mention of anything similar in any issues or on the roadmap and I wont be offended if it's rejected but it certainly would be helpful to have :)

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [x] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/qt/ts_qt.py
+++ b/src/tagstudio/qt/ts_qt.py
@@ -473,6 +473,21 @@ class QtDriver(DriverMixin, QObject):
         self.select_all_action.setEnabled(False)
         edit_menu.addAction(self.select_all_action)
 
+        self.select_inverse_action = QAction(Translations["select.inverse"], menu_bar)
+        self.select_inverse_action.triggered.connect(self.select_inverse_action_callback)
+        self.select_inverse_action.setShortcut(
+            QtCore.QKeyCombination(
+                QtCore.Qt.KeyboardModifier(
+                    QtCore.Qt.KeyboardModifier.ControlModifier
+                    ^ QtCore.Qt.KeyboardModifier.ShiftModifier
+                ),
+                QtCore.Qt.Key.Key_I,
+            )
+        )
+        self.select_inverse_action.setToolTip("Ctrl+Shift+I")
+        self.select_inverse_action.setEnabled(False)
+        edit_menu.addAction(self.select_inverse_action)
+
         self.clear_select_action = QAction(Translations["select.clear"], menu_bar)
         self.clear_select_action.triggered.connect(self.clear_select_action_callback)
         self.clear_select_action.setShortcut(QtCore.Qt.Key.Key_Escape)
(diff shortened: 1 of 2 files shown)


Computerdores (approved):
Looks good overall and works on my end.
Only nitpick I have is that I usually see this type of feature called "Invert Selection" and not "Select Inverse", but that is it.

zfbx:
> Only nitpick I have is that I usually see this type of feature called "Invert Selection" and not "Select Inverse", but that is it.

I'm curious what places you see it like that? Only place in my research that shows that is SolidWorks (I've not used it personally). Clip Studio has `Select > Invert Selected Area`. Affinity uses `Select > Invert Pixel Selection`. Gimp uses `Select > Invert` and photoshop (the one I'm most familiar with) uses `Select > Inverse`. All of them group them under a `Select` menu (except solidworks) so It's a bit hard to compare.

Photoshop_kDWx64eO9f
AffinityPhoto_D8SXb85Vnu
gimp-3_EgnXemHuW8
solidworks

I also just thought it worked best to fit in the menu and align closer with the translation but I'm open to change it though if other's think it would be clearer that way 😄

Computerdores:
> > Only nitpick I have is that I usually see this type of feature called "Invert Selection" and not "Select Inverse", but that is it.
> 
> I'm curious what places you see it like that? Only place in my research that shows that is SolidWorks (I've not used it personally). Clip Studio has `Select > Invert Selected Area`. Affinity uses `Select > Invert Pixel Selection`. Gimp uses `Select > Invert` and photoshop (the one I'm most familiar with) uses `Select > Inverse`. All of them group them under a `Select` menu (except solidworks) so It's a bit hard to compare.
> 
> Photoshop_kDWx64eO9f AffinityPhoto_D8SXb85Vnu gimp-3_EgnXemHuW8 solidworks
> 
> I also just thought it worked best to fit in the menu and align closer with the translation but I'm open to change it though if other's think it would be clearer that way 😄

I mean three out of four of the images you gave are already closer to "Invert Selection" than "Select Inverse". I also might be used to it from german now that I think about it, because there it is usually written that way (the sentence structure is similar so things like that are also often similar in my experience).
Being closer to the existing translations is definitely a fair point in general, however if I am not mistaken two of the close QActions are translated as "Select All" and "Clear Selection" so imo it doesn't really make a difference in that regard.

On a pure personal opinion basis: "Select Inverse" feels a bit inaccurate to me because really what it does isn't just selecting the inverse of what is currently selected, but also deselecting what is currently selected. Where as "Invert Selection" seems quite accurate, because for each visible entry whether they are selected will be inverted.

Hope that helps understand where I am coming from 😄

zfbx:
-grumbles- jk xDD But yeah, you make a good point.. it does more clearly suggest that it's unselecting with `Invert Selection` that `Select Inverse` I'll change it to that 😛

xarvex (approved):
After that small change, everything looks great and straightforward to me! Verified working on Linux.

PR #912, opened 2025-04-11
feat: add minimalist "__main__.py" to 'tagstudio' module

un-pogaz:
### Summary

Adding a minimalist `__main__.py` to `tagstudio` module to explicitly define the entry point, and so that TagStudio can be run directly from `python -m tagstudio`.

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/__main__.py
+++ b/src/tagstudio/__main__.py
@@ -1,0 +1,11 @@
+#!/usr/bin/env python
+# Copyright (C) 2024 Travis Abendshien (CyanVoxel).
+# Licensed under the GPL-3.0 License.
+# Created for TagStudio: https://github.com/CyanVoxel/TagStudio
+
+import sys
+
+from tagstudio.main import main
+
+if __name__ == "__main__":
+    sys.exit(main())


Computerdores:
Wouldn't it be better to just rename `main.py` to `__main__.py`?

un-pogaz:
No, it will be a breaking change. This PR is just a bypass to be compliant with the python standard.

Computerdores:
> No, it will be a breaking change.

Technically true, however, starting TS directly from this file is, to my knowledge, not intended anyways, see the wiki:
> TagStudio can now be launched via the `tagstudio` command in your terminal.

As long as `tagstudio.spec` and `pyproject.toml` are adjusted accordingly (both 1 line changes) nothing should change for the vast majority users and even for those users (or devs) where something does change, the effort to fix it is miniscule.

And so because I see no value in having two "main" files around and the impact on users is miniscule, I would still advocate for renaming `main.py` all together.

PR #924, opened 2025-04-25
feat: clickable links in text fields

### LINKED ISSUE #506: [Feature Request]: Add the ability to open the URL in the URL field
### Checklist

### Description

Allow the URL in the URL field to be opened in the user's browser.

### Solution

_No response_

### Alternatives

_No response_

TrigamDev:
### Summary
Text fields will now replace all* URLs in their text with clickable links. Text fields have also been changed to support links and be able to open them when clicked.

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality
    

---

Closes #506


python357-1:
Confirmed working on Linux x86 and macOs ARM. Not sure if there is a way to do this but maybe you could make it so that I don't have to type "http(s)://" in front of my URLs? ex: "tagstud.io" isn't clickable, but "https://tagstud.io" is.
Screenshot 2025-06-03 at 4 34 49 PM

TrigamDev:
Unfortunately, that's not exactly the easiest to do without creating a LOT of false positives. Requiring the protocol seems to be common for most apps/websites, so that's what I went with.

Computerdores:
> Unfortunately, that's not exactly the easiest to do without creating a LOT of false positives. Requiring the protocol seems to be common for most apps/websites, so that's what I went with.

one approach might be looking at the tld of the candidate domain, because which ones are valid is quite limited (there is a wikipedia page that lists all of them)

TrigamDev:
Hm, that's true, though there's still the possibility of it running into false positives if the user happens to include a period anywhere in a text field, which isn't ideal

Computerdores:
definitely, but this kind of automatic detection in text always does.

For that reason I personally favor manually marking things as links with markdown (having markdown WYSIWYG for notes is something that has been discussed before as well, not sure if there is commitment on that yet though), however I am not sure what would be best for the average joe (I write most documents in LaTeX or Markdown so I am not exactly representative of the average person in that regard)

python357-1:
I guess I didn't realize this also does it in text fields but I was talking about the URL field specifically. Maybe we could have full protocol checking in regular text fields and no checking in URL fields?

TrigamDev:
After looking into it a bit, it seems like there's over 1k TLDs that would have to be checked for, which.. isn't very feasible to check for

TrigamDev:
> definitely, but this kind of automatic detection in text always does.
> 
> For that reason I personally favor manually marking things as links with markdown (having markdown WYSIWYG for notes is something that has been discussed before as well, not sure if there is commitment on that yet though), however I am not sure what would be best for the average joe (I write most documents in LaTeX or Markdown so I am not exactly representative of the average person in that regard)

This could be a good solution, though it would have to be explained to the user somehow, as some people probably aren't very familiar with markdown

TrigamDev:
> > definitely, but this kind of automatic detection in text always does.
> > For that reason I personally favor manually marking things as links with markdown (having markdown WYSIWYG for notes is something that has been discussed before as well, not sure if there is commitment on that yet though), however I am not sure what would be best for the average joe (I write most documents in LaTeX or Markdown so I am not exactly representative of the average person in that regard)
> 
> This could be a good solution, though it would have to be explained to the user somehow, as some people probably aren't very familiar with markdown

Hell, if going that route, there's the possibility of just going ahead and adding markdown rendering support to text fields. Though, that's a whole other can of worms.

TrigamDev:
At least for now though, I think the current implementation is fine enough for the time being, unless any problems are found

python357-1:
I don't think we have to go that far, just a regex (like this, for example: `/^https?:\/\/(?:www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b(?:[-a-zA-Z0-9()@:%_\+.~#?&\/=]*)$/`) for text fields and any text in a URL field be clickable as a link.

TrigamDev:
> I don't think we have to go that far, just a regex (like this, for example: `/^https?:\/\/(?:www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b(?:[-a-zA-Z0-9()@:%_\+.~#?&\/=]*)$/`) for text fields and any text in a URL field be clickable as a link.

After checking it really quickly, it doesn't match the URL you're trying to match
Screenshot 2025-06-03 182916

python357-1:
Yeah, I think going without protocol in text fields would add too many false positives, like you said. Putting the full protocol in text fields and not having to do it in URL fields feels like a good compromise

TrigamDev:
In that case, I think a specific URL widget would have to be added, as currently the URL-related fields just use the text box widget

python357-1:
In that case I'm just going to say it's good as is once the ruff errors are fixed.

TrigamDev:
Thing is, the Ruff errors aren't related to my commit, they're from the source. Fixing them would mean editing random source code that's entirely unrelated to the feature itself, and should be left to a PR focused on fixing them.

TrigamDev:
Ok nvm, apparently I misremembered

src/tagstudio/qt/widgets/text.py
```diff
@@ -19,9 +21,22 @@ def __init__(self, title, text: str) -> None:
         self.text_label = QLabel()
         self.text_label.setStyleSheet("font-size: 12px")
         self.text_label.setWordWrap(True)
-        self.text_label.setTextInteractionFlags(Qt.TextInteractionFlag.TextSelectableByMouse)
+        self.text_label.setTextFormat(Qt.TextFormat.RichText)
```
Computerdores: I would prefer setting this to markdown, as (afaik) the long term plan is to have WYSIWYG markdown editing for various fields. This would also mean changing the substitution further down.

(Or not setting it at all so that people don't get used to / have their libraries heavily use RichText Formatting)

src/tagstudio/qt/widgets/text.py
```diff
@@ -19,9 +21,22 @@ def __init__(self, title, text: str) -> None:
...
         self.base_layout.addWidget(self.text_label)
         self.set_text(text)
 
     def set_text(self, text: str):
+        text = linkify(text)
         self.text_label.setText(text)
+
+# Regex from https://stackoverflow.com/a/6041965
+def linkify(text: str):
+    url_pattern = r"(http|ftp|https):\/\/([\w_-]+(?:(?:\.[\w_-]+)+))([\w.,@?^=%&:\/~+#-*]*[\w@?^=%&\/~+#-*])"
```
Computerdores: Would recognising `file://` be desirable as well? Because major browsers support that as well

src/tagstudio/qt/widgets/text.py
```diff
@@ -19,9 +21,22 @@ def __init__(self, title, text: str) -> None:
...
+    return re.sub(
+        url_pattern,
+        lambda url: f'<a href="{url.group(0)}">{url.group(0)}</a>',
+        text,
+        flags=re.IGNORECASE
+    )
```
Computerdores: Nit: Missing trailing new line

src/tagstudio/qt/widgets/text.py
ToxicMushroom: I think compiling this regex with re.compile() once and storing it as a global might make it quicker, thought ?

Computerdores: yeah good point

TrigamDev:
Disabled the Ruff "line too long" error (can't exactly add line breaks to the middle of a regex pattern) and added a trailing new line

Computerdores:
I think technically you can (see [re.VERBOSE](https://docs.python.org/3/library/re.html#re.VERBOSE)) but its fine here imo

TrigamDev:
Turned out to be unnecessary regardless

CyanVoxel:
Are there any more changes desired for this implementation in the short term? Even if it doesn't cover 100% of use cases or could be done without using regex, it would probably be better to have at least some system providing clickable links than nothing at the moment

CyanVoxel (approved):

TrigamDev:
> Are there any more changes desired for this implementation in the short term? Even if it doesn't cover 100% of use cases or could be done without using regex, it would probably be better to have at least some system providing clickable links than nothing at the moment

Sorry, forgot to reply. Short-term, yes it's complete.

I think a good long-term solution for handling links could possibly be to include a simple UI-based markdown editor for long form text fields, such as descriptions. That way the user can specify exactly what is and isn't a link, among being able to use the rest of markdown's styling.
This is an entirely different feature of course, and I can open a separate issue for it.

(Discord's "Post Guidelines" input as an example for a simple UI for it)
image

PR #933, opened 2025-05-15
fix: Page Navigation State

### LINKED ISSUE #804: [Bug]: Page Navigation State Incorrectly Refactored and Non-Functional
### Checklist

### TagStudio Version

Alpha v9.5.0 Pre-Release 3

macOS 15.1.1

### Description

Due to a series of ill-informed refactors the page navigation state system has become completely changed post-v9.4 and has broken core navigation functionality. These refactors have also tied frontend-specific state management into the SQL backend, making it difficult to separate and restore pre-v9.5 functionality. 

### Expected Behavior

The page navigation state should function more or less how it did in v9.1-v9.4. This includes the following core functions:
- Forward/backward navigation between "page states" of search results
- Controlled by the "<" and ">" buttons next to the search bar as well as the back/forward buttons on supported mice
- Page state includes:
    - Page contents
    - Page index
    - Search query of current page
    - Scrollbar position of page
- Navigation between pages should swap out these properties when navigating
- Page navigation states can be informed by *but not replaced by* the page navigation buttons at the bottom of the screen. These should **not** replace navigation states like they were refactored to do in 9.5, but rather serve as an entry point to create additional navigation frames for the navigation state system, such as in 9.4 and below.

As an analogy for how the prior and intended behavior worked, the library navigation state system should behave similar to a web browser's navigation system. A stack is created based on the previous pages you've visited, including the web addresses, scrollbar positions, and subpage locations. Navigating through these states should reflect the order in which you've visted them, not the order of how pages are set up on a website itself.

A copy of how the navigation should function is present in the navigation code from v9.1-v9.4. It also remains present in the v9.5 codebase, specifically inside one of the methods that was refactored to break this behavior:
```
# Ex. User visits | A ->[B]     |
#                 | A    B ->[C]|
#                 | A   [B]<- C |
#                 |[A]<- B    C |  Previous routes still exist
#                 | A ->[D]     |  Stack is cut from [:A] on new route
```

Regarding some of the newer intertwined state management *things* introduced in the v9.5 codebase: the page size limit should be purely a frontend attribute. This should not be part of the backend codebase, and should *not* be stored inside the `ts_library.sqlite` file. A full list of Entry ID results should be returned by the library backend and correctly sliced and worked with by the frontend. 

### Steps to Reproduce

N/A - Codebase issue

### Logs

N/A

Computerdores:
### Summary
Fix #804.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [ ] Basic functionality

### DIFF
--- a/src/tagstudio/core/library/alchemy/enums.py
+++ b/src/tagstudio/core/library/alchemy/enums.py
@@ -71,55 +71,45 @@ class SortingModeEnum(enum.Enum):
     PATH = "file.path"
 
 
+# TODO: rename to remove term "filter"
 @dataclass
 class FilterState:
     """Represent a state of the Library grid view."""
 
-    # these should remain
-    page_size: int
     page_index: int = 0
     sorting_mode: SortingModeEnum = SortingModeEnum.DATE_ADDED
     ascending: bool = True
 
-    # these should be erased on update
     # Abstract Syntax Tree Of the current Search Query
     ast: AST | None = None
 
-    @property
-    def limit(self):
-        return self.page_size
-
-    @property
-    def offset(self):
-        return self.page_size * self.page_index
+    @classmethod
+    def show_all(cls) -> "FilterState":
+        return FilterState()
 
     @classmethod
-    def show_all(cls, page_size: int) -> "FilterState":
-        return FilterState(page_size=page_size)
+    def from_search_query(cls, search_query: str) -> "FilterState":
+        return cls(ast=Parser(search_query).parse())
 
     @classmethod
-    def from_search_query(cls, search_query: str, page_size: int) -> "FilterState":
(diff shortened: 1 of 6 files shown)


Computerdores:
Everything should be working now.

Note: I had to change the constructing methods of `BrowsingState` to produce a string that is then later parsed. It would be good if those got extra scrutiny during the review to make sure I didn't miss any edge cases (I didn't find any during testing, but still)

CyanVoxel (approved):
Everything looks good in the code as far as I can tell, and the functionality seems to be restored to 9.4 behavior! Thank you so much for this major fix!!

PR #935, opened 2025-05-16
refactor: split QtDriver into View and Controller to follow MVC model

Computerdores:
### Summary
Splits the View aspects out of `QtDriver` into the existing Main Window class so that only the Control aspects remain.
This will reduce the size of `ts_qt.py` and should generally make the code more maintainable.

Note: This is based off of #933 as that PR also modifies `ts_qt.py` and merging afterwards would be a pain.
(which also means that atm the #933 commits show up in this PR as well)

### Tasks Completed
-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [ ] Basic functionality


Computerdores:
I was going to continue here, but it has been a while and this is already quite big, so I am putting this up for review as is.

If this rots for too long because of its size, I am fine with just closing it and maybe using it as a inspiration for smaller, future refactors that are easier to review

CyanVoxel:
Everything seems to function here, I think this would be good to pull as-is as long as the conflicts are cleaned up. It seems that there's a couple larger conflicts in `ts_qt.py` but other than that is just small leftovers from #933 and such. Since it's been a while if you'd rather not jump back into this I can take a look at untangling the conflicts and getting this merged

Computerdores:
apologies for the force push, but that was required since the commits from #933 were squashed which meant that a rebase was required to remove them from this branch

CyanVoxel (approved):
Thank you for your work on this! Any breaking up of `ts_qt.py` is always lovely to see

PR #939, opened 2025-05-26
fix: creating new tag now refreshes the menu using the current search…

### LINKED ISSUE #932: [Bug]: Tag search results not preserved after adding a tag
### Checklist

### TagStudio Version

Alpha 9.5.2

Windows 11

### Description

When adding a tag when the tag panel has an existing search, the results get reset to the default search, even though the search bar still has a search term. Adding and then removing a space triggers another search, returning them to the expected state.

### Expected Behavior

Upon adding a tag, the search results shouldn't be reset, and should likely trigger a search so that the results are up to date with the new tag that was added.

### Steps to Reproduce

1. Search something in the tags menu
2. Add a new tag

### Logs

_No response_

purpletennisball:
### Summary

Before when opening the Tag Manager, searching for a tag, and then creating a new tag the search results would be refreshed without taking account for the search text (Meaning the search results would be reset). Now instead it will refresh the search results with the search text.

Closes #932 

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/qt/modals/tag_database.py
+++ b/src/tagstudio/qt/modals/tag_database.py
@@ -51,7 +51,7 @@ class TagDatabasePanel(TagSearchPanel):
                     alias_ids=panel.alias_ids,
                 ),
                 self.modal.hide(),
-                self.update_tags(),
+                self.update_tags(self.search_field.text()),
             )
         )
         self.modal.show()


CyanVoxel (approved):
Everything seems to check out, thank you so much for the fix!

PR #940, opened 2025-05-27
fix: add parent tags to `folders_to_tags` macro and start tagging at root folder

rsazra:
### Summary

2 main things are changed for the `folders_to_tags` macro.
1. Tagging starts with folders in the root of the TagStudio directory (i.e. on the same level as .TagStudio). 
2. Parent tags are assigned.

Tested on my library of \~850 images, nested up to 3 layers of folders deep.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/qt/modals/folders_to_tags.py
+++ b/src/tagstudio/qt/modals/folders_to_tags.py
@@ -44,14 +44,14 @@ class BranchData:
 
 def add_folders_to_tree(library: Library, tree: BranchData, items: tuple[str, ...]) -> BranchData:
     branch = tree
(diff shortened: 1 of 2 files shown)


Computerdores (approved):
Works on x86_64 Windows 10. Code looks good too (that is to say your changes look to be fine, while that part of the code base desperately needs a refactor)

Issue #463, opened 2024-09-06
[Feature Request]: Press enter when adding fields


Cool-Game-Gev:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

When adding a new field to a file, add the ability to press enter and add the field without pressing the `Add` button.

### Solution

Allow the user to press enter when the `Add Field` window is open, which does the same as pressing the `Add` button. 

### Alternatives

The user could use the cursor to press the button, but this is undesirable for 2 reasons:

1. The cursor is slow, and most users prefer to use their keyboard
2. Many other programs allow the user to press enter in similar menus, so users assume it's a feature

CyanVoxel:
Added in #941

PR #941, opened 2025-05-27
feat: Press enter when adding fields

rsazra:
### Summary

As requested in https://github.com/TagStudioDev/TagStudio/issues/463.

I can only test on macOS, where the Return key is sent. I assume on other platforms it would Enter. This covers both, but needs to tested on Linux/Windows. 

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality


Computerdores (approved):
Works on x86_64 Windows 10. Code looks good too

PR #942, opened 2025-05-29
perf: Optimize db queries for preview panel

TheBobBobs:
### Summary

Add method `get_tag_hierarchy` to `Library` that returns a `dict[TagId, list[ParentId]]` and `dict[TagId, Tag]` containing every tag in the hierarchy. Update `FieldContainers` to use this new method.

Currently all places that call `Library.tag_display_name` already have the `Tag` object so we don't need to fetch it from db.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [X] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [X] Basic functionality


CyanVoxel (approved):
I apologize for getting around to this one so late. Everything seems to be working as intended and I'm glad that the tag hierarchy logic is now finally back into its own method. Thank you for working on this, and great changes!

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1490,6 +1490,36 @@ def get_tag_color(self, slug: str, namespace: str) -> TagColorGroup | None:
...
+
+            statement = select(Tag).where(Tag.id.in_(all_tag_ids))
+            statement = statement.options(
+                noload(Tag.parent_tags), selectinload(Tag.aliases), joinedload(Tag.color)
+            )
+            tags = session.scalars(statement).fetchall()
+            for tag in tags:
+                all_tags[tag.id] = tag
+            for tag in all_tags.values():
+                tag.parent_tags = {all_tags[p] for p in all_tag_parents.get(tag.id, [])}
```
CyanVoxel: Returning Tag objects with overwritten parent data resulted in the conflict issue propagating further in the UI stack, including flipping the tag's parents with its children in the BuildTagPanel due to the UI sharing that same "tainted" Tag object. While this *shouldn't* be an issue when the logic if flipped to match `main`, it still presents the question of whether or not this may mask further issues in the future.

TheBobBobs: Not manually setting fields on db objects would definitely be better.
Later this week I'll try experimenting with using joins for this data to see if it can be done without a performance difference.

After some testing I also found out this affects the state tracking sqlalchemy does. Which was causing issues when trying to save tag changes when the edit ui was opened from the preview panel. 
The solution is pretty simple but I couldn't find any documentation for the fix I did.

CyanVoxel: Sounds good! I'll go ahead and merge this PR in to main to get the rest of these changes live

PR #945, opened 2025-06-04
feat: Option to change tag click behaviour 

### LINKED ISSUE #944: [Feature Request]: Option to change tag click behaviour
### Checklist

### Description

Currently clicking on a Tag in the Preview Panel open the edit dialog for that Tag, previously this caused a search for that tag.

### Solution

Add a global setting to choose which behaviour should be used:
- open edit dialog
- search for Tag
- append tag to search

### Alternatives

_No response_

Computerdores:
### Summary

Resolves #944.

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality


CyanVoxel:
Looks good! Just needs some base strings in `en.json` and I think this should be good to go. Other than that, it *would* be ideal if the "Add to Search" functionality didn't add tags to searches they already exist in (thus killing the search), but I understand that since it's using a string manipulation approach that this would be a bit messy to tack that onto.

Computerdores:
Ok TIL that `A AND A` doesn't return anything that is definitely a bug.
As for the strings, I totally forgot to add those will do that really quick.

Computerdores:
I have added the missing strings and also found and fixed a bug because of which the settings dropdowns were always english even if the user set TS to be in some other language

PR #946, opened 2025-06-04
feat: custom modal making use of QDateTimeEdit

### LINKED ISSUE #943: [Feature Request]: Proper Edit Dialog for Datetime fields
### Checklist

### Description

Currently datetime fields can only be edited by typing them as a string in a specific format. For most users this is quite cumbersome compared to other applications.

### Solution

Generally it would be more desirable to have a proper date picker, while retaining good keyboard navigation.

Computerdores:
### Summary
Resolves #943.

### Tasks Completed
-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/qt/widgets/datetime_picker.py
+++ b/src/tagstudio/qt/widgets/datetime_picker.py
@@ -1,17 +1,42 @@
+import typing
 from collections.abc import Callable
 from datetime import datetime as dt
 from typing import cast
 
-from PySide6.QtCore import QDateTime, QLocale
+from PySide6.QtCore import QDateTime
 from PySide6.QtWidgets import QDateTimeEdit, QVBoxLayout
 
 from tagstudio.qt.widgets.panel import PanelWidget
 
+if typing.TYPE_CHECKING:
+    from tagstudio.qt.ts_qt import QtDriver
+
 DATETIME_FORMAT = "%Y-%m-%d %H:%M:%S"
 
 
+QDTF2DTF = {
+    "%d": "dd",
+    "%m": "MM",
+    "%y": "yy",
+    "%H": "HH",
+    "%M": "mm",
+    "%S": "ss",
+    "%Y": "yyyy",
+    "%I": "hh",
+    "%p": "AP",
+    "%x": "MM/dd/yy",
+}
+
+
+def qdtf2dtf(dtf: str) -> str:
+    out = dtf
+    for old, new in QDTF2DTF.items():
+        out = out.replace(old, new)
+    return out
+
+
 class DatetimePicker(PanelWidget):
-    def __init__(self, datetime: dt | str):
+    def __init__(self, driver: "QtDriver", datetime: dt | str):
         super().__init__()
         self.root_layout = QVBoxLayout(self)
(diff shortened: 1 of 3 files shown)


src/tagstudio/qt/widgets/text_line_edit.py
```diff
@@ -1,8 +1,3 @@
-# Copyright (C) 2025 Travis Abendshien (CyanVoxel).
-# Licensed under the GPL-3.0 License.
-# Created for TagStudio: https://github.com/CyanVoxel/TagStudio
-
-
```
CyanVoxel: Why the removal of the copyright and license here?

src/tagstudio/qt/widgets/datetime_picker.py
```diff
@@ -0,0 +1,61 @@
...
+        self.datetime_edit = QDateTimeEdit()
+        self.datetime_edit.setCalendarPopup(True)
+        self.datetime_edit.setDateTime(DatetimePicker.dt2qdt(datetime))
+        # sketchy way to show seconds without showing the day of the week;
+        # while also still having localisation
+        self.datetime_edit.setDisplayFormat(
+            QLocale.system()
+            .dateTimeFormat(QLocale.FormatType.ShortFormat)
+            .replace("HH:mm", "HH:mm:ss")
+        )
```
CyanVoxel: I feel the format here should mirror the datetime format in the user's TagStudio settings rather than the system locale. `QLocale.system()` also fails to recognize that I use 24 hour time on my system since it's technically not a part of my locale. While the date format that Qt uses is different, I *believe* it is possible to match equivalent formatting, it would just be a matter of figuring out which formats match such as `%m` -> `MM` and bringing over the delimiters.

Hardcoded example using `self.datetime_edit.setDisplayFormat("dd.MM.yyyy HH:mm:ss")`:
Screenshot 2025-06-05 at 17 24 06

As a side note, which may not be relevant anymore, on macOS I found that `QLocale.system().dateTimeFormat(QLocale.FormatType.ShortFormat)` evaluates to `M/d/yy h:mm  Ap` which bypaces the `.replace()` method here and doesn't show the seconds.
This is corrected by appending `.replace("h:mm", "hh:mm:ss")` (`h:mm:ss` also works).

src/tagstudio/qt/widgets/text_line_edit.py
Computerdores: Oops, only meant to remove the copyright part, will fix that (as for why I removed the copyright notice, see our [convo on discord](https://discord.com/channels/1229183630228848661/1232037962116759603/1327278110969692290))

src/tagstudio/qt/widgets/datetime_picker.py
Computerdores: I have adressed this in 9effa9d

CyanVoxel (approved):
Awesome work on this, thank you so much!

PR #679, opened 2025-01-02
feat: optimise AND queries

Computerdores:
Changes the way that ANDs are converted to SQL. This now uses less subqueries which has lead to a search time decrease when searching in TS.

My benchmark search previously took ~1min in TS and roughly ~46s when running the generated SQL to an external tool.
With these improvements the benchmark search now takes ~36s in TS and roughly ~1s in the external tool. This means there are further performance gains possible by optimising other parts of the search code.

Note:
- Currently the tag ids are hardcoded into the main query; I also tried directly querying these in the main query instead, but this only lead to a ~6s improvement while breaking some tests so I will not pursue this for now (should someone be interested in how I did that, checkout this commit 2615e7d from #673 that was reverted there)

PR #948, opened 2025-06-05
perf: optimize sql for or queries

TheBobBobs:
### Summary
Currently there is one check for each term in an OrList. This change will extract the tags out of the OrList and put them in one check which is then combined with the remaining terms in the OrList.

Parent tags that expand to many children tags also benefit from this optimization.

### Tasks Completed
-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [X] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [X] Basic functionality

### DIFF
--- a/src/tagstudio/core/library/alchemy/visitors.py
+++ b/src/tagstudio/core/library/alchemy/visitors.py
@@ -14,6 +14,7 @@ from tagstudio.core.library.alchemy.joins import TagEntry
 from tagstudio.core.library.alchemy.models import Entry, Tag, TagAlias
 from tagstudio.core.media_types import FILETYPE_EQUIVALENTS, MediaCategories
 from tagstudio.core.query_lang.ast import (
+    AST,
     ANDList,
     BaseVisitor,
     Constraint,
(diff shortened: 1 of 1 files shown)


src/tagstudio/core/library/alchemy/visitors.py
```diff
@@ -58,42 +59,15 @@ def __init__(self, lib: Library) -> None:
...
-        if len(tag_ids) > 1:
+        tag_ids, bool_expressions = self.__separate_tags(node.terms, only_single=True)
+        if len(tag_ids) > 0:
             bool_expressions.append(self.__entry_has_all_tags(tag_ids))
-        # If there is just one tag id, check the normal way
-        elif len(tag_ids) == 1:
-            bool_expressions.append(
-                self.__entry_satisfies_expression(TagEntry.tag_id == tag_ids[0])
-            )
-
```
Computerdores, line 96: Why did you remove this?

Computerdores (commented):
Works for me and runs on [this](https://github.com/TagStudioDev/TagStudio/pull/679#issuecomment-2585459596) library in ~13s instead of ~27s.

src/tagstudio/core/library/alchemy/visitors.py
TheBobBobs: In my testing using `__entry_has_all_tags` was slightly faster with one tag id. Sqlite will optimize out the group_by when just one tag_id is present. this can be checked with the query plan
```py
rows = session.execute(text(f"EXPLAIN QUERY PLAN {query_full}")).fetchall()
for row in rows:
    print(row)

```

TheBobBobs:
> Works for me and runs on [this](https://github.com/TagStudioDev/TagStudio/pull/679#issuecomment-2585459596) library in ~13s instead of ~27s.

This query takes 3 minutes for me on main.
Have you by chance created a custom index on tag_entries with columns `(entry_id, tag_id)`?

CyanVoxel:
> Works for me and runs on [this](https://github.com/TagStudioDev/TagStudio/pull/679#issuecomment-2585459596) library in ~13s instead of ~27s.

When using the "show dolor" query on this library, the time goes from around ~18.5 seconds to .18 - nice work on this!

PR #950, opened 2025-06-07
doc: Add QT MVC structure to style guide

Computerdores:
### Summary
Splits the style guide into a separate file, and documents the intended future structure of the qt part of the code base.

### Tasks Completed
Not applicable

PR #952, opened 2025-06-07
refactor(preview_panel): mvc split

Computerdores:
### Summary
Splits the current preview panel into View and Controller in accordance with the proposal in #950.

### Tasks Completed
-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality


Computerdores:
Should be ready for review now

pyproject.toml
```diff
@@ -84,7 +84,8 @@ qt_api = "pyside6"
 
 [tool.pyright]
 ignore = [".venv/**"]
-include = ["src/tagstudio/**"]
+include = ["src/tagstudio", "tests"]
+extraPaths = ["src/tagstudio", "tests"]
```
CyanVoxel, line 88: I'm curious what the need for the paths duplicated in the `extraPaths` field was

CyanVoxel (commented):
Great work on this! Everything looks good for the most part, just found one bug with the preview panel - when updating the name of a tag, visually it doesn't update and instead throws this error:
```
  File "/Users/cyanvoxel/Files/GitHub/TagStudio/src/tagstudio/qt/view/widgets/preview_panel_view.py", line 145, in set_selection
    if len(selected) == 0:
       ^^^^^^^^^^^^^
TypeError: object of type 'bool' has no len()
```

pyproject.toml
Computerdores: Not quite sure about the details any more, but these changes were needed to get refactoring and usage searches to work in VS Code. Otherwise they would miss the tests entirely

Computerdores:
fixed!

CyanVoxel (approved):

PR #953, opened 2025-06-10
refactor: offload thumbnail rendering to multiple processes

TheBobBobs:
### Summary
Renders can now be queued using the `ThumbnailManager` class which keeps a pool of processes to perform render jobs.

`CacheManager` has been rewritten. The caching folder layout has changed to `{hash[:2]}/{hash}-{mtime}.webp`
It does remove outdated cache items but doesn't support a max cache size.

### TODO
- Check that processes start and shutdown properly on all platforms
- Ensure rendering logic produces same results as current implementation
- Expand cache implementation?
- Cleanup code

### Tasks Completed
-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [X] Linux x86
    -   [ ] Linux ARM

-   Tested For:
    -   [X] Basic functionality

### DIFF
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -337,6 +337,7 @@ class Library:
                 return tag.name
 
     def open_library(self, library_dir: Path, storage_path: Path | None = None) -> LibraryStatus:
+        self.thumbnail_manager = ThumbnailManager(library_dir)
         is_new: bool = True
         if storage_path == ":memory:":
             self.storage_path = storage_path
@@ -353,8 +354,6 @@ class Library:
                         message="[JSON] Legacy v9.4 library requires conversion to v9.5+",
                         json_migration_req=True,
                     )
-
-        self.thumbnail_manager = ThumbnailManager(library_dir)
 
         return self.open_sqlite_library(library_dir, is_new)
 
--- a/src/tagstudio/qt/render.py
+++ b/src/tagstudio/qt/render.py
@@ -7,7 +7,7 @@ import math
 import os
 import struct
 import zipfile
-from concurrent.futures import ProcessPoolExecutor
+from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
 from copy import deepcopy
 from io import BytesIO
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


VasigaranAndAngel:
Cool!

Which module will handle loading the thumbnails? I want the thumbnail renderer with multiprocessing to be modular so i can generate a specific thumbnail when i wanted for different purposes. for example, i want to generate audio thumbnail as in the following video to use it in the music player widget.


https://github.com/user-attachments/assets/32459794-0a97-4b65-a9b0-15f5cab23cc8

TheBobBobs:
Create an instance of `RenderJob` and connect it's signals to your code like [here](https://github.com/TheBobBobs/TagStudio/blob/5e688ed824557c9da899c07f2508426855d617c1/src/tagstudio/qt/widgets/item_thumb.py#L213-L214).
You can then use the `ThumbnailManager` instance on `Library` to [queue](https://github.com/TheBobBobs/TagStudio/blob/5e688ed824557c9da899c07f2508426855d617c1/src/tagstudio/qt/ts_qt.py#L1665-L1667) a render passing in `RenderJob.on_finish` as the callback function.

CyanVoxel:
I'm a bit out of the loop on the purpose of this refactor, as the thumbnail renderer was already multi-threaded. I'm also not sure why the `CacheManager` was rewritten and no longer supports a maximum cache size

TheBobBobs:
With multi-threading the render threads and main thread have to share the [GIL](https://realpython.com/python-gil/#the-impact-on-multi-threaded-python-programs). The multi-thread approach is able to use more cpu still since the libraries used for image manipulation can do work without needing to aqcuire the GIL. But any python code or library code interacting with python objects will freeze the main ui thread.
The cache was rewritten to better work with multiple processes in mind. It is currently very basic but clearing old thumbnails when above `max_cache_size` shoud be easy to add.

Video showing the ui freeze when rendering thumbnails
`Multi-process left` `Multi-thread right`
https://github.com/user-attachments/assets/26fac00a-7dd3-48f4-ba8a-303bd7842ab5

PR #954, opened 2025-06-11
perf: Optimize page loading

TheBobBobs:
### Summary
Change `search_library` to only fetch `entry.id`.
Add `get_tag_entries` method to Library. Returns a `dict[tag_id, set[entry_ids]]`
Change `update_thumbs` to not fetch all tags for entries on page and instead use the above method to only check the tags it cares about.
move signal connection for `item_thumb` out of `update_thumbs` and into its init method.

This brings page load from 200ms->4ms

### Tasks Completed
-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [X] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [X] Basic functionality


CyanVoxel (approved):
I apologize for the big delay on getting to reviewing this.
Overall this seems to be a great improvement! We've been wanting to move (back) to referencing entries by ID ever since the change came into play in 9.5.

I'll go ahead and tidy up the merge conflict and get this merged. Thank you so much for your work on this!

PR #966, opened 2025-06-18
Wrong Date on Changelog fixed

ugurozturk:
Must see the git change list.


Computerdores (commented):
Good Catch! Although comparing it with the releases page it looks like it's not just the year but also month and day that are wrong, could you fix that as well?

PR #974, opened 2025-06-26
closes: #973 Completes traditional chinese translations

### LINKED ISSUE #973: [Feature Request]: Completion of tranditional chinese translations
### Checklist

### Description

`zh-Hant.json` has quite a few unfinished translations, as seen by the line count, `en.json` has about 300 lines, but `zh-Hant.json` is just over 200. for the benefit of traditional Chinese users, completing the translation is a QOL improvement.

### Solution

Looking at the existing translation written in `zh-Hant.json`, and comparing it to `en.json` and seeing which keys are not filled. Then using AI, translators, and personal experience, translating the rest of the untranslated text.

### Alternatives

None

ngivanyh:
### Summary
Completed translations of the traditional chinese translations. (i.e. modified `zh-Hant.json`)
There may be elements that differ from the one that already existed (e.g. 文庫 -> 文件庫)
Only one sentence hasn't been translated

### Tasks Completed: #973
-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [ ] Basic functionality


purpletennisball:
As far as I know translations should be done through the [Weblate Project](https://hosted.weblate.org/projects/tagstudio/). Although, I am not apart of the team for this project so I am not 100% sure

Computerdores:
To be honest your mention of using AI over in #973 makes me a bit weary, because we've had some bad experiences with people using AI to translate. I, for example, personally had to redo some translations that were technically correct translations for a specific word, but just completely incorrect in context (to the point where one wouldn't have understood what was meant). And with me having no idea about the chinese language, its not like I can double check them myself either.  It would probably be good if a native speaker or someone on a sufficient level of understanding could double check this.

(On the other hand one could argue that having translations at all would be better than none?)

purpletennisball:
> To be honest your mention of using AI over in #973 makes me a bit weary, because we've had some bad experiences with people using AI to translate.

Yeah I also didn't like the AI mention I think it's a total step in the wrong direction. Considering how LLM's occasionally just straight up lie I don't think it's a good idea. Plus, there is a reason that there is a WebLate for _people_ to contribute to.

ngivanyh:
Sorry, to reiterate. I PERSONALLY did all the translations, only in the end when there were a few stragglers did I use AI (and I checked with `zh-Hans.json`) (and yes I know there would be some phrasing differences between those who use traditional Chinese and those who use simplified Chinese) (I checked over the two different translations of AI and me in a diff file)

CyanVoxel:
Thank you for the translations - As purpletennisball mentioned, we normally use the [Weblate project](https://hosted.weblate.org/projects/tagstudio/) interface for adding translations. I'll go ahead and accept and merge this PR, but doing so introduces the risk of breaking the Weblate workflow unless the repo is properly locked on that side first (something I've made sure to do here). In the future, please contribute translations via Weblate.

PR #978, opened 2025-07-06
refactor(preview_thumb): mvc split

Computerdores:
### Summary

Splits the current preview thumb into View and Controller in accordance with the now accepted proposal in https://github.com/TagStudioDev/TagStudio/pull/950.

Note: This is based off #952 and will need to be <ins>**rebased**</ins> once that is merged.

Note: I preserved the change history for both `preview_thumb_controller.py` and `preview_thumb_view.py` via those first 3 commits, however this breaks if all commits are squashed. This means we can either not preserve the change history for those files and squash all of the commits in this PR, or we only squash the commits after that (I could do that myself, as it would require splitting the changes from the following commits and then merging after).

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality


Computerdores:
This is ready for review now (I would recommend waiting with the review until #952 is merged and this PR is rebased onto main)

CyanVoxel (approved):
I believe everything looks good here, thank you for your work on this!

Issue #979, opened 2025-07-06
[Feature Request]: Add thumbnail support for Krita documents (.kra)


FreeCap23:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

Krita documents should have icons.

<img width="437" height="258" alt="Image" src="https://github.com/user-attachments/assets/2e9acf0a-b1f8-4297-a244-dc4f30f8992b" />

Issue #379 is relevant.

Please let me know where I can upload the example files, if you need them.

### Solution

Here is how the previews look in Dolphin file manager.

<img width="387" height="186" alt="Image" src="https://github.com/user-attachments/assets/25dc91c5-4a41-4dff-8522-161ec2dd4914" />

Issue #977, opened 2025-07-06
[Feature Request]: Add thumbnail support for Krita and Open Raster files


Ambossmann:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

Krita (**.kra**, **.krz**) and Open Raster (**.ora**) files are currently not supported for thumbnail generation.

### Solution

Support thumbnail generation for these file types.
They are all standard zip files.
**.kra** and **.ora** usually contain a full resolution "mergedimage.png" at the top level. **.ora** files adhering to an older specification might not include this.

**.kra** and **.krz** should have a scaled down "preview.png" at the top level. It appears to be scaled so that the longer side is 256 pixels.
**.ora** has this file in "Thumbnails/thumbnail.png"

[Test.zip](https://github.com/user-attachments/files/21087828/Test.zip) contains the same image in each format. They are packed into a zip because github wouldn't let me attach them with their original endings.

CyanVoxel:
Completed in #985

PR #985, opened 2025-07-09
krita/open raster thumbs

mashed5894:
### Summary

Implemented krita (.kra/.krz) and open raster (.ora) thumbnails per #977 and #979.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/core/media_types.py
+++ b/src/tagstudio/core/media_types.py
@@ -327,6 +327,7 @@ class MediaCategories:
         ".odp",
         ".ods",
         ".odt",
+        ".ora",
     }
     _PACKAGE_SET: set[str] = {
         ".aab",


Pheubel:
I tried this PR on Windows, The krita files seem to work as expected, but i don't see a thumbnail for the open raster file that i exported with krita.

image

i wonder if the export is different in a way, i am personally not familiar with the format.

mashed5894:
> i wonder if the export is different in a way, i am personally not familiar with the format.

It follows the OpenDocument standard according to the documentation (https://www.openraster.org/baseline/file-layout-spec.html), so I simply added the extension to the list. You should be able to unzip your .ora file like any normal zip file and take a look at what the issue could be.

Issue #426, opened 2024-09-01
[Feature Request]: Double click on a tag to open "Edit Tag" window


Bamowen:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

In the Library Tag you currently need to right click > Edit to open the "Edit Tag" window, which is not very intuitive.

### Solution

It would be practical to have the option to open this "Edit Tag" window by double clicking on the desired Tag in the Library Tag.
![Screenshot from 2024-09-01 19-42-39](https://github.com/user-attachments/assets/10e70916-0467-4a6c-9f6f-bd429d69b77f)

mashed5894:
Hello! I took a quick look at this and it doesn't really seem like QPushButton is made for this purpose, and adding a double click timer might introduce some annoying delay into the normal functionality. An alternative solution might be to simply add an tiny edit button in the same style as the remove button.

Bamowen:
If indeed the double click is not the most elegant way to do it, adding a small button to open an edit button sounds fine.
Main point is to not hide this feature in a context menu (only), but have the option to quickly get to it :)

Computerdores:
@mashed5894 have you looked at [QWidget.mouseDoubleClickEvent](https://doc.qt.io/qtforpython-6/PySide6/QtWidgets/QWidget.html#PySide6.QtWidgets.QWidget.mouseDoubleClickEvent)?

rsazra:
Is there any reason a single click can't open the edit tag window? Clicking a tag just doesn't do anything right now, so why not add functionality to the regular, base interaction?

mashed5894:
> Is there any reason a single click can't open the edit tag window? Clicking a tag just doesn't do anything right now, so why not add functionality to the regular, base interaction?

It would make sense since the TagBox already does this, and it would feel bad from a ui design standpoint to make the same element have different behavior based on context.
[...]

PR #987, opened 2025-07-12
feat: click tags in tag library to edit

rsazra:
### Summary

After discussion in https://github.com/TagStudioDev/TagStudio/issues/426, it's been agreed a single left-click of a tag in the tab library should open the edit modal.

There is some discussion of a larger refactor, and while I think there is room for improvement with `TagWidget` and how it's used, this is more of a quick fix adding the functionality as it is currently intended.

I also removed all type errors and warnings in the file, though my techniques for doing so may not be the best. Would be happy to hear suggestions for improvement. These changes are all in the second commit if you'd like to see them (or the functional changes) in isolation.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/qt/modals/tag_search.py
+++ b/src/tagstudio/qt/modals/tag_search.py
@@ -264,8 +264,7 @@ class TagSearchPanel(PanelWidget):
                 self.scroll_layout.addWidget(new_tw)
 
         # Assign the tag to the widget at the given index.
-        tag_widget = self.scroll_layout.itemAt(index).widget()
-        assert isinstance(tag_widget, TagWidget)
+        assert isinstance(tag_widget := self.scroll_layout.itemAt(index).widget(), TagWidget)
         tag_widget.set_tag(tag)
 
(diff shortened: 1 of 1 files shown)


CyanVoxel:
I apologize for the delay on getting around to reviewing this -

It seems that the changes here affect both the "Tag Library" window as well as the "Add Tag" search panel, making it impossible to add tags to file entries as a result. These windows share the same underlying class so it was probably just an oversight.

As for the type fixes, they seem all fine and good although I'm a bit weary of using a walrus operator inside an assertion statement, since that's wrapping a necessary variable assignment inside a debug statement. Though it appears many of these type issues have already been fixed upstream, so no need to worry about it?

CyanVoxel:
Closing due to lack of activity

PR #990, opened 2025-07-14
Add Nix path to ffmpeg locations on macOS

thibmaek:
### Summary

`ffmpeg` was installed on my system using Nix (package manager, not NixOS) but undetected by TagStudio.
I added the path pointing to where Nix installs these binaries to the list of mac locations.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [ ] Basic functionality

### DIFF
--- a/src/tagstudio/qt/helpers/vendored/ffmpeg.py
+++ b/src/tagstudio/qt/helpers/vendored/ffmpeg.py
@@ -4,6 +4,7 @@
 
 import contextlib
 import json
+import os
 import platform
 import subprocess
 from shutil import which
@@ -20,7 +21,7 @@ FFMPEG_MACOS_LOCATIONS: list[str] = [
     "",
     "/opt/homebrew/bin/",
     "/usr/local/bin/",
-    f"/etc/profiles/per-user/{user}/bin"
+    f"/etc/profiles/per-user/{user}/bin",
 ]
 
 


xarvex:
Best approach is to check the PATH for `ffmpeg`, which I'm surprised isn't happening for macOS. This location is not guaranteed to be used on all Nix installations, other paths are used instead, especially when factoring in nix-darwin and home-manager.

Not checking the PATH for ffmpeg also means that installing TagStudio with Nix (which should be possible as I account for darwin, but I do not have a Mac to test) might break.

CyanVoxel:
> Best approach is to check the PATH for `ffmpeg`, which I'm surprised isn't happening for macOS. This location is not guaranteed to be used on all Nix installations, other paths are used instead, especially when factoring in nix-darwin and home-manager.
> 
> Not checking the PATH for ffmpeg also means that installing TagStudio with Nix (which should be possible as I account for darwin, but I do not have a Mac to test) might break.

Just for some documented clarification on this: macOS applications launched from the GUI (i.e. launched outside of a terminal) don't have access to the system PATH. That's why for FFmpeg we had to start keeping a list of all commonly known install locations.

@thibmaek Any updates on the failing workflow checks? I could provide some assistance if needed.

thibmaek:
@CyanVoxel forgot to import `os` since I edited the file with Neovim 😅 
Should pass now!

Issue #997, opened 2025-07-23
[Bug]: Parent and child tag IDs are swapped in the parent_tags DB table


HeikoWasTaken:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Alpha 9.5.2

### Operating System & Version

Arch Linux (Kernel: 6.15.5-arch1-1)

### Description

While comparing the `tags` and `tag_parents` tables within the `ts_library.sqlite` database, it appears that the parent and child tag IDs are reversed in the `tag_parents` table. This seems to happen with user-defined tags as well as the default `Archived`, `Favourite` and `Meta Tags` tags.

What's currently happening:
Child tag ID -> `parent_id` field
Parent tag ID -> `child_id` field

### Expected Behavior

Parent and child tag IDs should be assigned to the corresponding fields within the `tag_parents` table.

What _should_ be happening:
Parent tag ID -> `parent_id` field
Child tag ID -> `child_id` field

### Steps to Reproduce

1. Create a new tag.
2. Create a second tag and set the first tag as a parent.
3. Open the `ts_library.sqlite` database and observe the records in the `tags` table.
4. Compare the tag IDs to to those in the `tag_parents` table.

### Logs

<details>
  <summary>Tag Editor</summary>
  <img width="1271" height="692" alt="Image" src="https://github.com/user-attachments/assets/2ce6aeff-6c46-4de0-86d5-f3e725bcf8e1" />
</details>

<details>
  <summary>Database</summary>
  
  ```
  ts_library.sqlite> SELECT * FROM tag_parents;
  +-----------+----------+
  | parent_id | child_id |
  +-----------+----------+
  | 0         | 2        |
  | 1         | 2        |
  | 1001      | 1000     |
  +-----------+----------+
  3 rows in set
  Time: 0.015s
  ts_library.sqlite> SELECT * FROM tags;
  +------+-----------+-----------+--------------------+------------+-------------+--------+-------------------+
  | id   | name      | shorthand | color_namespace    | color_slug | is_category | icon   | disambiguation_id |
  +------+-----------+-----------+--------------------+------------+-------------+--------+-------------------+
  | 0    | Archived  | <null>    | tagstudio-standard | red        | 0           | <null> | <null>            |
[...]

PR #998, opened 2025-07-23
Fix: Swapped IDs in parent_tags DB table

HeikoWasTaken:
### Summary

Stops child/parent tag IDs from being assigned to incorrect fields in the `ts_library.sqlite` db's `tag_parents` table as described in #997.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [ ] Basic functionality


Computerdores:
Good fix, but for this to be merged a migration for existing libraries would be needed (tests would be nice too).

@CyanVoxel I could have sworn to have reviewed or written a fix for this at some point, do you remember anything regarding that?

CyanVoxel:
> Good fix, but for this to be merged a migration for existing libraries would be needed (tests would be nice too).
> 
> @CyanVoxel I could have sworn to have reviewed or written a fix for this at some point, do you remember anything regarding that?

There were several issues with swapped parent/child/"subtag" language during the migration to 9.5, this may have slipped by during those changes. I'm also not sure if just swapping the assignment will have the intended effect or if there needs to be additional logic changes reflected, though testing will either confirm or deny that

HeikoWasTaken:
> Good fix, but for this to be merged a migration for existing libraries would be needed (tests would be nice too).
> 
> @CyanVoxel I could have sworn to have reviewed or written a fix for this at some point, do you remember anything regarding that?

Sorry if I'm talking over things that have already been discussed, I couldn't find anything related to this.

For the migration, I believe checking whether the `Archived` and `Meta Tag` tags are **in**correctly set as parent/child should be enough to catch libraries that were created after this bug was introduced. But I'm not sure about older libraries which may have a mix of correct and incorrect parent/child relationships.

How would valid joins be differentiated from the swapped ones?

Computerdores:
> > Good fix, but for this to be merged a migration for existing libraries would be needed (tests would be nice too).
> > @CyanVoxel I could have sworn to have reviewed or written a fix for this at some point, do you remember anything regarding that?
> 
> Sorry if I'm talking over things that have already been discussed, I couldn't find anything related to this.

All good, I was just under the impression that this had already been fixed, but apparently not.
 
> For the migration, I believe checking whether the `Archived` and `Meta Tag` tags are **in**correctly set as parent/child should be enough to catch libraries that were created after this bug was introduced. But I'm not sure about older libraries which may have a mix of correct and incorrect parent/child relationships.
> 
> How would valid joins be differentiated from the swapped ones?

Generally, I would suggest you increment [`LibraryPrefs.DB_VERSION`](https://github.com/TagStudioDev/TagStudio/blob/717690827438406f16a9545ce749e2ee43d70a62/src/tagstudio/core/enums.py#L87) and then add another migration similar to the [existing ones](https://github.com/TagStudioDev/TagStudio/blob/717690827438406f16a9545ce749e2ee43d70a62/src/tagstudio/core/library/alchemy/library.py#L473-L497).

HeikoWasTaken:
> Generally, I would suggest you increment [LibraryPrefs.DB_VERSION](https://github.com/TagStudioDev/TagStudio/blob/717690827438406f16a9545ce749e2ee43d70a62/src/tagstudio/core/enums.py#L87) and then add another migration similar to the [existing ones](https://github.com/TagStudioDev/TagStudio/blob/717690827438406f16a9545ce749e2ee43d70a62/src/tagstudio/core/library/alchemy/library.py#L473-L497).

Thanks, I'm not very familiar with the codebase so that's a big help. Will take a crack at it.

HeikoWasTaken:
I was initially concerned about such a case:
1. Library is created prior to bugged version
2. Some tags are parented to other tags (correctly)
3. Library is updated to bugged version
4. More tags are parented to other tags (incorrectly)

Wherein the database would contain a mixture of correctly parented tags as well as incorrect ones, with no clear way of telling them apart.

However after some digging, it seems like this bug has been around since the first implementation of the `tag_parents` table, meaning that _all_ records in `tag_parents` can be considered incorrect, and safely swapped back around without worring about false-positives (phew).

Just leaving this here for clarity.

Computerdores:
> Wherein the database would contain a mixture of correctly parented tags as well as incorrect ones, with no clear way of telling them apart.

In that case there would be no way to tell the correct and incorrect ones apart anyway so no need to worry about that on your end either way 😄

HeikoWasTaken:
Just need to write some tests now, then should be ready for review

HeikoWasTaken:
How are the test database fixtures [in here](https://github.com/TagStudioDev/TagStudio/tree/main/tests/fixtures/empty_libraries) created? Just blanks made by that specific version of tagstudio? Or is there some sort of tooling/procedure required to properly generate them?

Computerdores:
> How are the test database fixtures [in here](https://github.com/TagStudioDev/TagStudio/tree/main/tests/fixtures/empty_libraries) created? Just blanks made by that specific version of tagstudio? Or is there some sort of tooling/procedure required to properly generate them?

Not quite sure, @CyanVoxel would know best.
That said, I believe they were just created using some version of TS and then maybe modified in an external tool in some cases.

HeikoWasTaken:
>That said, I believe they were just created using some version of TS and then maybe modified in an external tool in some cases.

I went ahead and did it that way, as far as I can tell migration tests don't care about the database's contents anyway, only that it can be successfully opened by [`Library.open_library()`](https://github.com/TagStudioDev/TagStudio/blob/717690827438406f16a9545ce749e2ee43d70a62/src/tagstudio/core/library/alchemy/library.py#L334).

PR #1003, opened 2025-08-02
refactor(tag_box): mvc split

Computerdores:
### Summary
Splits the current preview thumb into View and Controller in accordance with the STYLE.md.

(Figured I would refactor this *before* adding features to this for another thing I am working on.)

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality


src/tagstudio/qt/controller/components/tag_box_controller.py
```diff
@@ -0,0 +1,91 @@
...
+if typing.TYPE_CHECKING:
+    from tagstudio.qt.ts_qt import QtDriver
+
+logger = structlog.get_logger(__name__)
+
+
+class TagBoxWidget(TagBoxWidgetView):
+    on_update = Signal()
+
+    __entries: list[int]
```
CyanVoxel: It might not hurt to initialize the value to an empty list here? Pyright reports a [reportUninitializedInstanceVariable](https://docs.basedpyright.com/v1.31.0/configuration/config-files/#reportUninitializedInstanceVariable) warning
```suggestion
    __entries: list[int] = []
```

src/tagstudio/qt/controller/components/tag_box_controller.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def __init__(self, title: str, driver: "QtDriver"):
+        super().__init__(title, driver)
+        self.__driver = driver
+
+    def set_tags(self, tags: Iterable[Tag], entries: list[int]) -> None:
```
CyanVoxel: I believe the `_set_tags()`, `_on_click()`, `_on_remove()`, `_on_edit()`, and `_on_search()` methods should have `@override` decorators

src/tagstudio/qt/controller/components/tag_box_controller.py
CyanVoxel: Previously the selected entries were referenced directly via the `driver`'s `selected` list. Is there a benefit to having the list be kept as a separate copy here? Wouldn't there be a fear of the lists getting out of sync?

CyanVoxel (commented):
Approved when comments are addressed 👍

src/tagstudio/qt/controller/components/tag_box_controller.py
Computerdores: TL;DR This makes it possible to use the component to apply the tags to an entry that isn't selected in the main window.

I am working on a quick tagging tool which is in an extra window and uses the tag box for applying tags to the currently shown entry. Since the shown entry has nothing to do with the main window this is required to make it work. The tag boxes there could be something like e.g. a Characters box with the tags "Mario" and "Luigi", a game box with tags for different nintendo games, etc. And then you can quickly apply common tags to the shown entry and then go to the next entry. I want to use this in combination with hotkeys for the shown tags to quickly tag a large library with a couple of base tags (a set of 10-20 different ones).

src/tagstudio/qt/controller/components/tag_box_controller.py
Computerdores: I applied the override decorators in the latest commit, but I am getting a weird mypy error that there is no such base method. I don't really know what to do about that at this point, since the inheritance is correct and the method signature literally copied.

Edit: adding `@abstractmethod` on the base class versions doesn't help either

src/tagstudio/qt/controller/components/tag_box_controller.py
CyanVoxel: Gotcha, sounds good then!

Some quick tagging panels have been something I've been long meaning to implement. I'll be excited to see where that's going!

src/tagstudio/qt/controller/components/tag_box_controller.py
CyanVoxel: Hmm, I remember something similar happening with some other files such as `item_thumb.py`:

```
@override
    def enterEvent(self, event: QEnterEvent) -> None:  # type: ignore[misc]
        self.show_check_badges(show=True)
        return super().enterEvent(event)
```
In this case we had to add `# type: ignore[misc]` to appease Mypy which didn't believe the override complaint from Pyright. What's weird here though is that removing this comment for me will trigger the Mypy warning inside my VSCode extension but it never did that for me when looking at the tag_box methods here so it completely flew by me. Actually as of writing this, attempting to add that comment to these methods now seems to update the file in such a way where Mypy is complaining about them here.

So perhaps having the `# type: ignore[misc]` line is the way to go until we eventually phase out Mypy completely in favor of Pyright...

Computerdores:
Should be good to go now

CyanVoxel (approved):

PR #1016, opened 2025-08-05
ui: add arrow icons for navigation buttons

CyanVoxel:
### Summary

This PR replaces the placeholder angle bracket symbols ("<", ">") on the navigation icons with proper icons (via [boxicons](https://v2.boxicons.com/)).

The icons respond to the light and dark theme settings, however I'm not sure what the best way would be to have them respond to a disabled QPushButton state or background window dimming, the later of which happens to text in windows on at least macOS for me and leaves the icons popping a bit much. Both of which seem like larger systemic problems that need to be solved with working with Qt, and likely outside the scope of this PR.

image

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality


Computerdores (approved):
Doesn't run out of the box because of #1017 but the PR itself looks good

PR #1024, opened 2025-08-07
flake: use nixpkgs

poseidon-rises:
### Summary
Use libraries from nixpkgs instead of using custom packages. This also removes the need for the 'tagstudio-jxl' package, and therefor it has been deprecated.

**Maintainer**: @xarvex 

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [ ] Basic functionality


poseidon-rises:
Also this seems to have the funny side effect of fixing the nix package, as it was broken before. I haven't completely tested it yet because it requires a compilation but it seems to pass the step which was throwing an error before this fix.

xarvex:
Sorry it took so long to get to this. Forgive me, but what is `pillow-jpls`? This doesn't seem to be the same dependency we use for JXL.

Our dependency: https://github.com/Isotr0py/pillow-jpegxl-plugin
`pillow-jpls`: https://github.com/planetmarshall/pillow-jpls

Didn't catch this when you were talking about finding the package on Discord. I'm afraid this might make this PR invalid.

poseidon-rises:
> Sorry it took so long to get to this. Forgive me, but what is `pillow-jpls`? This doesn't seem to be the same dependency we use for JXL.
> 
> Our dependency: https://github.com/Isotr0py/pillow-jpegxl-plugin `pillow-jpls`: https://github.com/planetmarshall/pillow-jpls
> 
> Didn't catch this when you were talking about finding the package on Discord. I'm afraid this might make this PR invalid.
 
Yea I think you're right; That's my bad. I can open a new PR to remove just the exiv2 package instead.

xarvex:
> I can open a new PR to remove just the exiv2 package instead.

If the package in nixpkgs that you are thinking of is `py3exiv2`, that is (confusingly) a different library than what is used here.

`pyexiv2` - https://github.com/LeoHsiao1/pyexiv2
`py3exiv2` - https://launchpad.net/py3exiv2

They are incompatible with each other and have different implementations.

poseidon-rises:
> If the package in nixpkgs that you are thinking of is `py3exiv2`, that is (confusingly) a different library than what is used here.
> 
> `pyexiv2` - https://github.com/LeoHsiao1/pyexiv2 `py3exiv2` - https://launchpad.net/py3exiv2
> 
> They are incompatible with each other and have different implementations.

Na it's even worse, `exiv2` is a different exiv2: https://github.com/jim-easterbrook/python-exiv2

Anyway, I'll close this out. Sorry for wasting your time.

xarvex:
No time wasted here, it's all good and I still appreciate the good-faith PR!

PR #1029, opened 2025-08-12
feat: add random sorting

TheBobBobs:
A basic implementation of a random sort. Uses a seed to allow for pagination without changing order. Clicking search will generate a new seed.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [X] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [X] Basic functionality

### DIFF
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -914,7 +914,7 @@ class Library:
                 case SortingModeEnum.PATH:
                     sort_on = func.lower(Entry.path)
                 case SortingModeEnum.RANDOM:
-                    sort_on = func.sin(Entry.id + search.random_seed)
+                    sort_on = func.sin(Entry.id * search.random_seed)
 
             statement = statement.order_by(asc(sort_on) if search.ascending else desc(sort_on))
             if page_size is not None:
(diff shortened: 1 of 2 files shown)


purpletennisball:
Tested this out and it seems all good but just a suggestion: Maybe have the seed change on refresh? (Cmd/Ctrl+R)

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -913,6 +913,8 @@ def search_library(
                     sort_on = func.lower(Entry.filename)
                 case SortingModeEnum.PATH:
                     sort_on = func.lower(Entry.path)
+                case SortingModeEnum.RANDOM:
+                    sort_on = func.sin(Entry.id + search.random_seed)
```
Computerdores: I am a bit skeptical about the randomness of this. I would have done it like [this](https://stackoverflow.com/a/1253576) (would also have a different order every time), on the other hand, the randomness of this isn't that important so as long as it is good enough this could be fine.

TheBobBobs: Random() doesn't support passing in a seed so changing pages would give a new order. Using sin() allows consistent ordering when moving between pages.

PR #1038, opened 2025-08-19
fix(search): pass current `BrowsingState` to `from_tag_id()`

### LINKED ISSUE #1031: [Bug]: Searching with internal tag ids ignores sorting order
### Checklist

### TagStudio Version

9.5.3

Win11

### Description

Its seems the new version added a new quirk with searching  is that, for  whatever reason searching with internal tag ids (using the Rclick drop down mainly) makes the results ignore the instructed tag order and only sorts with  ascending order. if yours is not set to this, it will  ignore it unless you set it to ascending and back to descending again. (however this only works temporarily)

### Expected Behavior

The results always  show in the instructed tag order (ascend, descend)

### Steps to Reproduce

Do any search with the internal tag ids (yes it will break if you do it with regular  tag names  intermixed too)

### Logs

_No response_

CyanVoxel:
### Summary

This PR fixes `BrowsingState` objects created via the `from_tag_id()` not respecting the current sorting options by allowing the method to receive an optional `BrowsingState` object to use the current sorting information from and passing the current `BrowsingState` in all instances of this method call. Fixes #1031.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/core/library/alchemy/enums.py
+++ b/src/tagstudio/core/library/alchemy/enums.py
@@ -97,7 +97,24 @@ class BrowsingState:
         return cls(query=search_query)
 
     @classmethod
-    def from_tag_id(cls, tag_id: int | str) -> "BrowsingState":
+    def from_tag_id(
+        cls, tag_id: int | str, state: "BrowsingState | None" = None
+    ) -> "BrowsingState":
+        """Create and return a BrowsingState object given a tag ID.
+
+        Args:
+            tag_id(int): The tag ID to search for.
+            browsing_state(BrowsingState|None): An optional BrowsingState object to use
+                existing options from, such as sorting options.
+
+        """
+        logger.warning(state)
+        if state:
+            return (
+                cls(query=f"tag_id:{str(tag_id)}")
+                .with_sorting_direction(state.ascending)
+                .with_sorting_mode(state.sorting_mode)
+            )
         return cls(query=f"tag_id:{str(tag_id)}")
 
     @classmethod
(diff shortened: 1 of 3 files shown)


src/tagstudio/core/library/alchemy/enums.py
```diff
@@ -97,7 +97,24 @@ def from_search_query(cls, search_query: str) -> "BrowsingState":
...
+                existing options from, such as sorting options.
+
+        """
+        logger.warning(state)
+        if state:
+            return (
+                cls(query=f"tag_id:{str(tag_id)}")
+                .with_sorting_direction(state.ascending)
+                .with_sorting_mode(state.sorting_mode)
+            )
```
Computerdores: I think it would be better to only replace the query on the existing state:
```suggestion
            return state.with_search_query(f"tag_id:{str(tag_id)}")
```

src/tagstudio/core/library/alchemy/enums.py
```diff
@@ -97,7 +97,24 @@ def from_search_query(cls, search_query: str) -> "BrowsingState":
...
         return cls(query=f"tag_id:{str(tag_id)}")
```
Computerdores: I would prefer not duplicating the query string building, since it is kinda hacky and should imo be replaced at some point (would be better to build a syntax tree and stringify that, but the latter isn't implemented yet)

CyanVoxel:
@Computerdores Thank you for the review! Changes implemented 👍

PR #1039, opened 2025-08-19
refactor: make cache_manager thread safe

TheBobBobs:
### Summary
Fixes errors from multiple render threads adding/removing cache folders.
Fixes `register_existing_folders` only calculating the size for the last folder in the cache.

### Tasks Completed

-   Platforms Tested:
    -   [X] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [X] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [X] Basic functionality


src/tagstudio/core/library/alchemy/library.py
```diff
@@ -212,6 +213,8 @@ class Library:
     folder: Folder | None
     included_files: set[Path] = set()
 
+    cache_manager: CacheManager
+
```
CyanVoxel: This file (`src/tagstudio/core/library/alchemy/library.py`) is a backend file for the core library, while the `CacheManager` is a frontend system used for caching data for use by the Qt frontend, making it inappropriate to move here from the main Qt frontend file (`ts_qt.py`)

src/tagstudio/qt/cache_manager.py
```diff
@@ -2,189 +2,119 @@
...
-                    # FileNotFound errors due to trying to delete "._" files that have
-                    # already been deleted: https://bugs.python.org/issue29699
-                    with contextlib.suppress(FileNotFoundError):
-                        file.unlink()
+class CacheManager:
+    def __init__(self, library_dir: Path, max_folder_size_mb: int = 10, max_size_mb: int = 500):
+        self._lock = RLock()
+        self.cache_folder = library_dir / TS_FOLDER_NAME / THUMB_CACHE_NAME
+        self.max_folder_size = max_folder_size_mb * 1000 * 1000
+        self.max_size = max(max_size_mb, max_folder_size_mb) * 1000 * 1000
```
CyanVoxel: Why multiply it by two numbers in a row here? Why not by a single value (preferably a named value over magic numbers)

TheBobBobs:
This should now work with random files in the cache folder. As well as cache folders being deleted outside of tagstudio.
I ended up just removing the bytes to megabytes conversion.

src/tagstudio/qt/widgets/thumb_renderer.py
```diff
@@ -1414,29 +1393,9 @@ def fetch_cached_image(folder: Path):
                 mod_time = str(filepath.stat().st_mtime_ns)
             hashable_str: str = f"{str(filepath)}{mod_time}"
             hash_value = hashlib.shake_128(hashable_str.encode("utf-8")).hexdigest(8)
+            file_name = Path(f"{hash_value}{ThumbRenderer.cached_img_ext}")
+            image = fetch_cached_image(file_name)
 
-            # Check the last successful folder first.
-            if ThumbRenderer.last_cache_folder:
-                image = fetch_cached_image(ThumbRenderer.last_cache_folder)
```
CyanVoxel, line 1424: Why remove the last folder check? Now each time a cached image needs to be fetched, it is now mandatory to perform a sort operation  on the folder keys in `cache_manager.py`, then has linear search through up to every cache folder with a construction of a Path object through each iteration.
```py
for folder in sorted(self.folders.keys()):
    file_path = folder / file_name
    if file_path.exists():
        return file_path
```

The step of checking the last successful folder was a heuristic that was able to significantly improve performance by skipping these steps in many cases, as thumbnails that are generated on the same page are most likely stored in the same folder on disk.

TheBobBobs: Whoops. The plan was to move it to `CacheManager` but I forgot all about it.

I've now changed the folders field on `CacheManager` from a dictionary to a list sorted by last use. This should greatly reduce the number of folder checks.

PR #1041, opened 2025-08-20
perf: optimize refresh_dir by prefetching all paths

TheBobBobs:
### Summary
Instead of fetching each entry path one by one prefetch all of them into a set for fast checks.
Refresh time on a library with 400_000 entries went from `160s` to `13s`.

### Tasks Completed
-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [X] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [X] Basic functionality


CyanVoxel:
#897 also completely changes this area of the code and will cause this to conflict

TheBobBobs:
Replaced by #1042 which is based on #897

PR #1042, opened 2025-08-20
perf: optimize refresh_dir by prefetching all paths

TheBobBobs:
### Summary
Instead of fetching each entry path one by one prefetch all of them into a set for fast checks.
Refresh time on a library with 400_000 entries went from `160s` to `13s`.

### Tasks Completed
-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [X] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [X] Basic functionality

### DIFF
--- a/src/tagstudio/core/utils/refresh_dir.py
+++ b/src/tagstudio/core/utils/refresh_dir.py
@@ -56,6 +56,10 @@ class RefreshDirTracker:
 
         ignore_patterns = Ignore.get_patterns(library_dir)
 
+        if len(self.library.included_files) == 0:
+            paths = set(library_dir / p for p in self.library.all_paths())
+            self.library.included_files = paths
+
         if force_internal_tools:
             return self.__wc_add(library_dir, ignore_to_glob(ignore_patterns))
 
@@ -92,6 +96,7 @@ class RefreshDirTracker:
                         "--hidden",
                         "--ignore-file",
                         f'"{str(compiled_ignore_path)}"',
+                        str(library_dir),
                     ]
                 ),
                 cwd=library_dir,
@@ -115,8 +120,8 @@ class RefreshDirTracker:
         dir_file_count = 0
         self.files_not_in_library = []
 
-        for r in dir_list:
-            f = pathlib.Path(r)
+        for p in dir_list:
+            f = Path(p)
 
             end_time_loop = time()
             # Yield output every 1/30 of a second
@@ -136,8 +141,10 @@ class RefreshDirTracker:
             dir_file_count += 1
             self.library.included_files.add(f)
 
-            if not self.library.has_path_entry(f):
-                self.files_not_in_library.append(f)
+            relative_path = f.relative_to(library_dir)
+
+            if not self.library.has_path_entry(relative_path):
+                self.files_not_in_library.append(relative_path)
 
         end_time_total = time()
         yield dir_file_count
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -7,7 +7,7 @@ import re
 import shutil
 import time
 import unicodedata
-from collections.abc import Iterable, Iterator
+from collections.abc import Iterable, Iterator, Sequence
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from os import makedirs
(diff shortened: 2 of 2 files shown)


CyanVoxel:
So in my testing this appears to give mixed results depending on whether `wcmatch` or `ripgrep` is being used. For these tests I used one folder for `ripgrep` across both this PR and the ts_ignore PR (#897) that has ~60k files that are spread across an internal SSD and symlinked HDD NAS. For `wcmatch` I had to use a different folder due to it otherwise needing to take around 30 minutes to scan the same library due to `wcmatch` not skipping ignored directories in the same way that `ripgrep` does, so I instead used a different ~120k file external SDD for the `ripgrep` tests.

```
#897 - ripgrep, ~60k Files [NAS, HDD+SSD] (after .ts_ignore)
New Library:
0:13
0:20
0:12
0:20
On Refresh (Same Session):
<0:00
<0:00
<0:00
```

```
#897 - wcmatch, ~112k Files [EXT. SSD] (after .ts_ignore)
New Library:
1:02
1:06
1:03
On Refresh (Same Session):
0:32
0:31
0:31
```

```
#1042 - ripgrep, ~60k Files [NAS, HDD+SSD] (after .ts_ignore)
New Library:
2:04
1:18
1:18
On Refresh (Same Session):
1:39
1:14
1:21
```

```
#1042 - wcmatch, ~112k Files [EXT. SSD] (after .ts_ignore)
New Library:
0:39
0:39
0:36
On Refresh (Same Session):
0:29
0:29
0:29
0:29
```

Overall I found that this PR made the `ripgrep` scans significantly slower to scan and subsequently refresh, while the `wcmatch` scans were about ⅓ faster for initial scans with no notable improvement for subsequent refreshes, but nothing on the scale of the results in your description. I believe it's also worth noting that `Path` operations are *very* expensive and contribute to most of the time spent during these types of loops. The reason the `library.included_files` set was used previously is to avoid any path operations when checking if a path was already included in the library during the same session.

TheBobBobs:
I wasn't able to reproduce the poor performance with ripgrep on Windows or Linux but this change should be faster regardless of OS. Also fixes the `is_dir` check in `__rg_add` using relative paths by making ripgrep output absolute paths.

src/tagstudio/core/utils/refresh_dir.py
```diff
@@ -92,6 +96,7 @@ def __get_dir_list(self, library_dir: Path, ignore_patterns: list[str]) -> list[
                         "--hidden",
                         "--ignore-file",
                         f'"{str(compiled_ignore_path)}"',
+                        str(library_dir),
```
CyanVoxel, line 99: This needs to be escaped by quotes in order to work with paths that have spaces in their name, see the string above
```suggestion
                        f'"{str(library_dir)}"',
```

CyanVoxel (commented):
Testing with the non-quoted path fixed, this seems to have only exacerbated the `ripgrep` slowdown for me:
```
#897 - ripgrep, ~ 60k Files [NAS, HDD+SSD] (after .ts_ignore)
New Library:
0:13
0:20
0:12
0:20
```
```
#1042 - ripgrep, ~ 60k Files [NAS, HDD+SSD] (after .ts_ignore)
New Library:
3:08
2:26
```

Also for future reference please do not force push your branch during a review, there's a callout about that in the [CONTRIBUTING.md](https://github.com/TagStudioDev/TagStudio/blob/main/CONTRIBUTING.md) since it makes reviewing PRs more difficult

CyanVoxel:
I did a followup test with purely my external SSD instead of the mixed NAS and found some more strange results:
```
#897 - ripgrep, ~ 112k Files [EXT. SSD] (after .ts_ignore)
0:24
0:24
```
```
#1042 - ripgrep, ~ 112k Files [EXT. SSD] (after .ts_ignore)
0:28
0:27
```
While this is still slower, the difference isn't nearly as severe as it is with my NAS library. Which is especially strange given that `ripgrep` is providing the list of paths first, *before* the loop... Either way, something isn't quite right here

TheBobBobs:
Oh it's probably the fixed is_dir check causing a bunch of network requests it wasn't doing before.

TheBobBobs:
Also this last change only optimizes existing libraries. If there are lots of new files in the scan performance should be pretty much the same. Sorry forgot to mention that.
If this ends up working for you I'll add a check that lets scans with lots of new files benefit from this optimization.

PR #1047, opened 2025-08-23
fix(pyproject): allow Python 3.13 to function

xarvex:
### Summary

Python 3.13 doesn't seem to have any trouble from running TagStudio, apart from `audioop` being removed/deprecated. While we do not support outside package managers at the moment, more and more systems will be moving to it as the system interpreter, and simply allowing this doesn't cause any harm.

`pydub` is what currently depends on the removed `audioop` module. For now, added in `audioop-lts` (only when on a version at or above 3.13) while a long-term solution is figured out.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -28,7 +28,7 @@
       perSystem =
         { pkgs, ... }:
         let
-          python3 = pkgs.python312;
+          python3 = pkgs.python313;
         in
         {
           packages =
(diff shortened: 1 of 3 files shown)


xarvex:
Rebased to account for Nix now that #1048 was merged.
Edit: went a bit too fast, added in the `audioop-lts`.

Issue #988, opened 2025-07-13
[Bug]: Nix package fails due to Qt version mismatch


CommentatorForAll:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Latest git (rev 7176908)

### Operating System & Version

NixOS 25.05

### Description

i tried to run the project both with `nix run github:TagStudioDev/TagStudio#tagstudio` and `nix run` within the cloned repository, resulting in the following error messages:
```
2025-07-13 16:51:27 [error    ] [ThumbRenderer] Could not import the "pillow_jxl" module
Traceback (most recent call last):
  File "/nix/store/r3hf69imbw20w923x4ij5xdl9saz16ss-TagStudio-9.5.2/lib/python3.12/site-packages/tagstudio/qt/widgets/thumb_renderer.py", line 82, in <module>
    import pillow_jxl  # noqa: F401 # pyright: ignore[reportUnusedImport]
    ^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'pillow_jxl'
```
and later
```
Cannot mix incompatible Qt library (6.9.1) with this library (6.9.0)
[1]    94961 IOT instruction (core dumped)  nix run
```

resultin in an immediate crash. 

### Expected Behavior

all imports being found and not having an qt related crash

### Steps to Reproduce

1. boot a nixos system
2. run `nix run github:TagStudioDev/TagStudio#tagstudio`

### Logs

perovskite ~/coding/TagStudio  main 
❯ nix run                                        
2025-07-13 16:51:27 [info     ] [FFmpeg] Using FFprobe location: ffprobe (Found)
2025-07-13 16:51:27 [info     ] [FFmpeg] Using FFmpeg location: ffmpeg (Found)
2025-07-13 16:51:27 [error    ] [ThumbRenderer] Could not import the "pillow_jxl" module
Traceback (most recent call last):
  File "/nix/store/r3hf69imbw20w923x4ij5xdl9saz16ss-TagStudio-9.5.2/lib/python3.12/site-packages/tagstudio/qt/widgets/thumb_renderer.py", line 82, in <module>
    import pillow_jxl  # noqa: F401 # pyright: ignore[reportUnusedImport]
    ^^^^^^^^^^^^^^^^^
[...]

Issue #915, opened 2025-04-19
[Bug]: TagStudio breaks in nixos-unstable


zierf:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

97ee43c

### Operating System & Version

NixOS 25.05 (Warbler), 25.05.20250417.b024ced (NixOS/nixpkgs/b024ced1aac25639f8ca8fdfc2f8c4fbd66c48ef)

### Description

It looks like there is a regression for the GlobalMenu in the current packages from `nixos-unstable`.
(See [PR #870](https://github.com/TagStudioDev/TagStudio/pull/870) and it's linked issues.)

It doesn't work for [Nixpkgs Commit b024ced](https://github.com/NixOS/nixpkgs/commit/b024ced1aac25639f8ca8fdfc2f8c4fbd66c48ef).
If I'm not mistaken, it worked at least until [Nixpkgs Commit 2631b0b](https://github.com/NixOS/nixpkgs/commit/2631b0b7abcea6e640ce31cd78ea58910d31e650).

![GlobalMenu](https://github.com/user-attachments/assets/2827cabc-80b0-4385-81b5-5e442f9febb2)

---

I discovered this because I was trying to update my system and there is a build error.
There were also some other questionable messages about _fontconfig_ and _ffmpeg_.

Apart from the JXL plugin, I'm using the [installation suggested in the documentation](https://docs.tagstud.io/install/#nixos).

<details>
<summary>TagStudio Installation in Flake</summary>

`flake.nix`:
```Nix
tagstudio = {
  url = "github:TagStudioDev/TagStudio";
  inputs.nixpkgs.follows = "nixpkgs";
};
```

`home.nix`:
```Nix
inputs.tagstudio.packages.${pkgs.stdenv.hostPlatform.system}.tagstudio-jxl
```
</details>

In order to update my system at all, I had to remove `inputs.nixpkgs.follows = "nixpkgs";`.

However, as mentioned above, the global menu is now missing too.
If I saw it correctly, new QT packages were deployed. That one might have something to do with that.

### Expected Behavior

- A working global menu.
- Working build with `inputs.nixpkgs.follows = "nixpkgs";`,
[...]

PR #1048, opened 2025-08-23
fix(nix): fixup and rework, always use nixpkgs PySide/Qt

xarvex:
### Summary

~Depends: #1047~
Fixes: #915
Fixes: #988

Took me some time to completely work this out (and to get sorted from life stuff). The good part is that whatever issues were halting simply using the newer versions of PySide are seemingly no longer there. The wrapping done for Qt was fixed, it was not quite done correctly, which broke using hardware acceleration. Outside of that, this does some general touch ups to get to the state of the current repo.

To hopefully prevent future mismatches between the version of PySide, and therefore Qt, used in the project versus present in nixpkgs, I override where the PySide module from nixpkgs is always used, even in the dev shell. This is a trade-off from sticking to an environment closely resembling ideals, but managing the drift in versions clearly hasn't worked out. It might be worth checking out patching the dynamic library for Qt that PySide loads, but that might still have the issue of PySide's own version.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality
    
    -   [x] Nix package

### DIFF
--- a/nix/package/default.nix
+++ b/nix/package/default.nix
@@ -86,7 +86,6 @@ python3Packages.buildPythonApplication {
   dependencies =
     with python3Packages;
     [
-      audioop-lts
       chardet
       ffmpeg-python
       humanfriendly
--- a/flake.nix
+++ b/flake.nix
@@ -28,7 +28,7 @@
       perSystem =
         { pkgs, ... }:
         let
-          python3 = pkgs.python313;
+          python3 = pkgs.python312;
         in
         {
           packages =


nix/package/default.nix
```diff
@@ -89,15 +101,16 @@ python3Packages.buildPythonApplication {
       rawpy
       send2trash
       sqlalchemy
+      srctools
       structlog
       toml
       ujson
-      vtf2img
+      wcmatch
```
Computerdores, line 107: nit: iirc only ripgrep *or* wcmatch is needed with ripgrep being used if both are present

xarvex: Yeah, I was aware of that, but wasn't entirely sure how to go about it in terms of packaging. Perhaps what I might do is a toggle named `withRipgrep` and depending on the value of that, will include ripgrep or wcmatch.

Computerdores: From what I remember ripgrep support was added as a performance optimisation, so it might be fine to leave it as is and remove the wcmatch dependency. Otherwise for having a toggle: Do you mean as a CLI parameter or for the nix package? Because doing that for the nix package would require a nixos/home manager module no?

CyanVoxel: More recently I've had to utilize `wcmatch` in a couple other areas of the codebase (thumbnail renderer and file attributes in the preview panel) though *ideally* in the future I could get the same functionality out of `ripgrep` for those pattern matching tasks, but until then `wcmatch` is at least always required

xarvex: The only reason I would like to have both in the package is to allow us to test both situations of ripgrep being used and not use. And by toggle, I do mean in the package itself. If you look at `withJXLSupport` that is something that can be set in an override (though that is outputted in a `tagstudio-jxl` package). This would work the same way, although default to true.

Example usage:
```nix
{ inputs, pkgs, ... }:

{
  environment.systemPackages = [
    (inputs.tagstudio.packages.${pkgs.stdenv.hostPlatform.system}.tagstudio.override { withRipgrep = false; })
  ];
}
```

It doesn't need outputting as a package though as it will be mainly intended for testing purposes. Anyways, I will be pushing this in just a moment.

xarvex: Whoops, wrote that before seeing @CyanVoxel's response. I suppose that means then to keep the package as-is. Still, hope that helps explain what I meant @Computerdores.

xarvex:
Could have sworn Python 3.12 did not function while doing this, seems to be all good using it for the moment. Will simply adjust back to 3.13 depending on what happens with #1047.

That makes this mergeable.

PR #1058, opened 2025-08-25
refactor: store DB version inside `versions` table

CyanVoxel:
### Summary
This PR starts storing the DB version inside of a new `versions` table instead of the ill-named `preferences` table.
-   Deprecates the `preferences` table, set to be removed in a future TagStudio version
-   Introduces the `versions` table
    -   Has a string `key` column and an int `value` column
    -   The `key` column stores one of two values: `'INITIAL'` and `'CURRENT'`
    -   `'INITIAL'` stores the database version number in which in was created
        -   Pre-existing databases set this number to `100`
    -   `'CURRENT'` stores the current database version number
- Bump version to 101 (minor)
- Docstring tweaks to existing methods
- Prepended all DB migration methods with dunders to signify that they're private
image

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality


Computerdores:
What do you think about also storing when a migration happens? This would give us some debug info if we ever encounter a library with weird artefacts caused by a faulty migration

CyanVoxel:
> What do you think about also storing when a migration happens? This would give us some debug info if we ever encounter a library with weird artefacts caused by a faulty migration

That would be valuable information, but I don't think would work in this `versions` table specifically due to the data shape. There could be a new table called something like `migration_history` with an integer "version" column and a datetime "date" column that just keeps an ongoing log of when migrations happen. Outside of the scope of this PR, though

src/tagstudio/core/library/alchemy/models.py
```diff
@@ -311,8 +311,17 @@ def slugify_field_key(mapper, connection, target):
         target.key = slugify(target.tag)
 
 
+# NOTE: The "Preferences" table has been depreciated as of TagStudio 9.5.4
+# and is set to be removed in a future release.
 class Preferences(Base):
```
Computerdores, line 318: Nit: maybe mark this `@deprecated`?

CyanVoxel: I don't know why I've been under the impression that the `@deprecated` decorator itself was depreciated, glad to use it here!

Issue #1053, opened 2025-08-24
[Bug]: Empty frame as video preview if auto-play is paused


Qawes:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Alpha 9.5.3 Release

### Operating System & Version

Windows 10

### Description

Selecting a video file while auto-play is disabled, shows an empty frame in the sidebar. I'm guessing this is because the preview tries to grab the first frame, but it is not read in correctly (or at all)?

### Expected Behavior

Selecting a video file should show the first actual frame in the sidebar as preview.

### Steps to Reproduce

1. If Autoplay is enabled, select a video file, and stop the playback.
2. Select a video file
3. Empty frame shown in sidebar

CyanVoxel:
I've noticed that it's not just an issue when autoplay is toggled off, but also the general pause state. If the player is paused, regardless of the autoplay state, the next video clicked also does not have the first frame loaded.

PR #1059, opened 2025-08-25
fix : Empty frame as video preview if auto-play is paused #1053

SumithSudheer:
### Summary
fix :  Empty frame as video preview if auto-play is paused #1053

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [ ] Basic functionality

### DIFF
--- a/src/tagstudio/qt/widgets/media_player.py
+++ b/src/tagstudio/qt/widgets/media_player.py
@@ -455,7 +455,7 @@ class MediaPlayer(QGraphicsView):
             current = self.format_time(self.player.position())
             duration = self.format_time(self.player.duration())
             self.position_label.setText(f"{current} / {duration}")
-            
+
             # The key change: if autoplay is off, pause the player after the media is loaded
             if not self.autoplay.isChecked():
                 self.player.pause()
@@ -493,4 +493,3 @@ class VideoPreview(QGraphicsVideoItem):
     @override
     def boundingRect(self):
         return QRectF(0, 0, self.size().width(), self.size().height())
-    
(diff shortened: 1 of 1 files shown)


CyanVoxel:
Thank you for this PR! When testing I realized that the original report for #1053 actually under-reported the bug, and the issue is also present when the player is in the paused state and a new video is clicked (see [this comment](https://github.com/TagStudioDev/TagStudio/issues/1053#issuecomment-3229997220)):

> I've noticed that it's not just an issue when autoplay is toggled off, but also the general pause state. If the player is paused, regardless of the autoplay state, the next video clicked also does not have the first frame loaded.

Would you be willing to look into this as well for this PR?

SumithSudheer:
> Thank you for this PR! When testing I realized that the original report for #1053 actually under-reported the bug, and the issue is also present when the player is in the paused state and a new video is clicked (see [this comment](https://github.com/TagStudioDev/TagStudio/issues/1053#issuecomment-3229997220)):
> 
> > I've noticed that it's not just an issue when autoplay is toggled off, but also the general pause state. If the player is paused, regardless of the autoplay state, the next video clicked also does not have the first frame loaded.
> 
> Would you be willing to look into this as well for this PR?
Thanks for catching that! 🙏  
Yes, I’m willing to look into the paused state scenario as well. I’ll update this PR to handle both autoplay being disabled and the general paused state so that the first frame loads correctly when a new video is clicked.  

I’ll push the changes once I’ve tested them locally. 👍

CyanVoxel:
There now seems to be a regression where having autoplay checked and the player paused causes the original behavior:
image

Meanwhile having autoplay unchecked loads the first frame as desired.

CyanVoxel:
Closed in favor of #1104

PR #1062, opened 2025-08-26
ci: expand pyright ignore rules to vendored and tests

CyanVoxel:
### Summary

This PR sets the Pyright `reportMissingTypeStubs` rule to `false`, ignores the vendored pydub file(s), and ignores specific rules for tests that otherwise can't be reasonably accommodated.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,13 @@
+# pyright: reportArgumentType=false
+# pyright: reportAttributeAccessIssue=false
+# pyright: reportMissingParameterType=false
+# pyright: reportOptionalMemberAccess=false
+# pyright: reportPrivateUsage=false
+# pyright: reportUnknownParameterType=false
+# pyright: reportUnknownVariableType=false
+# pyright: reportUnusedParameter=false
+
+
 import sys
 from pathlib import Path
 from tempfile import TemporaryDirectory
@@ -155,7 +165,7 @@ def qt_driver(qtbot, library, library_dir: Path):
 
         driver.app = Mock()
         driver.main_window = Mock()
-        driver.main_window.preview_panel = Mock()
+        # driver.main_window.preview_panel = Mock()
         driver.main_window.thumb_grid = Mock()
         driver.main_window.thumb_size = 128
         driver.item_thumbs = []
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -166,14 +166,14 @@ def qt_driver(qtbot, library, library_dir: Path):
         driver.app = Mock()
         driver.main_window = Mock()
         # driver.main_window.preview_panel = Mock()
-        driver.main_window.thumb_grid = Mock()
+        # driver.main_window.thumb_grid = Mock()
         driver.main_window.thumb_size = 128
         driver.item_thumbs = []
-        driver.main_window.menu_bar.autofill_action = Mock()
+        # driver.main_window.menu_bar.autofill_action = Mock()
 
         driver.copy_buffer = {"fields": [], "tags": []}
-        driver.main_window.menu_bar.copy_fields_action = Mock()
-        driver.main_window.menu_bar.paste_fields_action = Mock()
+        # driver.main_window.menu_bar.copy_fields_action = Mock()
+        # driver.main_window.menu_bar.paste_fields_action = Mock()
 
         driver.lib = library
         # TODO - downsize this method and use it
--- a/tests/qt/test_build_tag_panel.py
+++ b/tests/qt/test_build_tag_panel.py
@@ -1,3 +1,12 @@
+# pyright: reportArgumentType=false
+# pyright: reportAttributeAccessIssue=false
+# pyright: reportMissingParameterType=false
+# pyright: reportOptionalMemberAccess=false
+# pyright: reportPrivateUsage=false
+# pyright: reportUnknownParameterType=false
+# pyright: reportUnknownVariableType=false
+# pyright: reportUnusedParameter=false
+
 from tagstudio.core.library.alchemy.models import Tag
 from tagstudio.qt.modals.build_tag import BuildTagPanel
 from tagstudio.qt.translations import Translations
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,12 +84,13 @@ ignore_errors = true
 qt_api = "pyside6"
 
 [tool.pyright]
-ignore = [".venv/**"]
+ignore = ["src/tagstudio/qt/helpers/vendored/pydub/", ".venv/**"]
 include = ["src/tagstudio", "tests"]
 reportAny = false
 reportIgnoreCommentWithoutRule = false
 reportImplicitStringConcatenation = false
 reportMissingTypeArgument = false
+reportMissingTypeStubs = false
 # reportOptionalMemberAccess = false
 reportUnannotatedClassAttribute = false
 reportUnknownArgumentType = false
--- a/tests/macros/test_dupe_entries.py
+++ b/tests/macros/test_dupe_entries.py
@@ -1,3 +1,7 @@
+# pyright: reportMissingParameterType=false
+# pyright: reportUnknownParameterType=false
+
+
 from pathlib import Path
 
 from tagstudio.core.library.alchemy.models import Entry
--- a/tests/macros/test_folders_tags.py
+++ b/tests/macros/test_folders_tags.py
@@ -1,3 +1,8 @@
+# pyright: reportMissingParameterType=false
+# pyright: reportUnknownParameterType=false
+# pyright: reportUnknownVariableType=false
+
+
 from tagstudio.qt.modals.folders_to_tags import folders_to_tags
 
 
--- a/tests/macros/test_missing_files.py
+++ b/tests/macros/test_missing_files.py
@@ -16,6 +16,7 @@ def test_refresh_missing_files(library: Library):
     registry = MissingRegistry(library=library)
 
     # touch the file `one/two/bar.md` but in wrong location to simulate a moved file
+    assert library.library_dir
     (library.library_dir / "bar.md").touch()
 
     # no files actually exist, so it should return all entries
--- a/tests/macros/test_refresh_dir.py
+++ b/tests/macros/test_refresh_dir.py
@@ -1,3 +1,7 @@
+# pyright: reportMissingParameterType=false
+# pyright: reportUnknownParameterType=false
+
+
 from pathlib import Path
 from tempfile import TemporaryDirectory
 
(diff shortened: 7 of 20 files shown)


CyanVoxel:
> Why did you choose to set so many warnings to be ignored in the tests? For example: In several of the tests you set `reportMissingParameterType=false`, `reportUnknownVariableType=false`, and the like, imo, it would be better to annotate the types in that situation since that makes it clear what type is being handled

Really what I would want in the end is the ability for Pyright to still perform checks while editing, but not have any errors and warnings from the test files factor in to whether or not a workflow test fails. There are many cases in the test files where Pyright is asking for type clarity in situations where it's just not possible to add. Even if more type annotations were added, it would still complain about "Unknown" types being returned from functions which we can't specify the return type from. Those "Unknowns" also propagate to return types, creating even more complicated messes of rule warnings. Some type annotations just seemingly refuse to work, for example whatever "QtBot" is. There are similar issues beyond just type annotations, for example when something like an "Args()" object is passed as a parameter to one of our functions that does not expect a strange mock test object in place of production arguments.

Overall I just don't want the impossible fixes to weigh into workflow checks, while also not pulling our (or at the very least my) hair out over Pyright nitpicking this that *may* be possible to fix or individually silence but not without spending a (subjectively) disproportionate amount of time and verbosity to whack-a-mole fix the warnings. Maybe this is all coming from a place of the tests wearing me down ever since they were introduced for *numerous* reasons, but I feel the tests shouldn't require *more* effort to satisfy Pyright than the actual production code.

> (also, autocompletion is otherwise broken for me or was in the past at least)

This miiiiiiiiiiight be due to eb2887e871b8002f55d00db7f39aa36eeb79674c - Ever since the extraPaths were added, I've had issues on my end in VSCode with it asking for type stubs constantly from `tagstudio` imports (also fixable by turning off that check, which I did but because of other imports) and VSCode auto-completion of `tagstudio` imports breaking for me ever since this was introduced. I remember I had asked when you added the "include" and "extraPaths" to the pyprojct.toml and you had mentioned that this *fixed* autocomplete for you. I had hoped that maybe it was just the "include" that fixed that for you for the tests and that the "extraPaths" may have been unnessisary in your fix, since it's what's been causing issues on my end. I gave @xarvex the go ahead to commit the changes to main and told myself I'd deal with any complaints of it breaking other users' completions if it ever popped up. Aaaaand it seems like once again it's popped up... I don't know why this stuff has *always* been a whack-a-mole issue of working correctly for some people and then not for others, and then whenever something changes to fix it for some people it breaks for others. If you or anyone one else can find an alternative fix then I can verify if it breaks on my end, or find a fix that I can apply on my end that lets you use the extraPaths. I'm really just so worn out by this stuff and just want it to *work...*

Computerdores:
> > Why did you choose to set so many warnings to be ignored in the tests? For example: In several of the tests you set `reportMissingParameterType=false`, `reportUnknownVariableType=false`, and the like, imo, it would be better to annotate the types in that situation since that makes it clear what type is being handled
> 
> Really what I would want in the end is the ability for Pyright to still perform checks while editing, but not have any errors and warnings from the test files factor in to whether or not a workflow test fails. There are many cases in the test files where Pyright is asking for type clarity in situations where it's just not possible to add. Even if more type annotations were added, it would still complain about "Unknown" types being returned from functions which we can't specify the return type from. Those "Unknowns" also propagate to return types, creating even more complicated messes of rule warnings. Some type annotations just seemingly refuse to work, for example whatever "QtBot" is. There are similar issues beyond just type annotations, for example when something like an "Args()" object is passed as a parameter to one of our functions that does not expect a strange mock test object in place of production arguments.
>
> Overall I just don't want the impossible fixes to weigh into workflow checks, while also not pulling our (or at the very least my) hair out over Pyright nitpicking this that _may_ be possible to fix or individually silence but not without spending a (subjectively) disproportionate amount of time and verbosity to whack-a-mole fix the warnings. Maybe this is all coming from a place of the tests wearing me down ever since they were introduced for _numerous_ reasons, but I feel the tests shouldn't require _more_ effort to satisfy Pyright than the actual production code. On the other hand it's also not that important so just throwing a TODO on it to remove the file wide ignores and address the issues instead would probably be fine too.

Yeah not failing the workflow because of pyright stuff in the tests would be good imo. But while I can't speak to the difficulties with type annotation, I feel like at least the low hanging fruit should be annotated (like `library: Library` in `test_folders_tags.py` for example).

> > (also, autocompletion is otherwise broken for me or was in the past at least)
> 
> This miiiiiiiiiiight be due to [eb2887e](https://github.com/TagStudioDev/TagStudio/commit/eb2887e871b8002f55d00db7f39aa36eeb79674c) - Ever since the extraPaths were added, I've had issues on my end in VSCode with it asking for type stubs constantly from `tagstudio` imports (also fixable by turning off that check, which I did but because of other imports) and VSCode auto-completion of `tagstudio` imports breaking for me ever since this was introduced. I remember I had asked when you added the "include" and "extraPaths" to the pyprojct.toml and you had mentioned that this _fixed_ autocomplete for you. I had hoped that maybe it was just the "include" that fixed that for you for the tests and that the "extraPaths" may have been unnessisary in your fix, since it's what's been causing issues on my end. I gave @xarvex the go ahead to commit the changes to main and told myself I'd deal with any complaints of it breaking other users' completions if it ever popped up. Aaaaand it seems like once again it's popped up... I don't know why this stuff has _always_ been a whack-a-mole issue of working correctly for some people and then not for others, and then whenever something changes to fix it for some people it breaks for others. If you or anyone one else can find an alternative fix then I can verify if it breaks on my end, or find a fix that I can apply on my end that lets you use the extraPaths. I'm really just so worn out by this stuff and just want it to _work..._

I added the `extraPaths` not because of auto complete, but because not having it breaks refactorings like renaming in VS Code.
However, for me, it seems like for some reason `extraPaths` is not needed anymore.

Also, additionally, the annotations seems to not be needed for autocomplete anymore (not sure if that is the AI stuff at work), however refactorings don't work without the annotations (for example renaming `Library.all_entries` does not apply correctly to `test_folders_tags.py` unless the parameter is annotated with `Library`) which can make refactoring a little annoying in my experience.

CyanVoxel:
I'll work on this some more and try to include type hints where I can. I was a bit hasty when doing this due to getting fed up with both Pyright and pytest shenanigans *(if you want a taste, try removing `test_sidecar.py`).* Not an excuse for laziness with some of these low hanging fruit ones, though.

CyanVoxel:
@Computerdores I was able to remove all of the ignores from the top of the test files, and added type hints in almost all the missing places. I even discovered a few tests that were actually completely broken and discovered how to delete the cursed `test_sidecar.py` file. All in all, definitely worth it to go in and do things in detail (even if I lost a bit of my sanity in the process).

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -896,6 +896,7 @@ def remove_entries(self, entry_ids: list[int]) -> None:
                 for i in range(0, len(entry_ids), MAX_SQL_VARIABLES)
             ]:
                 session.query(Entry).where(Entry.id.in_(sub_list)).delete()
+                session.flush()
```
Computerdores: Could you explain why you added this?
I ask because `flush` is a part of `commit`, so for pure functionality, it shouldn't be necessary. Unless it increases performance here?

src/tagstudio/qt/view/widgets/preview_panel_view.py
```diff
@@ -198,7 +198,7 @@ def add_buttons_enabled(self, enabled: bool):
         self.__add_tag_button.setEnabled(enabled)
 
     @property
-    def _file_attributes_widget(self) -> FileAttributes:  # needed for the tests
+    def file_attributes_widget(self) -> FileAttributes:  # needed for the tests
```
Computerdores: The idea behind marking this private was that other code shouldn't be messing with internal parts of the widget in order to decouple independent parts of the code base. Iirc, the only reason I didn't go for name mangling with `__` is because the tests need to access it.

tests/conftest.py
```diff
@@ -24,8 +31,9 @@ def cwd():
 def file_mediatypes_library():
     lib = Library()
 
-    status = lib.open_library(Path(""), ":memory:")
+    status = lib.open_library(Path(""), ":memory:")  # pyright: ignore[reportArgumentType]
```
Computerdores: Perhaps open_library should accept strings as the storage argument? Because while we currently don't make use of in-memory DBs outside of the tests, we do implement it in `open_library` so we might as well change the type annotations accordingly. 

(Also applies to the other instances of this ignore comment on `open_library`)

Computerdores (commented):
What are the instances of `qtbot.addWidget` without further usage of `qtbot` for?

src/tagstudio/core/library/alchemy/library.py
CyanVoxel: Meant to remove that after I finished testing things around to trace back a broken test, but initially added it because I was investigating the individual `delete` calls not being flushed before the final `commit`outside the loop. Should be fine to remove now though

tests/conftest.py
CyanVoxel: I completely looked past the fact that this is *our* method, I don't know why I shrugged it off as a SQLAlchemy one. Absolutely will do this 👍

CyanVoxel:
> What are the instances of `qtbot.addWidget` without further usage of `qtbot` for?

I could only find the pre-existing one in `qt_driver()` of `conftest.py`, and it looks like it wasn't actually required by the test. I'm really not too familiar with the QtBot system so just assumed it was there because it was needed, but seemingly not.

That and all other feedback should be addressed!

tests/qt/test_build_tag_panel.py
```diff
@@ -29,6 +44,7 @@ def test_build_tag_panel_remove_subtag_callback(library, generate_tag):
     assert child
 
     panel: BuildTagPanel = BuildTagPanel(library, child)
+    qtbot.addWidget(panel)
```
Computerdores, line 47: > What are the instances of qtbot.addWidget without further usage of qtbot for?

sorry, this is one of the places I meant by that; does this do anything? (genuine question, I don't know my way around pytestqt)

CyanVoxel: Oh yeah, so while I'm not sure what goes on behind the scenes with this, I know it's generally used in place of user GUI interaction. In this case, it (along with preventing the tests from hard crashing Python, which was the issue I ran into when removing the **EMPTY** `test_sidecar.py` file) and also prevents each widget being tested from spawning actual windows of them one at a time. This was behavior I observed after the tests were added in I believe 332 but didn't have the frame of reference to know if that was normal or not. My guess for a while now has been "no, that was not normal". But yeah, basically everything falls apart without this...

Computerdores (approved):

PR #1065, opened 2025-08-27
chore(thumb_renderer): prepare for pillow_heif removing AVIF support

xarvex:
### Summary
Was originally intending to bump `Pillow` to 11.3+, which code-wise we can do, but it appears that `mkdocs-material` is locked to version 10.2. This would have been important because `pillow_heif` has removed support for AVIF files because `Pillow` itself added support for the format (partial in 11.2.1, full in 11.3).

See: https://github.com/bigcat88/pillow_heif/pull/345
See: https://github.com/squidfunk/mkdocs-material/blob/6a9f156c2862437f9f143f01f668152968978839/pyproject.toml#L66C15-L66C16

Still, in preparation, I've added in `pillow_avif_plugin` which is compatible with both versions and will provide support in the meantime.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality
    
    -   [x] Nix package


Computerdores (approved):
changes look good and it works on win10

PR #1067, opened 2025-08-28
refactor: merge cyclicly imported files into library.py

CyanVoxel:
> [!note]
> ~~Blocked by #942, #948, and #1058 due to refactoring in overlapping areas.~~

### Summary
While I started this with the intention to merely clean up the Pyright errors inside the core library files, I ended up falling down the rabbit hole of discovering that the entire alchemy library directory has been riddled with [cyclical imports](https://docs.basedpyright.com/dev/configuration/config-files/#type-check-rule-overrides) that while not causing runtime issues, apparently do cause behind the scenes issues with Pyright and other type checkers while also indicating deeper issues with the architectural layering (see Eric Traut's [comment](https://github.com/microsoft/pyright/issues/746#issuecomment-647638663) on this as well as DetachHead's [comment](https://github.com/DetachHead/basedpyright/issues/960#issuecomment-2552777922)).

image

To address these issues I've chosen to merge nearly all of the code from `models.py`, `fields.py`, `joins.py`, `visitors.py`, and `default_color_groups.py` into `library.py`. In the process I've also addressed any Pyright issues along the way but have otherwise left the code unchanged. While this increases the size of `library.py` by about 33%, it *does* end up grouping together a lot of highly related functionality that maybe shouldn't have been siloed off into their own co-dependent files to begin with. I'd be happy to break out any code that doesn't need to be in here at this point, but this file may just need to be a big boy to avoid these types of cycles...

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality


CyanVoxel:
Closed in favor of #1103.

PR #1068, opened 2025-08-28
refactor: unwrap instead of assert not None

Computerdores:
### Summary
As discussed on discord, introduces an `unwrap` method which accepts an argument of type `T | None` and returns `T` or throws an exceptions if the argument is `None`.

### Tasks Completed
-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/core/library/alchemy/models.py
+++ b/src/tagstudio/core/library/alchemy/models.py
@@ -126,8 +126,8 @@ class Tag(Base):
 
     def __init__(
         self,
+        name: str,
         id: int | None = None,
-        name: str | None = None,
         shorthand: str | None = None,
         aliases: set[TagAlias] | None = None,
         parent_tags: set["Tag"] | None = None,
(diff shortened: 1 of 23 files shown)


src/tagstudio/core/library/alchemy/models.py
```diff
@@ -146,6 +146,7 @@ def __init__(
         self.shorthand = shorthand
         self.disambiguation_id = disambiguation_id
         self.is_category = is_category
+        # TODO: @reviewer any idea what this assert is for? This will always be true, no?
         assert not self.id
```
CyanVoxel: As far as I can tell this doesn't serve a purpose and should be safe to remove

src/tagstudio/core/utils/missing_files.py
```diff
@@ -41,7 +41,7 @@ def match_missing_file_entry(self, match_entry: Entry) -> list[Path]:
 
         Works if files were just moved to different subfolders and don't have duplicate names.
         """
-        assert self.library.library_dir
+        assert self.library.library_dir is not None
```
CyanVoxel: Why not use the `unwrap()` method here?

CyanVoxel (commented):
Great addition! Approved when comments are addressed 👍

Computerdores:
should all be addressed

PR #1082, opened 2025-09-01
feat: auto mnemonics

Computerdores:
### Summary
Reimplement KDE's system for autogenerating mnemonics[^1] which also incorporates `&` markings when present, but doesn't require them to be present.

Currently only works for the drop downs but could probably be adopted for the menu bar as well.

Example - File menu with all `&` removed from translations:
grafik

Possible Future PRs:
- removing `&` from translations where now unnecessary
- adopting it for the QMenuBar at the top
- adopting it for buttons in windows and the like

### Tasks Completed
-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality

[^1]: See [Wiki](https://api.kde.org/frameworks/kirigami/html/classMnemonicAttached.html), [Headerfile](https://api.kde.org/frameworks/kirigami/html/mnemonicattached_8h_source.html) and [Source Code](https://invent.kde.org/frameworks/kirigami/-/blob/master/src/mnemonicattached.cpp)

### DIFF
--- a/src/tagstudio/qt/main_window.py
+++ b/src/tagstudio/qt/main_window.py
@@ -52,6 +52,142 @@ if typing.TYPE_CHECKING:
 logger = structlog.get_logger(__name__)
 
 
+def remove_accelerator_marker(label: str) -> str:
+    """Remove existing accelerator markers (&) from a label."""
+    result = ""
+    skip = False
+    for i, ch in enumerate(label):
+        if skip:
+            skip = False
+            continue
+        if ch == "&":
+            # escaped ampersand "&&"
+            if i + 1 < len(label) and label[i + 1] == "&":
+                result += "&"
+                skip = True
+            # otherwise skip this '&'
+            continue
+        result += ch
+    return result
+
+
+# Additional weight for first character in string
+FIRST_CHARACTER_EXTRA_WEIGHT = 50
+# Additional weight for the beginning of a word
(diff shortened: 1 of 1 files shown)


src/tagstudio/qt/main_window.py
```diff
@@ -52,6 +52,142 @@
 logger = structlog.get_logger(__name__)
 
 
+def remove_accelerator_marker(label: str) -> str:
```
CyanVoxel, line 55: Just a note, not requiring a change here: There's some similar functionality in `change_language()` of `translations.py` that's currently only targeting macOS

CyanVoxel (approved):
Looks good on Fedora 42 KDE and doesn't intrude on macOS 15 👍
Might be good to move this to a separate file rather than keep it in the `main_window.py` one, but like you said that can be for a future PR.

src/tagstudio/qt/main_window.py
Computerdores: give me a sec to fix that

Computerdores: nvm, separate pr in a sec ^^

CyanVoxel: Whoops, just now saw this

PR #1097, opened 2025-09-05
feat: added trigger to prevent  a tag to be its own parent

### LINKED ISSUE #1074: [Bug]: Tags Can Be Parented to Each Other and Freeze the Program [#875 Regression]
### Checklist

### TagStudio Version

Alpha 9.5.3+ (main branch)

macoS 15.5

### Description

Tags can erroneously be parented to each other, and that causes a regression of #875 where the program freezes when trying to recursively trace the ancestor hierarchy (for notably when attempting to view tags in the preview panel.

Image

### Expected Behavior

- `Tag A` should not be shown in the "Add Parent Tags" list of `Tag B` if `Tag A` has `Tag B` as a parent
- Cyclical parent tag relationships should not be allowed in the database
- The program should not allow a scenario in which it locks up like this

### Steps to Reproduce

1. Have two tags, `Tag A` and `Tag B`
2. Edit `Tag A` to have `Tag B` as a parent, and then edit `Tag B` to have `Tag A` as a parent
3. Add one of the tags to a file to preview, or click on a file that already has one of the tags present to preview and observe the freeze

To undo these changes, you can relaunch TagStudio and edit the tags from the "Manage Tags" menu in order to undo the relationship.

### Logs

N/A

RustyNova016:
### Summary

Implemented a trigger to prevent inserting a tag parent that will result in a tag being its own parent. 

Right now, the trigger is tested, but not the actual migration (VScode tunnels can't tunnel QT apps sadly...), so if someone want to test before I get to it, feel free to do so

close: #1074

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [ ] Basic functionality


src/tagstudio/core/library/alchemy/library.py
```diff
@@ -681,6 +685,51 @@ def __apply_db100_parent_repairs(self, session: Session):
             session.commit()
             logger.info("[Library][Migration] Refactored TagParent column")
 
+    def __apply_db102_schema_changes(self, session: Session):
+        """Add a trigger to prevent having a tag being its own parent"""
+        add_trigger = text(
+            """
+            CREATE TRIGGER `trigger_before_insert_tag_parents` BEFORE INSERT ON `tag_parents` BEGIN
```
RustyNova016, line 692: Forgot `FOR EACH ROW`...

And updates

CyanVoxel:
What's the status of this PR?

PR #1101, opened 2025-09-06
feat: add cached thumb quaity and resolution settings

### LINKED ISSUE #802: [Feature Request]: Thumbnail caching, option to control file compression.
### Checklist

### Description

Currently Thumbnail caching https://github.com/TagStudioDev/TagStudio/pull/694 cache thumbnails with the default compression settings of PIL for webp. While the thumbnail caching feature already implements some features to reduces the file size like resizing the picture, caching may still grow big on large library.

### Solution

Having an option to control compression levels or directly increase compression would make the cache size even smaller. As thumbnails are displayed in low resolution in the preview box, compression on images would be less notified anyway while reducing size, write and read operations (yes, probably a reduction in kb, but imagine a very large library with lots of cache, that's still an good improvement).

### Alternatives

_No response_

CyanVoxel:
### Summary

This adds two new settings to the config file: `cached_thumb_quality` which controls the amount of compression used in the cached .webp images, and `cached_thumb_resolution` which determines the maximum dimension of a cached thumbnail image. This also fixes an uncaught exception that occurred when deleting the inner thumbs folder(s) on disk while the program was running.

After #1084 I've been toying with the idea of making certain power-user focused settings such as these only accessible in the config file in an attempt to keep clutter in the GUI down. Organization of the settings UI as more settings are introduced is outside the scope of this PR though, the important note is that I'm just adding these as config only options at the moment and am open to reconsidering that in the future.

Closes #802

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/core/global_settings.py
+++ b/src/tagstudio/core/global_settings.py
@@ -24,6 +24,10 @@ DEFAULT_GLOBAL_SETTINGS_PATH = (
 DEFAULT_THUMB_CACHE_SIZE = 500  # Number in MiB
 MIN_THUMB_CACHE_SIZE = 10  # Number in MiB
 
+# See: https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#webp-saving
+DEFAULT_CACHED_IMAGE_QUALITY = 80
+DEFAULT_CACHED_IMAGE_RES = 256
+
 
 class Theme(IntEnum):
     DARK = 0
@@ -56,6 +60,8 @@ class GlobalSettings(BaseModel):
     open_last_loaded_on_startup: bool = Field(default=True)
     generate_thumbs: bool = Field(default=True)
     thumb_cache_size: float = Field(default=DEFAULT_THUMB_CACHE_SIZE)
+    cached_thumb_quality: int = Field(default=DEFAULT_CACHED_IMAGE_QUALITY)
+    cached_thumb_resolution: int = Field(default=DEFAULT_CACHED_IMAGE_RES)
     autoplay: bool = Field(default=True)
     loop: bool = Field(default=True)
     show_filenames_in_grid: bool = Field(default=True)
--- a/src/tagstudio/qt/cache_manager.py
+++ b/src/tagstudio/qt/cache_manager.py
@@ -12,7 +12,7 @@ import structlog
 from PIL import Image
 
 from tagstudio.core.constants import THUMB_CACHE_NAME, TS_FOLDER_NAME
-from tagstudio.core.global_settings import DEFAULT_THUMB_CACHE_SIZE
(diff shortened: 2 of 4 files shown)


kitsumed:
> Organization of the settings UI as more settings are introduced is outside the scope of this PR though, the important note is that I'm just adding these as config only options at the moment and **am open to reconsidering that in the future**.

My personal thought on the subject is that we should add a toggle like "Advanced Settings" or "Advanced Mode" and give user access to more internal settings like theses. Or else documents them on the [docs.tagstud.io/](https://docs.tagstud.io/) website.

CyanVoxel:
> > Organization of the settings UI as more settings are introduced is outside the scope of this PR though, the important note is that I'm just adding these as config only options at the moment and **am open to reconsidering that in the future**.
> 
> My personal thought on the subject is that we should add a toggle like "Advanced Settings" or "Advanced Mode" and give user access to more internal settings like theses. Or else documents them on the [docs.tagstud.io/](https://docs.tagstud.io/) website.

I agree - I plan on adding a page on the docs site going over the settings in-depth soon as there's now a good amount of them. And future UI rework of the settings panel could definitely include an "Advanced" toggle to show additional detailed settings in-app. For now though, this PR at least gets the features in the program and can be integrated with a UI PR at a later date and be included on the docs site in the meantime.

Computerdores:
> I agree - I plan on adding a page on the docs site going over the settings in-depth soon as there's now a good amount of them. And future UI rework of the settings panel could definitely include an "Advanced" toggle to show additional detailed settings in-app. For now though, this PR at least gets the features in the program and can be integrated with a UI PR at a later date and be included on the docs site in the meantime.

What do you think about just adding another tab that is literally called "Advanced"? For me that would be clear enough and also out of the way. (We could also always add a warning or something at the top if that is deemed necessary)

CyanVoxel:
> What do you think about just adding another tab that is literally called "Advanced"? For me that would be clear enough and also out of the way. (We could also always add a warning or something at the top if that is deemed necessary)

There's actually a couple other aspects I'd like to change/add first, one being the proper addition of "Library" settings that are separate from the "Application" settings and also have a default overrideable configuration, and the other being something like a horizontal tab breakup of the settings to put things into categories as the number of settings increases (i.e. "UI", "Search", "Behavior", etc. where applicable. While an "Advanced" tab would work with the current UI, it would be another thing that would need to be refactored before more changes like that, since having those settings in their own tab would go against the "Application/[Possibly Default Library]/Library" main tab setup I have in mind to replace the current "Global aka Application"-only tab with.

PR #1103, opened 2025-09-06
refactor: fix most pyright issues in `library/alchemy/`

CyanVoxel:
### Summary
A version of #1067 that addresses the same Pyright issues in `src/tagstudio/core/library/alchemy/`, minus the large refactoring due to the [`reportImportCycles`](https://docs.basedpyright.com/v1.31.4/configuration/config-files/#reportImportCycles) error.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -2,6 +2,10 @@
 # Licensed under the GPL-3.0 License.
 # Created for TagStudio: https://github.com/CyanVoxel/TagStudio
 
+# NOTE: This file contains nessisary use of deprecated first-party code until that
+# code is removed in a future version (prefs).
+# pyright: reportDeprecated=false
+
 
 import re
 import shutil
@@ -18,7 +22,7 @@ from warnings import catch_warnings
 
 import sqlalchemy
 import structlog
-from humanfriendly import format_timespan
+from humanfriendly import format_timespan  # pyright: ignore[reportUnknownVariableType]
 from sqlalchemy import (
     URL,
     ColumnExpressionArgument,
@@ -40,6 +44,7 @@ from sqlalchemy import (
 )
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import (
+    InstanceState,
     Session,
     contains_eager,
     joinedload,
@@ -47,6 +52,7 @@ from sqlalchemy.orm import (
     noload,
     selectinload,
 )
+from typing_extensions import deprecated
 
 from tagstudio.core.constants import (
     BACKUP_FOLDER_NAME,
@@ -81,8 +87,8 @@ from tagstudio.core.library.alchemy.enums import (
 from tagstudio.core.library.alchemy.fields import (
     BaseField,
     DatetimeField,
+    FieldID,
     TextField,
-    _FieldID,
 )
 from tagstudio.core.library.alchemy.joins import TagEntry, TagParent
 from tagstudio.core.library.alchemy.models import (
@@ -210,9 +216,9 @@ class Library:
     """Class for the Library object, and all CRUD operations made upon it."""
 
     library_dir: Path | None = None
-    storage_path: Path | str | None
+    storage_path: Path | str | None = None
     engine: Engine | None = None
-    folder: Folder | None
+    folder: Folder | None = None
     included_files: set[Path] = set()
 
     def __init__(self) -> None:
@@ -224,7 +230,7 @@ class Library:
     def close(self):
         if self.engine:
             self.engine.dispose()
-        self.library_dir: Path | None = None
+        self.library_dir = None
         self.storage_path = None
         self.folder = None
         self.included_files = set()
@@ -300,8 +306,8 @@ class Library:
             ]
         )
         for entry in json_lib.entries:
-            for field in entry.fields:
-                for k, v in field.items():
+            for field in entry.fields:  # pyright: ignore[reportUnknownVariableType]
+                for k, v in field.items():  # pyright: ignore[reportUnknownVariableType]
                     # Old tag fields get added as tags
                     if k in LEGACY_TAG_FIELD_IDS:
                         self.add_tags_to_entries(entry_ids=entry.id + 1, tag_ids=v)
@@ -319,8 +325,8 @@ class Library:
         end_time = time.time()
         logger.info(f"Library Converted! ({format_timespan(end_time - start_time)})")
 
-    def get_field_name_from_id(self, field_id: int) -> _FieldID:
-        for f in _FieldID:
+    def get_field_name_from_id(self, field_id: int) -> FieldID | None:
+        for f in FieldID:
             if field_id == f.value.id:
                 return f
         return None
@@ -482,7 +488,7 @@ class Library:
                     except IntegrityError:
                         session.rollback()
 
-            for field in _FieldID:
+            for field in FieldID:
                 try:
                     session.add(
                         ValueType(
@@ -562,7 +568,7 @@ class Library:
             # Repair "Description" fields with a TEXT_LINE key instead of a TEXT_BOX key.
             desc_stmd = (
                 update(ValueType)
-                .where(ValueType.key == _FieldID.DESCRIPTION.name)
+                .where(ValueType.key == FieldID.DESCRIPTION.name)
                 .values(type=FieldTypeEnum.TEXT_BOX.name)
             )
             session.execute(desc_stmd)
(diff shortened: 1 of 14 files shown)


src/tagstudio/core/library/alchemy/library.py
```diff
@@ -2,6 +2,10 @@
 # Licensed under the GPL-3.0 License.
 # Created for TagStudio: https://github.com/CyanVoxel/TagStudio
 
+# NOTE: This file contains nessisary use of deprecated first-party code until that
+# code is removed in a future version (prefs).
+# pyright: reportDeprecated=false
+
```
Computerdores:
```suggestion
# NOTE: This file contains necessary use of deprecated first-party code until that
# code is removed in a future version (prefs).
# pyright: reportDeprecated=false

```

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1118,7 +1118,7 @@ def remove_tag(self, tag: Tag):
                 tags_query = select(Tag).options(
                     selectinload(Tag.parent_tags), selectinload(Tag.aliases)
                 )
-                tag = session.scalar(tags_query.where(Tag.id == tag.id))
+                tag_: Tag = unwrap(session.scalar(tags_query.where(Tag.id == tag.id)))
```
Computerdores: Doesn't need to be a part of this PR, but a couple things I noticed about this method:

- This method doesn't need to be passed a Tag, only the ID would be enough. Because of that imo it would be better to only accept the ID
- the return value is never used anywhere (also don't really see the sense in it beyond doing it because it was already fetched anyways)
- there is seemingly no reason to fetch the tag from the DB, because the deletion could be done by ID instead

I took the liberty to apply those things (untested):
```python
    def remove_tag(self, tag_id: int) -> None:
        with Session(self.engine, expire_on_commit=False) as session:
            try:
                child_tags = session.scalars(
                    select(TagParent).where(TagParent.child_id == tag_id)
                ).all()
                aliases = session.scalars(select(TagAlias).where(TagAlias.tag_id == tag_id))

                for alias in aliases or []:
                    session.delete(alias)

                for child_tag in child_tags or []:
                    session.delete(child_tag)
                    session.expunge(child_tag)

                disam_stmt = (
                    update(Tag)
                    .where(Tag.disambiguation_id == tag_id)
                    .values(disambiguation_id=None)
                )
                session.execute(disam_stmt)
                session.flush()

                session.query(Tag).filter_by(id=tag_id).delete()
                session.commit()
            except IntegrityError as e:
                logger.error(e)
                session.rollback()
```

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1735,7 +1735,13 @@ def update_color(self, old_color_group: TagColorGroup, new_color_group: TagColor
             else:
                 self.add_color(new_color_group)
 
-    def update_aliases(self, tag, alias_ids, alias_names, session):
+    def update_aliases(
+        self,
+        tag: Tag,
+        alias_ids: list[int] | set[int],
+        alias_names: list[str] | set[str],
```
Computerdores: What about this?
```suggestion
        alias_ids: Iterable[int],
        alias_names: Iterable[str],
```

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1822,26 +1828,29 @@ def set_version(self, key: str, value: int) -> None:
...
         with Session(self.engine) as session:
             if isinstance(key, LibraryPrefs):
-                return session.scalar(select(Preferences).where(Preferences.key == key.name)).value
+                return session.scalar(select(Preferences).where(Preferences.key == key.name)).value  # pyright: ignore
             else:
-                return session.scalar(select(Preferences).where(Preferences.key == key)).value
+                return session.scalar(select(Preferences).where(Preferences.key == key)).value  # pyright: ignore
 
-    def set_prefs(self, key: str | LibraryPrefs, value: Any) -> None:
+    def set_prefs(self, key: str | LibraryPrefs, value: Any) -> None:  # pyright: ignore[reportExplicitAny]
```
Computerdores, line 1845: Should probably also get the `@deprecated`

src/tagstudio/core/library/alchemy/models.py
```diff
@@ -280,8 +285,8 @@ class ValueType(Base):
     key: Mapped[str] = mapped_column(primary_key=True)
     name: Mapped[str] = mapped_column(nullable=False)
     type: Mapped[FieldTypeEnum] = mapped_column(default=FieldTypeEnum.TEXT_LINE)
-    is_default: Mapped[bool]
-    position: Mapped[int]
+    is_default: Mapped[bool]  # pyright: ignore[reportUninitializedInstanceVariable]
+    position: Mapped[int]  # pyright: ignore[reportUninitializedInstanceVariable]
```
Computerdores, line 289: Why are these not initialised?

src/tagstudio/core/library/alchemy/visitors.py
```diff
@@ -174,6 +169,8 @@ def __separate_tags(
                         elif len(ids) == 1:
                             tag_ids.append(ids[0])
                             continue
+                    case _:
+                        pass
```
Computerdores: I would throw an exception here instead since an unhandled Constraint would lead to an incorrect query and no ConstraintType should ever be unhandled

src/tagstudio/core/library/alchemy/library.py
CyanVoxel: *Would you believe me if I said that this is my most frequently misspelled word and that I have corrected it so many times that at some point I accidentally added it to my local spell checker exception list out of a misclick?...*

src/tagstudio/core/library/alchemy/library.py
CyanVoxel: The `Iterable` type does not have a `remove()` method which is used later on:
image

src/tagstudio/core/library/alchemy/models.py
CyanVoxel: The "Field" code is a spiderweb I do not understand and would rather not poke at until the day I finally get to rip it out of the program.

- `is_default` does not seem to be used and is leftover from early database development assumptions
- `position` afaik this is not used either and is also the byproduct of early assumptions

src/tagstudio/core/library/alchemy/library.py
Computerdores: Don't worry, being able to correct native speakers gives me a huge ego boost from time to time, so all good😄

src/tagstudio/core/library/alchemy/library.py
Computerdores: `collections.abc.MutableSequence` should be another option then

CyanVoxel: That seems to work! I'll go ahead and change other instances of set | int in this file

src/tagstudio/core/library/alchemy/visitors.py
CyanVoxel: I agree, that makes sense - however... I'm now realizing that there are 9 of our test cases failing due to falling under this case:
```yaml
FAILED tests/test_search.py::test_and[orange and filetype:png-5] - NotImplementedError: Unhandled constraint
FAILED tests/test_search.py::test_and[square and filetype:jpg-2] - NotImplementedError: Unhandled constraint
FAILED tests/test_search.py::test_and[orange filetype:png-5] - NotImplementedError: Unhandled constraint
FAILED tests/test_search.py::test_and[green path:*inherit*-4] - NotImplementedError: Unhandled constraint
FAILED tests/test_search.py::test_or[filetype:jpg or tag:orange-11] - NotImplementedError: Unhandled constraint
FAILED tests/test_search.py::test_or[red or filetype:png-28] - NotImplementedError: Unhandled constraint
FAILED tests/test_search.py::test_or[filetype:jpg or path:*comp*-11] - NotImplementedError: Unhandled constraint
FAILED tests/test_search.py::test_not[special:untagged or not filetype:jpg-26] - NotImplementedError: Unhandled constraint
FAILED tests/test_search.py::test_parentheses[filetype:png and (tag:square or green)-12] - NotImplementedError: Unhandled constraint
```

CyanVoxel: Doing slightly some more digging and this appears to be for the cases of `ConstraintType.FileType`, `ConstraintType.Path`, and one `ConstraintType.Special`. I think I could just add those into the switch and give them passes? Since that was effectively the behavior before, and an exception will still be raised for anything new or unexpected coming through

PR #1095, opened 2025-09-04
refactor: untangle backend and frontend files

CyanVoxel:
### Summary
This is a **large** refactor of the Qt UI structure with the following intentions:
- Separate out erroneously mixed frontend and backend files
- Organize UI files that were previously in incorrect or misleading folders
- Remove unnecessarily specific or broad UI folders
- Delete unused file(s)
- Prepare un-refactored UI files for upcoming MVC refactors by moving them to a new temporary `qt/mixed` directory
- Rename files to be consistent with the project's naming schemes
- Prepare the layout for a `ThumbRenderer` refactor split
- Fix some minor Pyright errors

### Old Layout
```
├── src
│   └── tagstudio
│       ├── core
│       │   ├── library
│       │   │   ├── alchemy
│       │   │   └── json
│       │   ├── query_lang
│       │   └── utils
│       ├── qt
│       │   ├── controller
│       │   │   ├── components
│       │   │   └── widgets
│       │   │       └── preview
│       │   ├── helpers
│       │   │   └── vendored
│       │   │       └── pydub
│       │   ├── modals
│       │   ├── utils
│       │   ├── view
│       │   │   ├── components
│       │   │   └── widgets
│       │   │       └── preview
│       │   └── widgets
│       │       ├── paged_panel
│       │       └── preview
```

### New Layout
```py
├── src
│   └── tagstudio
│       ├── core
│       │   ├── library
│       │   │   ├── alchemy
│       │   │   │   └── registries
│       │   │   └── json
│       │   ├── query_lang
│       │   └── utils
│       ├── qt
│       │   ├── controllers
│       │   ├── helpers
│       │   ├── mixed  # Temporary until MVC refactor is complete
│       │   ├── models
│       │   ├── previews
│       │   │   └── vendored
│       │   │       └── pydub
│       │   ├── utils
│       │   └── views
│       │       ├── layouts
│       │       └── styles
```

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality
    -   [ ] PyInstaller executable

Issue #1061, opened 2025-08-26
[CI]: Use Pyright instead of MyPy for Type Checking


CyanVoxel:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

This has been discussed privately for a while or mentioned here and there, but I wanted to establish a coordinated intention to replace MyPy with Pyright as the project's static type checker. As the project has grown, MyPy has suffered from worsening performance that's impacted development and slowed down the CI workflows.

### Solution

Pyright is a more performant option, which also provides several improved checks that MyPy does not offer. The pyproject.toml has already included rules for Pyright that has been continuously tweaked over the past few months to suit the needs of the project.

The reason why there hasn't been a switch so far (besides dialing the Pyright rules) is because much of the existing codebase does not adhere to the new rules yet and would require substantial refactoring. Much of the codebase has already been undergoing these sorts of refactors, and new code has been loosely enforcing these rules for quite a while now.

**ANY REFACTORS SHOULD NOT BE MADE IN BULK** and should instead be made in small PRs targeting specific files or small related groups of files. Large refactors are significantly more likely to cause conflicts with other open PRs and take significantly longer to review.

PR #1107, opened 2025-09-07
fix: properly delete tag_parents row when deleting tag

### LINKED ISSUE #1105: [Bug]: TagParents not being deleted when deleting Tag
### Checklist

### TagStudio Version

main branch (Alpha 9.5.4+)

macOS 15.6.1

### Description

TagParents are not being deleted from the `tag_parents` table in the database when deleting a Tag from the `tags` table. I came across this error before #1103 was merged but didn't track down the cause until afterwards.

### Expected Behavior

When a tag is deleted from the database, all references of that tag should also be removed. This includes any rows in `tag_parents` in which that tag's ID was involved in a relationship.

Any current libraries that are affected by this issue also need to be repaired and have these invalid tag_parents deleted.

### Steps to Reproduce

1. Create tag, example "Parent"
2. Create tag, example "Child"
3. Add Parent tag to Child's parent tags
4. Add Parent tag to an entry
5. In the tag manager, delete the "Parent" tag
6. Click on the entry to view the traceback in the console, or investigate the DB to see that the tag_parent row is still present

### Logs

When viewing an entry that has a tag referencing a deleted parent:
```
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/qt/views/preview_panel_view.py", line 165, in set_selection
    self._fields.update_from_entry(entry_id)
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/qt/mixed/field_containers.py", line 116, in update_from_entry
    self.update_granular(entry.tags, entry.fields, update_badges)
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/qt/mixed/field_containers.py", line 126, in update_granular
    categories = self.get_tag_categories(entry_tags)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/qt/mixed/field_containers.py", line 173, in get_tag_categories
    hierarchy_tags = self.lib.get_tag_hierarchy(t.id for t in tags)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/core/library/alchemy/library.py", line 1608, in get_tag_hierarchy
    tag.parent_tags = {all_tags[p] for p in all_tag_parents.get(tag.id, [])}
                       ~~~~~~~~^^^
KeyError: 1000
```

CyanVoxel:
### Summary
- Fixes #1105 by properly removing `tag_parents` rows when deleting tags *(presumably caused by the parent_id child_id swap, however the ORM logic needed to be further modified in order to even function so it could have been pre-existing too)*
- Fixes outdated imports caused by #1095 that were missed due to being in the middle of the file and mypy not being sufficient to catch these (see #1061)
- Fixed incompatible type hints caused by #1103
- Repairs existing databases affected by this issue, bumping the minor version to **102**
    - **NOTE:** This is a one-time operation. If a user were to open their library in an older version of TS, then create the same deletion scenario as before, then those rows would stick around in their DB. The condition is now error handled, but the data would still remain. My personal recommendation: **just don't do this.**

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality


RustyNova016:
That would be easier with `ON DELETE CASCADE`, and probably set the table to strict mode for it work

CyanVoxel:
> That would be easier with `ON DELETE CASCADE`, and probably set the table to strict mode for it work

That's likely for the best, however I'm unable to get it working with the ORM we're using. The [SQLAlchemy documentation](https://docs.sqlalchemy.org/en/20/orm/cascades.html) is denser than a tungsten pound cake and when it comes to our schema code I can't tell what's intentional and what's nonsensical. I can make a new task issue to eventually use this, but for now I feel my hands are a bit tied for time and this fix, while not the most ideal, is at least something to get out for tomorrow's patch.

PR #1110, opened 2025-09-07
fix: revert mnemonic removal logic

### LINKED ISSUE #1100: [Bug]: Escaped Ampersands in Translations Malfunctioning
### Checklist

### TagStudio Version

Alpha 9.5.4

macOS 15.6.1

### Description

Translation strings that have an escaped ampersand (e.g. "&Search && Relink", which also includes a mnemonic) are not properly showing up in the UI:
Image
Image

The English strings affected are:
```json
"entries.unlinked.search_and_relink": "&Search && Relink",
"tag.create_add": "Create && Add \"{query}\"",
```

### Expected Behavior

An escaped ampersand string(such as "&Search && Relink" or "Create && Add") should display with a single ampersand in the UI and without the second ampersand in the pair becoming a mnemonic:

Image
Image

### Steps to Reproduce

1. Visit a the "Fix Unlinked Entries" modal to see the "Search & Relink" text or type in a tag name that doesn't exist in the "Manage Tags" or "Add Tag" modals to see the "Create & Add" text and observe the mangled ampersands in the translations.

### Logs

N/A

CyanVoxel:
### Summary
Fixes #1100 by reverting the mnemonic removal logic to the previous method that was used for macOS-specific code.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/qt/mnemonics.py
+++ b/src/tagstudio/qt/mnemonics.py
@@ -8,21 +8,7 @@ from PySide6.QtWidgets import QMenu
 
 def remove_mnemonic_marker(label: str) -> str:
     """Remove existing accelerator markers (&) from a label."""
-    result = ""
-    skip = False
-    for i, ch in enumerate(label):
-        if skip:
-            skip = False
-            continue
-        if ch == "&":
-            # escaped ampersand "&&"
-            if i + 1 < len(label) and label[i + 1] == "&":
-                result += "&"
-                skip = True
-            # otherwise skip this '&'
-            continue
-        result += ch
-    return result
+    return label.replace("&&", "<ESC_AMP>").replace("&", "", 1).replace("<ESC_AMP>", "&&")
 
 
 # Additional weight for first character in string


src/tagstudio/qt/mnemonics.py
```diff
@@ -8,21 +8,7 @@
...
-    result = ""
-    skip = False
-    for i, ch in enumerate(label):
-        if skip:
-            skip = False
-            continue
-        if ch == "&":
-            # escaped ampersand "&&"
-            if i + 1 < len(label) and label[i + 1] == "&":
-                result += "&"
```
Computerdores, line 20: Can you try changing this to `result += "&&"`? I think that should fix it too and would technically be more robust

CyanVoxel: That also seems to fix it - It was just difficult to parse the logic and see what was going wrong where. I can go ahead and make this change though, I'm not the biggest fan of using a placeholder like in the original logic, even if it did simplify it to an easier to follow one-liner

Computerdores (approved):

PR #1111, opened 2025-09-07
fix: use first file as epub thumbnail instead of the last.

Sola-ris:
### Summary

If an ePub contains no ComicInfo.xml the first image should be used as the fallback thumbnail.
Currently, the last image will be used due to a missing `break` after the first image is loaded.

Fixes https://github.com/TagStudioDev/TagStudio/issues/1051 (at least for zip based ePubs)

#### Before
before

#### After
after

#### Files used for testing

* [NoComicInfo.cbz.zip](https://github.com/user-attachments/files/22198469/NoComicInfo.cbz.zip)
* [NoPages.cbz.zip](https://github.com/user-attachments/files/22198471/NoPages.cbz.zip)

Again renamed because of GitHub upload limitiations.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/qt/previews/renderer.py
+++ b/src/tagstudio/qt/previews/renderer.py
@@ -885,6 +885,7 @@ class ThumbRenderer(QObject):
                         ):
                             image_data = zip_file.read(file_name)
                             im = Image.open(BytesIO(image_data))
+                            break
         except Exception as e:
             logger.error("Couldn't render thumbnail", filepath=filepath, error=type(e).__name__)
 


Computerdores (approved):
Was about to mention this when @CyanVoxel merged the prior PR ^^

PR #1113, opened 2025-09-08
fix: apply unwrap where necessary

Computerdores:
### Summary
Progress on #1061.

Applies `unwrap` in bunch of places where it was necessary. Also contains some other small things I fixed when I came across them.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [ ] Basic functionality

### DIFF
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -1856,11 +1856,11 @@ class Library:
             if isinstance(key, LibraryPrefs):
                 return unwrap(
                     session.scalar(select(Preferences).where(Preferences.key == key.name))
-                ).value
+                ).value  # pyright: ignore[reportUnknownVariableType]
             else:
                 return unwrap(
                     session.scalar(select(Preferences).where(Preferences.key == key))
-                ).value
+                ).value  # pyright: ignore[reportUnknownVariableType]
 
     # TODO: Remove this once the 'preferences' table is removed.
     @deprecated("Use `get_version() for version and `ts_ignore` system for extension exclusion.")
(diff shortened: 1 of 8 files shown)


Computerdores:
tests pass for me locally unfortunately. From what I can see in the logs here, the error is not in my PR per sé, it was just not noticed before because the unwrap was missing

CyanVoxel:
> tests pass for me locally unfortunately. From what I can see in the logs here, the error is not in my PR though it was just not noticed before because the unwrap was missing

I have the same tests failing for me locally (macOS)

Computerdores:
since this is only meant to be a refactor I have reverted the small changes that caused the tests to fail. Determining whether those are bugs or not will be left for a future PR

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1850,13 +1850,17 @@ def set_version(self, key: str, value: int) -> None:
...
+    def prefs(self, key: str | LibraryPrefs):
         # load given item from Preferences table
         with Session(self.engine) as session:
             if isinstance(key, LibraryPrefs):
-                return session.scalar(select(Preferences).where(Preferences.key == key.name)).value  # pyright: ignore
+                return unwrap(
+                    session.scalar(select(Preferences).where(Preferences.key == key.name))
+                ).value
             else:
-                return session.scalar(select(Preferences).where(Preferences.key == key)).value  # pyright: ignore
```
CyanVoxel: Is there a reason for restoring the three warnings for this method?
image

Computerdores: I am still not seeing that issue, but I have readded the comment while I debug my setup

CyanVoxel: I've still got the two `reportUnknownVariableType` warnings here
image

Computerdores: should be fixed fr

CyanVoxel (approved):

PR #1114, opened 2025-09-08
fix: renderer type fixes

Computerdores:
### Summary

Mostly type fixes in renderer also some other small ones. Split from #1113 to keep PR size on the smaller end.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [ ] Basic functionality


Computerdores:
fixed merge conflict

src/tagstudio/qt/previews/renderer.py
```diff
@@ -970,7 +971,7 @@ def _font_short_thumb(self, filepath: Path, size: int) -> Image.Image:
             filepath (Path): The path of the file.
             size (tuple[int,int]): The size of the thumbnail.
         """
-        im: Image.Image = None
+        im: Image.Image
```
CyanVoxel: This change causes an error on line 1025, among others:
```
Error calling Python override of QThread::run(): Traceback (most recent call last):
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/qt/ts_qt.py", line 140, in run
    job[0](*job[1])
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/qt/previews/renderer.py", line 1499, in render
    image = self._render(
            ^^^^^^^^^^^^^
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/qt/previews/renderer.py", line 1684, in _render
    image = self._font_short_thumb(_filepath, adj_size)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/qt/previews/renderer.py", line 1025, in _font_short_thumb
    return im  # pyright: ignore[reportPossiblyUnboundVariable]
           ^^
UnboundLocalError: cannot access local variable 'im' where it is not associated with a value
```

`None` is expected as valid state for `im` in much of the thumbnail renderer code to represent a state where the image could not be rendered and needs to be replaced with a default icon before being returned. Similar issues are also arising elsewhere in this file from removing the `None` assignment (and type hint, if applicable). The `None` assignment also prevents the variables from becoming unbound.

Computerdores: should be fixed now, I was not aware of that aspect of this code

Computerdores:
for me all pyright issues are fixed now. Only thing I am not a particular fan of is the number of ignore comments on `get` calls to the ResourceManager, but didn't know a better solution off the top of my head

CyanVoxel:
> Only thing I am not a particular fan of is the number of ignore comments on `get` calls to the ResourceManager, but didn't know a better solution off the top of my head

My longer term solution is to overhaul that thing to have different getters deepening on the type you expect. I have *no* idea what's going on with the type stuff over there, only that it's guaranteed that no other part of the code can ever be happy with the type it spits out...

src/tagstudio/qt/previews/renderer.py
```diff
@@ -665,7 +666,7 @@ def _audio_album_thumb(filepath: Path, ext: str) -> Image.Image | None:
                     artwork = Image.open(BytesIO(flac_covers[0].data))
             elif ext in [".mp4", ".m4a", ".aac"]:
                 mp4_tags: mp4.MP4 = mp4.MP4(filepath)
-                mp4_covers: list = mp4_tags.get("covr")
+                mp4_covers: list = mp4_tags.get("covr")  # pyright: ignore[reportAssignmentType]
```
CyanVoxel: Just out of curiosity, are you also seeing the `reportUnknownVariableType` warning for these lines?
image

CyanVoxel (approved):

src/tagstudio/qt/previews/renderer.py
Computerdores: I didn't otherwise I would have set it to ignored. Only have nvim available today though, so can't check

PR #1129, opened 2025-09-14
perf: Optimize searching tags with DB indexes

### LINKED ISSUE #1352: [Bug]: Reduced search performance in 9.5.7
### Checklist

### TagStudio Version

9.5.7

Win11

### Description

i decided to try the new version however it seems to run slower, significantly so

(in a big library) Searching freezes the program for a little bit, opening it takes 5 minits (compared to .08 seconds in 9.5.6), adding tags also briefly freezes the program compared to before where all of these was basicly instant
(this "big" library has 36k entries for context)

checking a library with only 7k entries it takes 1.6 seconds to fully open in .7   , in .6 it only takes .01 seconds
 
seems there may be an optimization gremlin somewhere  if its making it take this long

### Expected Behavior

it should be about as smooth as 9.5.6 

### Steps to Reproduce

couldn't tell you if there's more specific details but Large libraries seem  to be a requirement

### Logs

_No response_

TheBobBobs:
### Summary
Create indexes on columns commonly used in queries.
`Tag.name, Tag.shorthand` finding tags by name
`TagParent.child_id` fetching tag hierarchies
`TagEntry.entry_id` fetching entries with their tags

Change query in `Library.search_tags` so it will use above indexes.
Sort results before applying limit to prevent truncating tags that should be prioritized.

### Tasks Completed
-   Platforms Tested:
    -   [x] Linux x86
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -75,7 +75,6 @@ from tagstudio.core.library.alchemy.constants import (
     DB_VERSION_LEGACY_KEY,
     JSON_FILENAME,
     SQL_FILENAME,
-    TAG_CHILDREN_QUERY,
 )
 from tagstudio.core.library.alchemy.db import make_tables
 from tagstudio.core.library.alchemy.enums import (
@@ -1071,13 +1070,10 @@ class Library:
     def search_tags(self, name: str | None, limit: int = 100) -> list[set[Tag]]:
         """Return a list of Tag records matching the query."""
         with Session(self.engine) as session:
-            query = select(Tag).outerjoin(TagAlias).order_by(func.lower(Tag.name))
-            query = query.options(
-                selectinload(Tag.parent_tags),
-                selectinload(Tag.aliases),
-            )
-            if limit > 0:
-                query = query.limit(limit)
+            query = select(Tag.id, Tag.name).outerjoin(TagAlias)
+
+            if limit > 0 and (not name or len(name) == 1):
+                query = query.limit(limit).order_by(func.lower(Tag.name))
 
             if name:
                 query = query.where(
@@ -1088,35 +1084,24 @@ class Library:
                     )
                 )
 
-            direct_tags = set(session.scalars(query))
-            ancestor_tag_ids: list[Tag] = []
-            for tag in direct_tags:
-                ancestor_tag_ids.extend(
-                    list(session.scalars(TAG_CHILDREN_QUERY, {"tag_id": tag.id}))
-                )
-
-            ancestor_tags = session.scalars(
-                select(Tag)
-                .where(Tag.id.in_(ancestor_tag_ids))
-                .options(selectinload(Tag.parent_tags), selectinload(Tag.aliases))
-            )
-
-            res = [
-                direct_tags,
-                {at for at in ancestor_tags if at not in direct_tags},
-            ]
-
+            tags = list(session.execute(query))
             logger.info(
                 "searching tags",
                 search=name,
                 limit=limit,
                 statement=str(query),
-                results=len(res),
+                results=len(tags),
             )
 
-            session.expunge_all()
+            tags.sort(key=lambda t: t[1].lower())
+            if limit <= 0:
+                limit = len(tags)
+            tag_ids = [t[0] for t in tags[:limit]]
 
-            return res
+            hierarchy = self.get_tag_hierarchy(tag_ids)
+            direct_tags = {hierarchy.pop(id) for id in tag_ids}
+            ancestor_tags = set(hierarchy.values())
+            return [direct_tags, ancestor_tags]
 
     def update_entry_path(self, entry_id: int | Entry, path: Path) -> bool:
         """Set the path field of an entry.
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -1073,7 +1073,7 @@ class Library:
         name = name.lower()
 
         def sort_key(text: str):
-            return (not text.startswith(name), text)
+            return (not text.startswith(name), len(text), text)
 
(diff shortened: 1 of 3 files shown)


src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1054,55 +1067,61 @@ def search_library(
...
-            query = query.options(
-                selectinload(Tag.parent_tags),
-                selectinload(Tag.aliases),
-            )
-            if limit > 0:
-                query = query.limit(limit)
+            query = select(Tag.id, Tag.name)
+
+            if limit > 0 and not name:
+                query = query.limit(limit).order_by(func.lower(Tag.name))
```
Computerdores: This causes the sorting to happen after truncating the results, which differs from the behaviour when searching, no?

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1054,55 +1067,61 @@ def search_library(
...
+
+            tags.sort(key=lambda t: sort_key(t[1]))
+            seen_ids = set()
+            tag_ids = []
+            for row in tags:
+                id = row[0]
+                if id in seen_ids:
+                    continue
+                tag_ids.append(id)
+                seen_ids.add(id)
```
Computerdores: Couldn't this be written as the following?
```py
tags = dict(tags)
tag_ids = sorted(tags.keys(), key=lambda t: sort_key(tags[t]))
del tags # not sure if this is makes a diff, but `tags` could become quite large and triggering gc on it sooner can't hurt
```
this is both simpler code wise and should be faster by only sorting the deduplicated list

src/tagstudio/qt/mixed/tag_search.py
```diff
@@ -218,32 +218,15 @@ def update_tags(self, query: str | None = None):
...
+        direct_tags, ancestor_tags = self.lib.search_tags(name=query, limit=tag_limit)
 
-        if query and query.strip():
-            for tag in raw_results:
-                if tag.name.lower().startswith(query_lower):
-                    priority_results.add(tag)
+        all_results = [t for t in direct_tags if t.id not in self.exclude]
+        for tag in ancestor_tags:
+            if tag.id not in self.exclude:
+                all_results.append(tag)
```
Computerdores: This code previously handled `self.exclude` being None, is there a reason you removed that?

Computerdores:
```suggestion
        all_results = [t for t in direct_tags if t.id not in self.exclude]
        all_results += [t for t in ancestor_tags if t.id not in self.exclude]
```

src/tagstudio/core/library/alchemy/library.py
TheBobBobs: This is so it will use the order from `Tag.name` or `TagAlias.name` depending on which comes first for each tag.

src/tagstudio/qt/mixed/tag_search.py
TheBobBobs: Its type is `list[int]` and I couldn't find any code that could cause it to be None.

src/tagstudio/core/library/alchemy/library.py
TheBobBobs: Yes this was not sorting by `len(text)` before truncating. The previous implementation only sorted priority results by len so I've updated `sort_key` to do that. Which will make this `order_by` statement and `sort_key` produce the same results when no query is provided.

src/tagstudio/qt/mixed/tag_search.py
TheBobBobs: Ended up doing this to avoid creating extra lists.
`all_results.extend(t for t in ancestor_tags if t.id not in self.exclude)`

Computerdores: good reason ^^, I missed that

src/tagstudio/core/library/alchemy/library.py
Computerdores: in that case the following should work since `dict.keys()` maintains insertion order and dict deduplicates by key:
```py
tags.sort(key=lambda t: sort_key(t[1]))
tag_ids = dict(tags).keys()  # get the deduplicated list of ids
```

Computerdores: other than that I think this is good to merge

Computerdores (approved):

Computerdores:
Sorry for the delay on the review had to work on my thesis ^^

CyanVoxel:
Sorry for my own inactivity on this, I haven't pulled this so far since I wanted to do some rigorous testing to ensure that no issues slipped though with query performance. I'll get back to this in the near future and let you know if I find anything quirky on my end

mm12:
A quick look - this does fix my problem mentioned [here](https://github.com/TagStudioDev/TagStudio/issues/1239#issuecomment-3981099751) - and with results that make it hard to believe such a huge performance loss existed in the first place.

Testing with complicated queries that rely on changed i modded in (such as path regex search and date-added search) seem to have about the same performance.

CyanVoxel (changes requested):
I'm so sorry that I haven't gotten back to this as soon as I intended. I hope that you are still willing to work on this, but if not please let me know at your earliest convenience. 
***
General search performance seems to have greatly improved while retaining the same results from the limited query tests I've done against the `main` branch version. However, I've noticed that the tag search now returns undesirable results compared to the current live behavior:

`main` behavior, returns tags with text "blue" plus child tags (tags with blue as a parent):
Image

#1129 behavior - returns tags with text "blue" and what seems to be ancestor tags or something adjacent:
Image

CyanVoxel (approved):
Awesome! Thank you so much for the quick changes! And thank you again for bearing with me with the long review delay! 😅

Issue #951, opened 2025-06-07
[Bug]: Search `A AND A` returns no results


Computerdores:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

git (c100babd9fc80829703d4d2da2985e873e136b02)

### Operating System & Version

Any

### Description

Searching for `A AND A`, where A is any tag in the library, returns no results, even if a search for `A` does. This breaks the expected search semantics.

### Expected Behavior

A search for `A AND A` and a search for `A` should always return the same results.

A regression test for this should also be added.

### Steps to Reproduce

1. Open a Library
2. Search for `A AND A`, with A replaced by any tag
3. Observe lack of results

PR #1138, opened 2025-09-21
fix: Fix searching `A AND A` returning no results

TrigamDev:
Fixes searching `<Tag A> AND <Tag A>` returning no results, when it should return the same results as `<Tag A>` ([#951](https://github.com/TagStudioDev/TagStudio/issues/951)).

I'm not very familiar with the backend and haven't done very in-depth testing, so please point out if this causes any unintended side effects.

### Summary
Tweaked the `__separate_tags()` method in `src/tagstudio/core/library/alchemy/visitors.py` to, rather than a list, use a set for collecting the tag IDs of each term of an AST `ANDList` and `ORList` to prevent duplicate tag IDs from being returned.

#### Cause of Bug
The SQL query built from the search `<Tag A> AND <Tag A>` will always fail due to this section:
```
WHERE tag_entries.tag_id IN (1, 1)
GROUP BY tag_entries.entry_id
HAVING count(DISTINCT tag_entries.tag_id) = 2
```
Since it's getting rows where `count(DISTINCT tag_entries.tag_id) = 2`, but it's only searching for rows with a single tag ID, `count(DISTINCT tag_entries.tag_id)` will always equal `1`.

This doesn't apply just to `ANDList`s containing two of the same tag, any `ANDList` containing _at least_ two duplicate tags will cause the query to fail. So a search such as `<Tag A> AND <Tag B> AND <Tag A> AND <Tag A>` would also fail.
However, if the duplicate tag is ambiguous, the SQL query built filters entries in a fundamentally different way that doesn't suffer the same issue, and the search acts as expected.

### Tasks Completed
-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/core/library/alchemy/visitors.py
+++ b/src/tagstudio/core/library/alchemy/visitors.py
@@ -146,7 +146,7 @@ class SQLBoolExpressionBuilder(BaseVisitor[ColumnElement[bool]]):
     def __separate_tags(
         self, terms: list[AST], only_single: bool = True
     ) -> tuple[list[int], list[ColumnElement[bool]]]:
-        tag_ids: list[int] = []
+        tag_ids: set[int] = set()
         bool_expressions: list[ColumnElement[bool]] = []
 
         for term in terms:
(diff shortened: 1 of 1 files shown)


CyanVoxel:

PR #1139, opened 2025-09-22
feat: add hidden tags

### LINKED ISSUE #49: [Feature Request]: Hidden Tags
REcently I reset my windows system and started fresh. This always brings new possiblities to try other/new things and there's nothing newer than an alpha project which just released. Anyways, I'm currently in the process of using TagStudio as my main form of storing and searching for files, but I've come over an "inconvenience". A very minor one at that, however I still got this idea from it:

"Conditional Statements" would be a system where you can specify what should happen to an entry if it has a certain tag. The main point, imo, would be to remove certain entries from being viewed, except if you explicitly search for them. The reasoning for this is, I have a few files that I don't want people to see when I'd stream or similar. I tagged them with "Personal" and it would be really cool if it were possible to hide such entries.

This might sound very insignificant... and probably is, but I'm sure one could extend that system to also have other abilities which I could not have imagined.

TrigamDev:
### Summary
Adds the ability to set tags as 'hidden', which will cause entries tagged with them to be excluded from searches.
This also bumps the database version to `103` due to the database changes.

image

image

### Tasks Completed
- Implementation
    - [x] Add `is_hidden` field to the `tags` table
    - [x] Set the Archived tag to be hidden
    - [x] Add "Is Hidden" checkbox to the edit tag panel
    - [x] Exclude entries tagged with hidden tags from searches
    - [x] Add checkbox for allowing hidden tags in searches
    - [x] Document database changes in docs 
-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality

---

Closes #49

### DIFF
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -1841,19 +1841,6 @@ class Library:
             )
             session.add(parent_tag)
 
-    def get_hidden_tag_ids(self) -> set[int]:
-        """Get a set containing the IDs of all of the hidden tags."""
-        hidden_tag_ids: set[int] = set()
-
-        with Session(self.engine) as session:
-            root_hidden_tag_ids = session.scalars(
-                select(Tag.id).where(Tag.is_hidden == True)  # noqa: E712
-            ).all()
-            for root_hidden_tag_id in root_hidden_tag_ids:
-                hidden_tag_ids.add(root_hidden_tag_id)
-
-        return hidden_tag_ids
-
     def get_version(self, key: str) -> int:
         """Get a version value from the DB.
 
--- a/src/tagstudio/resources/translations/en.json
+++ b/src/tagstudio/resources/translations/en.json
@@ -146,7 +146,7 @@
     "home.thumbnail_size.mini": "Mini Thumbnails",
     "home.thumbnail_size.small": "Small Thumbnails",
     "home.thumbnail_size": "Thumbnail Size",
-    "home.show_hidden_entries": "Show Hidden Entries?",
+    "home.show_hidden_entries": "Show Hidden Entries",
     "ignore.open_file": "Show \"{ts_ignore}\" File on Disk",
     "json_migration.checking_for_parity": "Checking for Parity...",
     "json_migration.creating_database_tables": "Creating SQL Database Tables...",


src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1003,13 +1039,28 @@ def search_library(
...
+                hidden_tag_ids = self.get_hidden_tag_ids()
+                hidden_tag_constraints: list[Constraint] = list(
+                    map(
+                        lambda tag_id: Constraint(ConstraintType.TagID, str(tag_id), []),
+                        hidden_tag_ids,
+                    )
+                )
+                hidden_tag_ast = Not(ORList(hidden_tag_constraints))
+
+                ast = hidden_tag_ast if not ast else ANDList([search.ast, hidden_tag_ast])
```
CyanVoxel: It shouldn't be necessary to grab each hidden tag ID to compare against for each query, instead the query should be able to be amended with something like a `WHERE` expression that checks if it has any tags with the `is_hidden` property directly. The following should work:
```py
if search.exclude_hidden_entries:
    statement = statement.where(~Entry.tags.any(Tag.is_hidden))
```
The `.any()` method should return`TRUE` if any of the `Entry.tags` have `.is_hidden`, and the `~` will negate the whole expression so it evaluates to `FALSE` if the `is_hidden` property is present and `TRUE` otherwise.

src/tagstudio/resources/translations/en.json
```diff
@@ -146,6 +146,7 @@
     "home.thumbnail_size.mini": "Mini Thumbnails",
     "home.thumbnail_size.small": "Small Thumbnails",
     "home.thumbnail_size": "Thumbnail Size",
+    "home.exclude_hidden_entries": "Exclude Hidden Entries?",
```
CyanVoxel: Personally I feel this phrasing would make more sense if it were inverted, i.e. "Show Hidden Entries". This would also require the logic of the checkbox to be flipped.
image

TrigamDev:
Made the requested changes.

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1800,6 +1841,19 @@ def update_parent_tags(self, tag: Tag, parent_ids: list[int] | set[int], session
...
+
+        with Session(self.engine) as session:
+            root_hidden_tag_ids = session.scalars(
+                select(Tag.id).where(Tag.is_hidden == True)  # noqa: E712
+            ).all()
+            for root_hidden_tag_id in root_hidden_tag_ids:
+                hidden_tag_ids.add(root_hidden_tag_id)
+
+        return hidden_tag_ids
+
```
CyanVoxel: This method is no longer being used

CyanVoxel (approved):
Fixed the merge conflict and removed the unused method, thank you again for your work on this!

Issue #1147, opened 2025-09-24
[Feature Request]: Thumbnail preview for `pdn` files


TrigamDev:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

There are currently no thumbnail previews for `pdn` files (the project file format for [Paint.NET](https://www.getpaint.net/)).

### Solution

Add a thumbnail renderer for `pdn` files.

### Alternatives

No, `pdn` files currently have the default, blank thumbnail.

PR #1149, opened 2025-09-24
feat: render `.pdn` thumbnails.

Sola-ris:
### Summary

Add support for rendering `.pdn` thumbnails.
Fixes https://github.com/TagStudioDev/TagStudio/issues/1147

`.pdn` files have an XML header that contains the thumbnail as a base64 encoded `.png` file.
The file layout is the following:
* 4 byte magic number (`PDN3`)
* 24-bit little-endian int specifying the header's length
* The XML header
* The actual image data which we don't care about

The header itself looks like this:
```xml
<pdnImage width="800" height="600" layers="1" savedWithVersion="5.109.9343.2610">
    <custom>
        <thumb png="base64 string" />
    </custom>
</pdnImage>
```

#### Sources for the file format
* http://justsolve.archiveteam.org/wiki/Paint.NET_image
    * Apparently there's a compresed version of the file format, but I couldn't get Paint.NET to produce that one
* https://github.com/rivy/OpenPDN
    * https://github.com/rivy/OpenPDN/blob/master/src/Data/Document.cs#L64-L120
    * https://github.com/rivy/OpenPDN/blob/master/src/DocumentWorkspace.cs#L2594-L2610
* https://github.com/addisonElliott/pypdn/blob/master/pypdn/reader.py#L142-L143
* The test file opened in a text editor

#### Before
before

#### After
after

#### File used for testing
[pdn.zip](https://github.com/user-attachments/files/22522512/pdn.zip) (The actual file is archived since GitHub doesn't allow uploading `.pdn` files)

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/core/media_types.py
+++ b/src/tagstudio/core/media_types.py
@@ -651,6 +651,7 @@ class MediaCategories:
         MODEL_TYPES,
         OPEN_DOCUMENT_TYPES,
         PACKAGE_TYPES,
+        PAINT_DOT_NET_TYPES,
         PDF_TYPES,
         PLAINTEXT_TYPES,
         PRESENTATION_TYPES,


TrigamDev:
> `.pdn` files have an XML header that contains the thumbnail as a base64 encoded `.png` file.

Well that's convenient

TrigamDev:
Transparent backgrounds appear to be interpreted as pure black in the thumbnail preview, but not in the sidebar preview.

Sola-ris:
I'll take a look

Sola-ris:
Regular PNGs have a background added, if they're in RGBA. I've given the thumbnail the same treatment
with-background

TrigamDev:
I haven't noticed any other images in my library with a transparent background having a background added. And I'd argue that if it is a behavior in other file types, it should be fixed as it's very unintuitive and unwanted behavior.

Sola-ris:
Happens for [regular images](https://github.com/TagStudioDev/TagStudio/blob/main/src/tagstudio/qt/previews/renderer.py#L1128-L1131) and [exr image](https://github.com/TagStudioDev/TagStudio/blob/main/src/tagstudio/qt/previews/renderer.py#L1106-L1110).
I have no opinion on it, but I don't think it's something that should be changed as part of this issue.

TrigamDev:
Yeah I agree that it should be a separate issue..
Thanks for your work on the thumbnail rendering btw, didn't expect someone to start working on these so quickly.

Sola-ris:
You're welcome :D

CyanVoxel (approved):
Sorry for the delay on getting this approved - thank you so much for your work on this!

Issue #1145, opened 2025-09-24
[Feature Request]: Thumbnail preview for `mdp` files


TrigamDev:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

There are currently no thumbnail previews for `mdp` files (the project file format for [Medibang Paint](https://medibangpaint.com/en/) and [FireAlpaca](https://firealpaca.com/)).

### Solution

Add a thumbnail renderer for `mdp` files.

### Alternatives

No, `mdp` files currently have the default, blank thumbnail.

### `mdp` Format

While I don't know *exactly* what the schema is, I've noticed a few things at a glance:
- Begins with `mdipack`.
- Contains an XML block describing basic information about the project file, including dimensions, timestamps, and layer data.
Example XML block:
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<Mdiapp width="1368" height="2261" dpi="350" checkerBG="false" bgColorR="116" bgColorG="124" bgColorB="170">
    <CreateTime time="1708562941" timeString="2024-02-21T19:49:01" />
    <UpdateTime time="1717740744" timeString="2024-06-07T02:12:24" rev="458" />
    <Thumb width="154" height="256" bin="thumb" />
    <Snaps />
    <Guides />
    <ICCProfiles enabled="false" cmykView="false" blackPoint="true" renderingIntent="perceptual" />
    <Animation enabled="false" showNextPrev="true" baseLayer="false" fps="24" />
    <Layers active="9">
        <Layer name="Sketch 1" width="1368" height="2261" mode="normal" alpha="28" id="2" parentId="15" binType="2" bin="layer0img" type="32bpp" />
        <Layer name="Sketch 2" width="1368" height="2261" mode="normal" alpha="40" id="10" parentId="15" visible="false" binType="2" bin="layer1img" type="32bpp" />
        <Layer name="Body" width="1368" height="2261" mode="normal" alpha="102" id="17" parentId="15" binType="2" bin="layer2img" type="32bpp" />
        <Layer name="Left Arm" width="1368" height="2261" mode="normal" alpha="102" id="18" parentId="15" binType="2" bin="layer3img" type="32bpp" />
[...]

PR #1153, opened 2025-09-28
feat: render `.mdp` thumbnails.

Sola-ris:
### Summary
Add support for rendering `.mdp` thumbnails.
Fixes https://github.com/TagStudioDev/TagStudio/issues/1145

#### File layout
Mostly translated from the repo linked below.

The first 20 bytes are a binary header called `TMDIPack`.
```c
struct TMDIPack {
    char header[8]; // "mdipack "
    unsigned long version; // always 0
    unsigned long mdiSize; // size of the XML header
    unsigned long version; // size of the actual project data
}
```

After the binary header comes an XML header called `MDI`.
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<Mdiapp width="1600" height="1200" dpi="350" checkerBG="true" bgColorR="255" bgColorG="255" bgColorB="255">
    <CreateTime time="1758967935" timeString="2025-09-27T12:12:15" />
    <UpdateTime time="1758967935" timeString="2025-09-27T12:12:15" rev="1" />
    <Thumb width="256" height="192" bin="thumb" />
    <Snaps />
    <Guides />
    <ICCProfiles enabled="false" cmykView="false" blackPoint="true" renderingIntent="perceptual" />
    <Layers active="1">
        <Layer ofsx="0" ofsy="0" width="1600" height="1200" mode="normal" alpha="255" visible="true" protectAlpha="false" locked="false" clipping="false" masking="false" maskingType="0" group="-1" id="0" draft="false" parentId="-1" name="Ebene1" binType="2" bin="layer0img" type="32bpp" />
        <Layer ofsx="0" ofsy="0" width="1600" height="1200" mode="normal" alpha="255" visible="true" protectAlpha="false" locked="false" clipping="false" masking="false" maskingType="0" group="-1" id="1" draft="false" parentId="-1" name="Ebene2" binType="2" bin="layer1img" type="32bpp" />
    </Layers>
    <appInfo name="MediBang Paint Pro" rev="92" bitType="64" />
</Mdiapp>
```

The relevant bits are the `Thumb` element and its attributes as well as the amount of `Layer` elements.

After the XML comes the `MDIBIN` which holds the embedded thumbnail as well as the actual project data.
The `MDIBIN` contains with a 132-byte `TPackerHeader` followed by a blob of data whose length is specified in the header.
This repeats for each layer as well as the thumbnail.

The `TPackerHeader` looks like this
```c
struct TPackerHeader {
    char identifier[4]; // 'PAC '

    unsigned long chunkSIze; // Size of the TPackerHeader and layer / thumbnail data
    unsigned long streamType; // 0 for uncompressed, 1 for zlib
    unsigned long streamSize; // Size of the layer / thumbnail data
    unsigned long outSize; // Size after decompression

    unsigned char reserved[48]; // Buffer, all 0
    unsigned char archiveName[64]; // Name of the layer, padded. The one of the thumbnail is called "thumb", otherwise it's the layer's "bin" attribute from the xml header.
}
```

The thumbnail is stored as a continious blob of BGRA image data whose dimensions are specified in the attributes of the `Thumb` element in the XML header.

I _think_ the thumbnail is always the first block in the `MDIBIN`, but I couldn't find anything explicitly saying so.

While the thumbnail is in BGRA, it doesn't seem to store any transparency data. It always shows up with a solid white background, even in Windows explorer.
explorer

I also wanted to document the format in the code, but I always ended up with a docstring several times the size of the actual implementation, so I left it out.

#### Before
before

#### After
after

#### Sources for the file format
I Got most of the info from [this repo](https://github.com/rsuzaki/mdp_format/wiki). It gets one thig wrong, namely the color mode of the thumbnail. It claims its in ARGB, when it's actually in BGRA. I found that out via [this GIMP plugin for `.mdp` files](https://github.com/weeb-poly/gimp-file-mdp-plugin/blob/main/src/file-mdp-thumb-load.c#L73)

#### File used for testing
[layers-color-transparency.zip](https://github.com/user-attachments/files/22581913/layers-color-transparency.zip)
Zipped because of GitHub upload limitations,

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/docs/library/index.md
+++ b/docs/library/index.md
@@ -86,20 +86,21 @@ Audio thumbnails will default to embedded cover art (if any) andfallback to gene
 
 Preview support for office documents or well-known project file formats varies by the format and whether or not embedded thumbnails are available to be read from. OpenDocument-based files are typically supported.
 
-| Filetype                      | Extensions            | Preview Type                                                               |
-| ----------------------------- | --------------------- | -------------------------------------------------------------------------- |
-| Blender                       | `.blend`, `.blend<#>` | Embedded thumbnail :material-alert-circle:{ title="If available in file" } |
-| Keynote (Apple iWork)         | `.key`                | Embedded thumbnail                                                         |
-| Krita[^3]                     | `.kra`, `.krz`        | Embedded thumbnail :material-alert-circle:{ title="If available in file" } |
-| MuseScore                     | `.mscz`               | Embedded thumbnail :material-alert-circle:{ title="If available in file" } |
-| Numbers (Apple iWork)         | `.numbers`            | Embedded thumbnail                                                         |
-| OpenDocument Presentation     | `.odp`, `.fodp`       | Embedded thumbnail                                                         |
-| OpenDocument Spreadsheet      | `.ods`, `.fods`       | Embedded thumbnail                                                         |
-| OpenDocument Text             | `.odt`, `.fodt`       | Embedded thumbnail                                                         |
-| Pages (Apple iWork)           | `.pages`              | Embedded thumbnail                                                         |
-| PDF                           | `.pdf`                | First page render                                                          |
-| Photoshop                     | `.psd`                | Flattened image render                                                     |
-| PowerPoint (Microsoft Office) | `.pptx`, `.ppt`       | Embedded thumbnail :material-alert-circle:{ title="If available in file" } |
+| Filetype                             | Extensions            | Preview Type                                                               |
+| ------------------------------------ | --------------------- | -------------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


TrigamDev:
Holy shit, thank you so much, I had no idea there was an existing project that already worked out how the file format was laid out

CyanVoxel (approved):
Looks good as far as I can tell, thank you so much for you work on this!

PR #1166, opened 2025-10-13
feat: update notification

Computerdores:
### Summary
Displays a notification to the user when a newer tagstudio version is available.

### Tasks Completed
-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality


Computerdores:
I am not sure about the current logic for whether it is outdated or not.
Currently if any newer version with a suffix (e.g. `1.2.3-pr1`) is available then the notification is displayed (and the comparison is currently broken if both have a suffix).

Alternatively latest versions with suffixes could just generally be ignored. This would mean that if `1.2.3-pr1` is the latest release, `1.2.2` is the latest full release and the user has `1.2.1` installed then no notification would displayed (I think; if RCs are ignored by the API endpoint than RCs don't matter anyways).

TrigamDev:
Is there a reason to not use [semver](https://pypi.org/project/semver/) for comparing the versions? It'd likely be a lot more reliable and consistent than creating a custom implementation for essentially the same thing.

As for versions with suffixes, I think having them be ignored would probably be best. Recommending non-release versions to normal users probably isn't a good idea, and most users who do use non-release versions likely keep an eye on version releases already.

Computerdores:
I have replaced the regex parsing with the semver depency as suggested and have also cleaned up the comparison logic there to be more readable.

Also, the update notification modal now links directly to the page for the latest release (for people who aren't familiar with github).

CyanVoxel (approved):
Everything checks out on my end, thank you for your work on this!

PR #1168, opened 2025-10-18
feat: add URL field type

### LINKED ISSUE #1140: [Feature Request]: URL field type
### Checklist

### Description

As was brought up in #924, it would be a good idea to add a dedicated URL field type, rather than having links be embedded in text fields (though, both could coexist if needed). This would remove the reliance on a regex pattern for detecting URLs (which can never be completely perfect), and allow the user to specify what is and isn't a URL.

### Solution

A URL field type could be added, which functions nearly identically to a text field, but will interpret all content as a URL (even if it isn't a valid URL). This could allow for more niche URL types to be embedded, such as URLs with non-HTTP/HTTPS protocols (for my use case, it would be very helpful to be able to embed `discord://` links).
These fields could potentially also provide functionality for naming the URL, which would be displayed in place of the URL itself.

### Alternatives

#924 works well for most common cases, but as mentioned in the discussion, isn't perfect, and will miss plenty of valid URLs. In my opinion, it's better to add a new URL field type and leave the decision of what is and isn't a URL up to the user.

TrigamDev:
### Summary

- Adds a new `URL` field type, which replaces URL detection in text fields.
  - The value of the URL field is interpreted as a link, regardless of if it's valid or not, giving the user more control and preventing cases where valid URLs aren't detected.
  - The user is also able to choose a title for the link, which will be displayed in place of the URL value.

Closes #1140 

image

image

### Changes Made

> (Headers don't directly correlate with the file paths of the changed files, these are more or less arbitrary)

#### Core
- Added a new field type, `URL`, which takes a URL title and URL value.
- Added a new `url_fields` table to store the new URL fields, having fields for the URL title and URL value.
- The built-in fields `URL` and `SOURCE` have been changed to be of the field type `URL`.
- Entries have been modified to load in the URL fields associated with it, as is the case with text fields and datetime fields.
- Bumps the database version to `104` (in context of #1139 bumping it to `103`).
- Two migration steps have been added for database version `104`:
  - First, both the `URL` and `SOURCE` fields are changed from being of type `TEXT_LINE` to being of type `URL`
  - Second, all text fields that have a type key which is itself of type `URL` (so the two changed above) will be moved over to the `url_fields` table.
- Both URL and text fields, when empty, will have their value stored as `None`, and will account for having `None` values.

#### QT
- Reverts the URL detection made in #924 
- Adds a new `UrlWidget` that extends `TextWidget` which accepts a URL title and URL value and displays an embedded link.
- Adds a new `EditUrl` modal, providing inputs for editing the title and value of a URL field.

#### Mixed
- To account for URL fields being the first field type to have multiple parameters, field saving has been modified to allow an arbitrary number of arguments.
  - The save callback of `PanelModal`s only returning a string, they can now return an arbitrary number of arguments of any type.
    - Not quite related to URL fields, but the save callback has had it's type hint modified to also allow for the return type to be a tuple of an arbitrary length with type `None`. This is because most (if not all) save callbacks had this return type, which kept causing warnings.
  - The `get_content()` method of `PanelModal`s can now return an arbitrary number of arguments of any type instead of just string.
  - Adding and updating fields in the database, instead of taking a single argument for the `value` field, now takes `kwargs`, with each key corresponding to a field in the field's table, and each value corresponding to the value being written.
    - All previous field types have had their `add_field_to_entry` and `update_entry_field` calls to account for this change.
    - Updated the `test_library` and `test_update_entry_with_multiple_identical_fields` tests to account for this change.

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -14,6 +14,10 @@ Thank you so much for showing interest in contributing to TagStudio! Here are a 
 -   **Please don't** create pull requests that consist of large refactors, _especially_ without discussing them with us first. These end up doing more harm than good for the project by continuously delaying progress and disrupting everyone else's work.
 -   If you wish to discuss TagStudio further, feel free to join the [Discord Server](https://discord.com/invite/hRNnVKhF2G)!
 
+<!-- prettier-ignore -->
+!!! note
+    If the fix is small and self-explanatory (i.e. a typo), then it doesn't require an issue to be opened first. Issue tracking is supposed to make our lives easier, not harder. Please use your best judgement to minimize the amount of work for everyone involved.
+
 ### Contribution Checklist
 
 -   I've read the [Feature Roadmap](roadmap.md) page
@@ -25,8 +29,10 @@ Thank you so much for showing interest in contributing to TagStudio! Here are a 
 -   **_I mean it, I've found or created an issue for my feature/fix!_**
 
 <!-- prettier-ignore -->
-!!! note
-    If the fix is small and self-explanatory (i.e. a typo), then it doesn't require an issue to be opened first. Issue tracking is supposed to make our lives easier, not harder. Please use your best judgement to minimize the amount of work for everyone involved.
+!!! failure "Unacceptable Code"
+    The following types of code will NOT be accepted to the project:
+    -   Code that is not yours or does not have a compatible license with TagStudio's [own one](../LICENSE)
+    -   Code that you do not understand and/or cannot explain
 
 ## Creating a Development Environment
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -92,7 +92,7 @@ hide:
 
     ***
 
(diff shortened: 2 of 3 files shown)


TrigamDev:
I'll be using this branch as my primary version of TagStudio for now, so I'll make sure to update here if I come across any issues.

TrigamDev:
> I'll be using this branch as my primary version of TagStudio for now, so I'll make sure to update here if I come across any issues.

Appears that when copying fields, it only copies the `value` field. I'll tweak it real quick to use `**kwargs` instead.

PR #1189, opened 2025-11-03
fix(nix): replace wrapGAppsHook with wrapGAppsHook3

Ambossmann:
### Summary

Replaces all occurrences of wrapGAppsHook with wrapGAppsHook3.

wrapGAppsHook has been aliased to wrapGAppsHook3 for a long time (since NixOS/nixpkgs#307077). Recently (NixOS/nixpkgs#456065) this alias was converted to a throw, breaking the build with newer nixpkgs versions.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/nix/package/default.nix
+++ b/nix/package/default.nix
@@ -6,7 +6,7 @@
   qt6,
   ripgrep,
   stdenv,
-  wrapGAppsHook,
+  wrapGAppsHook3,
 
   pillow-jxl-plugin,
 
@@ -30,7 +30,7 @@ python3Packages.buildPythonApplication {
     # Should be unnecessary once PR is pulled.
     # PR: https://github.com/NixOS/nixpkgs/pull/271037
     # Issue: https://github.com/NixOS/nixpkgs/issues/149812
-    wrapGAppsHook
+    wrapGAppsHook3
   ];
   buildInputs = [
     qt6.qtbase
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -51,7 +51,7 @@ let
       # Should be unnecessary once PR is pulled.
       # PR: https://github.com/NixOS/nixpkgs/pull/271037
       # Issue: https://github.com/NixOS/nixpkgs/issues/149812
-      wrapGAppsHook
+      wrapGAppsHook3
     ];
     buildInputs = with pkgs.qt6; [
       qtbase


poseidon-rises:
I was about to open my own PR for this, guess someone beat me to it. I tested it on my machine and it seems to work without issue, so I got no gripes.

Computerdores:
just a thought: this change comes from [nixpkgs!307077](https://github.com/NixOS/nixpkgs/pull/307077) which also mentions that `wrapGAppsHook3` is for GTK3 while `wrapGAppsHook4` is for GTK4. I have no idea why we even need to do anything regarding GTK since we don't use GTK, but whatever Gnome thing this is about might in the future require us changing to `wrapGAppsHook4`

poseidon-rises:
I'm no expert on the subject by any stretch of the imagination, but from reading around it seems like it has something to do with the gnome file picker breaking down when called from a Qt app, causing the Qt app to crash. 

The hook does some magic to fix it, not sure on the details but there's really not much we can do about it. [This](https://github.com/NixOS/nixpkgs/pull/271037) would patch the file picker to not do that but it hasn't been touched for a few months now. 

Either way, since we don't use GTK ourselves, we should probably use `wrapGAppsNoGuiHook`, which will apply the fix without pulling in GTK as a dependency.

xarvex:
Thanks very much for the fix @Ambossmann, I have been busy here with university, so I haven't had my eyes on this as much as I should have. Very interesting, I actually hadn't realized this *was* an alias.

> just a thought: this change comes from [nixpkgs!307077](https://github.com/NixOS/nixpkgs/pull/307077) which also mentions that `wrapGAppsHook3` is for GTK3 while `wrapGAppsHook4` is for GTK4. I have no idea why we even need to do anything regarding GTK since we don't use GTK, but whatever Gnome thing this is about might in the future require us changing to `wrapGAppsHook4`

This is more about how Qt hooks in to be platform “native” on different systems, in this case the hook takes care of the `gtk3` platform theme (which can be forced with `QT_QPA_PLATFORMTHEME=gtk3`). There is not a GTK4 equivalent at the moment.

> Either way, since we don't use GTK ourselves, we should probably use wrapGAppsNoGuiHook, which will apply the fix without pulling in GTK as a dependency.

Unfortunately, this does not suffice for dependency requirements, and the application will crash.

Issue #1195, opened 2025-11-10
[Bug]: Crash when adding file to library with Japanese filename


SkeleyM:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Main branch

### Operating System & Version

Windows 10 22H2

### Description

When adding a file/creating a new library  that includes a file with a japanese name the program freezes and an error appears in the console.

### Expected Behavior

The file should add just fine

### Steps to Reproduce

1.  Create a new library
2.  Include a file in the library called "こんにちは.png" (the offending name in my case)
3.  Refresh directories

### Logs

2025-11-10 21:33:32 [info     ] [Ignore] No updates to the .ts_ignore detected last_mtime=1762809066.321023 library=WindowsPath('C:/Users/user/Pictures') new_mtime=1762809066.321023
2025-11-10 21:33:32 [info     ] [Refresh: Using ripgrep for scanning]
Exception in thread Thread-46 (_readerthread):
Traceback (most recent call last):
  File "c:\python312\Lib\threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "c:\python312\Lib\threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
  File "c:\python312\Lib\subprocess.py", line 1599, in _readerthread
    buffer.append(fh.read())
                  ^^^^^^^^^
  File "c:\python312\Lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 1: character maps to <undefined>
Error calling Python override of QRunnable::run(): Traceback (most recent call last):
  File "C:\Users\user\Documents\TagStudio\src\tagstudio\qt\utils\custom_runnable.py", line 18, in run
    self.function()
  File "C:\Users\user\Documents\TagStudio\src\tagstudio\qt\utils\function_iterator.py", line 21, in run
    for i in self.iterable():
             ^^^^^^^^^^^^^^^
[...]

PR #1199, opened 2025-11-11
fix: call ripgrep with explicit utf-8 encoding.

Sola-ris:
### Summary

Call ripgrep with explicit `UTF-8` encoding to prevent issues with multi-byte characters.

When calling `subprocess.Popen`, (which is where `silent_subprocess::silent_run` eventually ends up) with `text=True`[ without specifying the encoding](https://github.com/python/cpython/blob/3.12/Lib/subprocess.py#L876-L877), python will use the encoding returned by `locale::getencoding`.

On Linux this returns `UTF-8` while on Windows it will return `cp1252` which causes issues with multi-byte characters like em dashes or Japanese characters.
linux
win

Fixes https://github.com/TagStudioDev/TagStudio/issues/1195.

#### Before
before

I've omitted `こんにちは.txt` from the screenshot since it hangs the program with the error reported in https://github.com/TagStudioDev/TagStudio/issues/1195

#### After
after

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -40,6 +40,7 @@ jobs:
             libxcb-xinerama0 \
             libxkbcommon-x11-0 \
             libyaml-dev \
+            ripgrep \
             x11-utils
 
       - name: Execute pytest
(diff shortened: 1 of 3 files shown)


CyanVoxel:
Sorry for the delay on pulling this, and thank you so much for your work on this fix!

PR #1217, opened 2025-11-18
refactor: thumbnail renderers

TrigamDev:
### Summary
- The renderers have been split out from `src/tagstudio/qt/previews/renderer.py`, moving them into their own files in `src/tagstudio/qt/previews/renderers/*_renderer.py`.
- Renderers, instead of accepting a long list of variables, accept a `RendererContext` object, which provides the file path, file extension, image size, pixel ratio, and whether the image is being used for a grid thumbnail.
- The `.tar` and `.7zip` file wrappers have been split out from `src/tagstudio/qt/previews/renderer.py`, moving them into their own files in `src/tagstudio/qt/helpers/file_wrappers/archive/*_file.py`.
  - Wrappers have also been for `.zip` and `.rar` files. This is because:
    - Checking against the archive file type in `renderer.py` would result in a circular dependency, plus checking against `type[ArchiveFile]` is nicer.
    - Some tweaks needed to be made to handle a weird case where zip files were including the name of the file in their file paths. Using a wrapper makes this much easier to handle cleanly and include the fix for the other archive files.
    - Some slight naming tweaks were made to better match the usual formatting. Since the wrappers were made to emulate the API of zip files, a wrapper needed to be added around zip files to make them consistent with the tweaked naming.
    - A helper method was added for checking if a file is contained within the archive file. This isn't necessary by any means, it's just nice to have.
- `.exr` files are now handled better.
  - Not entirely sure if it's just from the files I tested with, but thumbnails weren't able to be generated for them, and their stats (width/height) couldn't be loaded.
  - They now have a new thumbnail renderer using `OpenEXR`. `OpenEXR` is also now used to get their width and height.
- Fixed iWork files (sometimes) failing to generate a thumbnail.
  - Again, not sure if it's just the files I was testing with, but they weren't able to generate their thumbnails due to the issue mentioned above with `.zip` files including their name in their file paths.
- Various other tweaks, such as more descriptive variable naming, added type hints, and preferring functions over static methods.

### Notes
I wasn't sure what to do with the following files, so I just left them where they were.
- `src/tagstudio/qt/previews/vendored/ffmpeg.py`
- `src/tagstudio/qt/previews/vendored/pydub/audio_segment.py`
- `src/tagstudio/qt/previews/vendored/pydub/utils.py`

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality


CyanVoxel:
Closing in favor of #1462

Issue #1215, opened 2025-11-18
[Feature Request]: Selection history


louisl4696:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

I'm not so sure how to describe this feature in technical terms so if you'll allow, I made a bit of a user story below.

Feature : possibility to undo / redo changes to the selection in the thumbnail grid.

Picture this :
You're scrolling through your images, meticulously selecting all those containing cats, creating a massive selection one Ctrl+Click at a time. When you're done, you'll add your shiny new "cat" tag to them all in one fell sweep.
But your pinky finger weakens, and you click on a file without holding down Ctrl. The dreaded words appear : "One item selected". In an instant, the past 5 minutes of intensive thumbnail scrutinizing are lost. Your carefully crafted selection vanished. If only you could bring it back !

### Solution

Hitting a keyboard should undo the most recent change to the selection, every successive shortcut input going back one selection state further in the history.

Another shortcut could be used to undo the undoing, moving forward in the history.

refreshing the grid (for instance by performing a search) should clear the history.

TrigamDev:
I think this could be expanded to instead be more of an "Action history", where selection changes are just one of the actions it tracks, along with things like tag and field changes.

louisl4696:
I mainly created this issue because I got a bit carried away while working on #226 and implemented it. The history is managed by a decorator added to all functions that change the selection, which includes those I developed for arrow key navigation. so it would be really annoying to separate them into two PRs (or maybe my git skill is just too limited for this kind of stuff)

louisl4696:
> I think this could be expanded to instead be more of an "Action history", where selection changes are just one of the actions it tracks, along with things like tag and field changes.

Personally I think It's probably best to keep the "real" action history separate from the "selection" history, because the selection can change a lot (think arrow key browsing) creating loads of noise for the "real" actions in the history.

TrigamDev:
True, I'm just used to some programs including selection changes into the same history as actions

louisl4696:
I can think of some myself too. But also some that don't ! For instance Windows file explorer does not save selection states in the history.

[...]

PR #1220, opened 2025-11-19
feat(thumbnail grid): arrow key navigation & selection history 

### LINKED ISSUE #226: Navigation with arrow keys
suggestion to add ability to navigate through previews with with the arrow keys

louisl4696:
- Added arrow key navigation for thumbnail grid #226
- Implemented selection history with undo/redo #1215

#### Arrow key navigation
Arrow keys events in MainWindow are intercepted by an eventFilter and trigger callbacks in ThumbGridLayout. Shift key events are also intercepted but do not block the event from continuing down Qt's event pipeline. Arrow key navigation mimics the behaviour of windows file explorer as outlined in #226, including multi-selection.
closes #226 

#### Selection history
The last 30 (number chosen arbitrarily) selections are stored in a deque. Hitting R pops the most recent selection from the deque, and pushes the current selection to a second deque for redo-ing. Hitting Shift+R pops from the redo deque. It's very satisfying to use (I may be biased)
This feature could later be included into a higher level history, see #1215 for the full discussion !
Selection state changes are saved in the deques via a decorator for methods that modify the selection.
Because that includes the methods used for arrow key navigation, I wasn't able to cleanly separate these two features into two PRs... Sorry 😅

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/qt/thumb_grid_layout.py
+++ b/src/tagstudio/qt/thumb_grid_layout.py
@@ -2,6 +2,7 @@ import math
 import time
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, override
+from collections import deque
 
 from PySide6.QtCore import QPoint, QRect, QSize
 from PySide6.QtGui import QPixmap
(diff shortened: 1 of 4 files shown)


TrigamDev:
I've noticed that occasionally, the scrollbar will suddenly jump to a different position (often the bottom) seemingly randomly while navigating with arrow keys.

TrigamDev:
Some other issues I've noticed:
- Occasionally, navigating with arrow keys will skip over entries
- Navigating entries will override navigating search bar autocomplete suggestions

TrigamDev:
> * Occasionally, navigating with arrow keys will skip over entries

This seems to start happening after interacting with an entry's tags.

louisl4696:
I'm back, but busier than ever :')

Yeah, I noticed the behaviours you mentioned too after using it for some time.
I way I implemented it, pressing the arrow key only makes changes the selection.

Concerning the autocomplete, that's an oversight on my end... I actually didn't know that feature existed x)

For the scrolling, my best guess is that it has something to do with selecting an entry that's outside of the visible range, something that isn't possible to do with just the mouse...

It seems to me that modifying tags is (sometimes) causing the inputs to be counted twice, but why that is I don't know yet

PR #1224, opened 2025-11-23
fix: 'Add Tag to Selected' action fails

TrigamDev:
### Summary
Fixes a bug where using the 'Edit > Add Tag to Selected' action would fail and not add the chosen tag.

Was brought up in https://discord.com/channels/1229183630228848661/1229309667528806420/1442232941509083197.

(Recreated from #1223 as it was closed when I renamed the branch)

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/src/tagstudio/qt/mixed/field_containers.py
+++ b/src/tagstudio/qt/mixed/field_containers.py
@@ -10,7 +10,7 @@ from datetime import datetime as dt
 from warnings import catch_warnings
 
 import structlog
-from PySide6.QtCore import Qt, Signal
+from PySide6.QtCore import Qt
 from PySide6.QtGui import QGuiApplication
 from PySide6.QtWidgets import (
     QFrame,
@@ -22,7 +22,6 @@ from PySide6.QtWidgets import (
     QWidget,
 )
 
-from tagstudio.core.constants import TAG_ARCHIVED, TAG_FAVORITE
 from tagstudio.core.enums import Theme
 from tagstudio.core.library.alchemy.enums import FieldTypeEnum
 from tagstudio.core.library.alchemy.fields import (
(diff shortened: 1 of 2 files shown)


CyanVoxel:
Thank you for the quick fix! Though I believe this uncovered a related issue where setting the tags from this action causes the selection in the preview panel to reset. Good news is that it looks straightforward enough to fix, with the cause seeming to be the lambda variable assignment on line 371 of `ts_qt.py`. Here I've removed the variable `s` that was being set with the initial empty value and am instead using the the direct `self.selected` value which uses the correct selection value when the lambda is called:
```py
self.add_tag_modal.tsp.tag_chosen.connect(
            lambda t: (
                self.add_tags_to_selected_callback([t]),
                self.main_window.preview_panel.set_selection(self.selected),
            )
        )
```
I also noticed a similar use of `s=self.selected` in the lambda above this one, though haven't looked into the repercussions of that one...

TrigamDev:
I've also noticed while testing just now that, when adding tags through the 'Add Tag' button, the badges associated with that tag will update as expected. But when doing the same with the 'Add Tag to Selected' action, the badges will only update upon viewing the entry in the preview panel.

TrigamDev:
> I've also noticed while testing just now that, when adding tags through the 'Add Tag' button, the badges associated with that tag will update as expected. But when doing the same with the 'Add Tag to Selected' action, the badges will only update upon viewing the entry in the preview panel.

Went ahead and fixed this as well

PR #1232, opened 2025-11-26
ci: add pyright workflow

Computerdores:
### Summary
Adds a pyright workflow using reviewdog (see #1061).

While this is intended to be merged once the code base is ready, in the mean-time I whipped this up to give us an idea were we are in the process of getting there. 

PS: I didn't remove the mypy workflow, but it might make sense to put that into this PR as well.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [ ] Basic functionality


Computerdores:
~~not quite sure why atm the check passes on the PR (incorrect) but not on the commit (correct), but I suspect that review dog is filtering the errors on the PR because they don't stem from anything modified here, in that case this wouldn't be a problem (and indeed the desired behaviour) in normal use~~

Computerdores:
There are now PRs open for all remaining Pyright errors, once those are merged this PR should be mergable, save for any issues that show up in review.

pyproject.toml
```diff
@@ -99,13 +100,11 @@ ignore = [
...
 ]
 include = ["src/tagstudio", "tests"]
-reportAny = false
-reportIgnoreCommentWithoutRule = false
+# reference for the settings here: https://github.com/microsoft/pyright/blob/main/docs/configuration.md
 reportImplicitStringConcatenation = false
 reportMissingTypeArgument = false
 reportMissingTypeStubs = false
 # reportOptionalMemberAccess = false
-reportUnannotatedClassAttribute = false
```
CyanVoxel, line 109: These settings are specific to [basedpyright](https://docs.basedpyright.com/latest/), which while that's not the workflow we're going with at least gives some some extra depth to anyone who chooses to use it locally (at the very least me)

pyproject.toml
```diff
@@ -45,6 +45,7 @@ dependencies = [
 dev = ["tagstudio[mkdocs,mypy,pre-commit,pyinstaller,pytest,ruff]"]
 mkdocs = ["mkdocs-material[imaging]>=9.6.14", "mkdocs-redirects~=1.2"]
 mypy = ["mypy==1.15.0", "mypy-extensions==1.*", "types-ujson~=5.10"]
+pyright = ["pyright~=1.1.405"]
```
CyanVoxel: I believe this can be safely bumped to 1.1.409
```suggestion
pyright = ["pyright~=1.1.409"]
```

pyproject.toml
Computerdores: iirc normal pyright throws a warning / error when invalid settings are present

CyanVoxel: I get output in the log but as far as I know it's not a proper warning or error and pyright just moves on:
```
Config contains unrecognized setting "reportAny".
Config contains unrecognized setting "reportIgnoreCommentWithoutRule".
Config contains unrecognized setting "reportUnannotatedClassAttribute".
```

Computerdores: I have added them back, but we should check prior to merging this whether the workflow passes after merging in #1368 (via main)

Computerdores:
All comments are resolved, main is merged in and pyright passes on pull_request and [push](https://github.com/Computerdores/TagStudio/tree/ci/pyright)

CyanVoxel (approved):

Issue #808, opened 2025-02-18
[Feature Request]: Tag Management for multiple files [improvement]


MonsterSe7en:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

# Tag Management for Multiple Files

Selecting multiple files currently results in a blank Tag area.
Is this intended behavior? It would be helpful if selecting multiple files could display both **common tags** (shared across all files) and **partial tags** (present on some files).

## Screenshots for clarity:
---
## Single file selection shows tags as expected:
![Image](https://github.com/user-attachments/assets/a75cf405-0d39-41ff-aabf-e396f0eaddd9)

## Multiple file selection shows blank Tag field:
![Image](https://github.com/user-attachments/assets/dea94e42-160c-4d9d-841d-eb18159d2379)

# Tags Window

The ability to add tags to multiple files is brilliant, could we also implement bulk tag removal? Perhaps using middle-click in the Add Tags window to remove tags?

(I did find the Manage Tags window Ctrl + M, but that ended up deleting the Tag from the database 😭)

Currently there's no visual confirmation when tags are added to multiple files in the Add Tags Window (Checking the background window Tags field works) 
Also note The Tags field in the background window only shows tag updates for single file selections, not when multiple files are selected.

## Real-time tag updates
Note how the Tag field updates in real-time when adding tags (visible in the background window's Tag Field, right side):
![Image](https://github.com/user-attachments/assets/7b54e68a-6750-4e3b-ba77-b1bdf5624526)

## Is it possible to display a checkbox next to the Tags in the Tags Window and the main window (Bottom Right corner) :

A check ☑️ if **all** the selected files have the Tag

A Dash ➖ Patrial Tag When **some** selected files have the selected Tag

An Empty 🟪 unchecked box when **none** of the selected files have the Tag added

I know the plus icon was just removed😶 (Thanks for removing the plus icon that was Awesome!!!) 
I prefer the latest implementation where the whole box/Tag can be clicked instead of a smaller plus box, I'd like to retain that behavior and include a tick box 

This would give users both:

- The convenience of clicking anywhere on the tag

- Clear visual confirmation through the checkbox state

Thank you for taking the time to read this.
Cheers!

CyanVoxel:
[...]

PR #1238, opened 2025-11-29
feat: extend preview panel multi-selection with shared tag editing and update tests

JasonReb0rn:
### Summary
- Show shared tags for multiple selected entries and allow bulk tag removal
- Enable preview panel actions for multi-selection and adjust related tests
- Update preview file attributes text to indicate multi-selection and shared tags

### Tasks Completed

-   Platforms Tested:
    -   [X] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [X] Basic functionality


TrigamDev:
When you have an entry selected, select a second, large entry that takes a while to load, and then selecting a third entry before the second loads, the tags and fields will be stuck in a state where they don't update, even when selecting other entries. They will be the tags and fields of the very first entry.

TrigamDev:
Also, I think it would be nice if tags and fields that aren't shared between all selected entries are still displayed, but grayed out or low opacity

JasonReb0rn:
> When you have an entry selected, select a second, large entry that takes a while to load, and then selecting a third entry before the second loads, the tags and fields will be stuck in a state where they don't update, even when selecting other entries. They will be the tags and fields of the very first entry.

Thank you for catching that. I believe I've addressed the race condition via timestamp-based cutoff.

Regarding the display of non-shared tags, I added a separate "mixed" tag section, which still allows for bulk tag removal. I think this approach is cleaner than graying tags out, but I'm happy to adjust if you prefer a different way forward.

TrigamDev:
> I'm happy to adjust if you prefer a different way forward.

I should clarify that I'm not representative of Cyan or the project or anything, I'm just speaking as to what I'd personally prefer.

> I believe I've addressed the race condition via timestamp-based cutoff.

Sadly I'm still experiencing the issue.

> Regarding the display of non-shared tags, I added a separate "mixed" tag section, which still allows for bulk tag removal. I think this approach is cleaner than graying tags out

Instead of having them display as a category, it might be nicer to instead add a second field box for all the mixed fields. It'd allow the mixed tags to display with categories as well as allow mixed fields to be displayed.

It would also be better to tweak the "Showing tags shared by all entries" message to instead say "tags and fields" or something similar.

src/tagstudio/qt/mixed/field_containers.py
```diff
@@ -422,29 +502,36 @@ def write_tag_container(
...
+                self.driver,
+            )
+            container.set_inner_widget(inner_widget)
+
+        # For mixed tag containers, mark the widget so it can gray out all tags.
+        if is_mixed:
+            inner_widget.set_mixed_only(True)
+            container.set_title(Translations["preview.partial_tags"])
+        else:
+            inner_widget.set_mixed_only(False)
```
Computerdores: this seems like it would cause the wrong title to be shown if inner_widget was previously a non-mixed one, since it doesn't set the title in that case

src/tagstudio/qt/mixed/field_containers.py
```diff
@@ -111,8 +111,87 @@ def update_from_entry(self, entry_id: int, update_badges: bool = True):
         self.cached_entries = [entry]
         self.update_granular(entry.tags, entry.fields, update_badges)
 
+    def update_from_entries(self, entry_ids: list[int], update_badges: bool = True):
```
Computerdores: In many other cases in the codebase functionality that can be done for either one or several items is implemented in a single method that checks whether a list or a single item was provided (or if the list only has one item). Doing the same for `update_from_entries` and `update_from_entry` would simplify the call in several places

src/tagstudio/qt/controllers/tag_box_controller.py
```diff
@@ -33,6 +34,38 @@ def __init__(self, title: str, driver: "QtDriver"):
...
+            for tag_id, entries in tag_entries.items():
+                if set(entries) >= required:
+                    shared_tag_ids.add(tag_id)
+
+        super().set_tags(tags_)
+
+        # Gray out tags that are not shared across all selected entries.
+        from tagstudio.qt.mixed.tag_widget import TagWidget  # local import to avoid cycles
+
+        layout = getattr(self, "_TagBoxWidgetView__root_layout", None)
```
Computerdores: this is fragile in regards to refactoring and generally regarded as bad practice since it violates separation of concerns.

(When a private variable *really* needs to accessed from outside of the declaring class it should be marked with a single underscode in the beginning instead, altough it should be carefully considered whether it makes sense in each specific case since it can lead to spaghetti code where every thing is interdependent and it this becomes hard to make changes without unforeseen consequences)

src/tagstudio/qt/controllers/tag_box_controller.py
```diff
@@ -33,6 +34,38 @@ def __init__(self, title: str, driver: "QtDriver"):
...
+        if layout is not None:
+            for i in range(layout.count()):
+                item = layout.itemAt(i)
+                widget = item.widget()
+                if isinstance(widget, TagWidget) and widget.tag:
+                    if self.__mixed_only or widget.tag.id not in shared_tag_ids:
+                        widget.setEnabled(False)
```
Computerdores: this violates the core idea of MVC[^1]. UI element related stuff should be handled on the view side not on the controller side. I would suggest moving this to `TagBoxWidgetView.set_tags` since that is called right before anyway (would also fix the issue from my previous comment, about the access to the private field)

[^1]: See the [Style Guide](https://docs.tagstud.io/style/#qt) for an explanation

Computerdores (commented):
Generally this is looking good code wise (haven't had time to test yet), but there is a couple of things that could be improved.

(I tried to be verbose in my explanations, but if anything is unclear let me know)

TrigamDev:
Adding a new field won't trigger the fields list to update. In order to see newly added fields, changing the selection is required.

TrigamDev:
This closes #808

JasonReb0rn:
I greatly appreciate the feedback from both @TrigamDev and @Computerdores , and I apologize for the lack of response. This was originally a quick fix for my own immediate needs, but I’ll work on cleaning it up over the weekend.

TrigamDev:
> I apologize for the lack of response.

You're fine, you don't need to worry
And I'm looking forward to seeing this being finished, I've been using it and it's proven really useful.

JasonReb0rn:
Ok, I've spent some more time on this and made a broader pass over the bulk edit preview behavior.

The changes here mainly focus on how the tags and fields panel updates during multi-selection. Shared tags and fields are now separated more clearly from ones that only exist on some of the selected entries, the panel refresh behavior after bulk edits is more consistent, and the tag rendering logic was cleaned up a bit based on the review feedback.

Issue #1254, opened 2025-12-12
[Bug]: URL Regex is broken (2 character fix)


pwall2222:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Alpha 9.5.6

### Operating System & Version

NixOS 25.05

### Description

The regex for URLs missbehaves, and causes `-` (dashes) in the path segment of the url to not be recognised.

This is because it takes `#-*` to be a range instead of a collection of [`#`, `-`, `*`] to fix this we only have to add a `\` to put it like so `#\-*`
https://github.com/TagStudioDev/TagStudio/blob/88d0b47a86821ccfadba653f30a515abce5b24b0/src/tagstudio/qt/mixed/text_field.py#L38
```regex
(http|ftp|https):\/\/([\w_-]+(?:(?:\.[\w_-]+)+))([\w.,@?^=%&:\/~+#\-*]*[\w@?^=%&\/~+#\-*])
```

### Expected Behavior

For the URLs containing `-` in the path part to be recognised and clickable as a whole.

### Steps to Reproduce

1. Add URL Field
2. Add URL with `-` in the path part for example `https://docs.tagstud.io/preview-support/`
3. Only `https://docs.tagstud.io/preview` get's highlighted

CyanVoxel:
Fixed in #1255

PR #1255, opened 2025-12-12
fix: escape dash in URL regex

TrigamDev:
### Summary
Escapes a dash in the URL regex that was preventing URLs with dashes from being correctly matched.

Closes and fix provided by #1254.

> Although this is also already fixed by #1168, this is a quick patch until then.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality


CyanVoxel (approved):
Looks good to me, thank you for the (insanely) quick fix!

PR #1265, opened 2025-12-16
fix(nix): add requests and semver to nix package

Ambossmann:
### Summary
The addition of the update notification in #1166 added `requests` and `semver` to the dependencies. They were however not added to the nix package, breaking the build.

This adds both of them as dependencies to the nix package. It also add `requests` to `pythonRelaxDeps` as `requests` is already on 2.32.5 in nixpkgs but the `pyproject.toml` specifies `requests~=2.31.0`.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/nix/package/default.nix
+++ b/nix/package/default.nix
@@ -80,6 +80,8 @@ python3Packages.buildPythonApplication {
     "py7zr"
     "pyside6"
     "rarfile"
+    "requests"
+    "semver"
     "structlog"
     "typing-extensions"
   ];
@@ -103,6 +105,8 @@ python3Packages.buildPythonApplication {
       pyside6
       rarfile
       rawpy
+      requests
+      semver
       send2trash
       sqlalchemy
       srctools


xarvex:
Thank you for this! I have been busy with my last semester at university so I didn't catch this.

One small note: can you specify `semver` in `pythonRelaxDeps` too? Currently the version is matching, but given the dependency isn't critical at the moment, any future version mismatch should not be a build failure.

I don't have access to my computer right now, so I'll be pulling later today.

M3L6H:
I can verify I ran into the same problem & @Ambossmann's fix worked for me. Also running a Linux x86 system.

xarvex:
Trip took longer than expected, but all checks out to me here, thank you for the fix!

PR #1290, opened 2026-01-15
fix: UI uses translated keys if available

### LINKED ISSUE #1316: [Bug]: Some UI elements only use English strings, even when the translated keys are available
### Checklist

### TagStudio Version

TagStudio main branch

Windows 11

### Description

The following UI elements are not translated in the application even though they're translated on weblate & in the .json file (I'm using Japanese for the examples; I also noted the keys they're supposed to be using):

"Remove Field" title bar: `library.field.remove`
"Remove" button: `generic.remove_alt` & `generic.remove`
Image
Thumbnail size options: `home.thumbnail_size` keys
Image
"All Tags" option: `tag.all_tags`
Image
Filepath visibility options: `settings.filepath.option` keys
Tag click action options: `settings.tag_click_action` keys (except `settings.tag_click_action.label`)
Theme options: `settings.theme` keys (except `settings.theme.label`)
Splash screen options: `settings.splash.option` keys
https://github.com/user-attachments/assets/2239456a-30de-436d-b2e2-850421cb12e4
In the "Edit Color" dialog,
"Color" placeholder: `color.placeholder`
Image
Folders to tags macro description: `folders_to_tags.description`
Image
In the "Delete file(s)" dialogs,
"Yes" button: `generic.yes`
"No" button: `generic.no`
Also, the dialogs try to use the `trash.title` keys instead of the `trash.dialog.title` keys
Image
Image
In the "Delete tag" dialog,
"Cancel" button: `generic.cancel` & `generic.cancel_alt`
"OK" doesn't have a key
Image

Supersedes #1289 

### Expected Behavior

These UI elements should be translated when a translation exists

### Steps to Reproduce

1. Change the language to something that isn't English
2. Restart TagStudio

### Logs

_No response_

Math-Bee:
### Summary
The settings dropdowns, thumbnail size options, and "all tags" tag search setting now get their translated text during `.addItem`:
The language is changed within `ts_qt`'s `__init__()` function, but class definitions are executed when the file's imported; This means that if a translation is defined in a class definition (or outside any class/function), it uses the default language (english) rather than the correct language.
The translation variable definitions *could* have been moved inside the `__init__()` function of their class, but because they're not used by anything other than their comboboxes, it's simpler & more space efficient to just directly add them to their combobox

The deleting tags dialog is now a warning to match the deleting fields dialog, & uses the `Discard` `StandardButton` instead of the `Ok` one. The deleting fields dialog now uses `StandardButton`s instead of `ButtonRole`s for simplicity:
The `exec()` function doesn't return the button's `ButtonRole`s' value, it returns a value that depends on when you add the button; This wasn't super clear, especially since the statement that checks what button was pressed compares the result with `ButtonRole.ActionRole.value`, which only worked because it equals `3`, and the correct button was the 2nd button added. The sort of button-checking implied by the code only actually works if you use `StandardButton`s, which is why it uses them now.

The color placeholder, the description of the folders to tags macro, and the deleting files dialog just needed to swap the explicit text for translated text.

Closes #1316.

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality


Math-Bee:
Drafting this to fix #1316

PR #1293, opened 2026-01-17
Fix faulty progress bars

### LINKED ISSUE #1216: [Bug]: Off-By-N in relink progress bar
### Checklist

### TagStudio Version

Alpha 9.5.6

Win11

### Description

Relinking entities shows it has already completed relinking `n` entries successfully, but that it is in the process of attempting to relink the `n-1`th item. 

Image

### Expected Behavior

The counter of what item is being attempted should always be the same or higher than the number of matches. 
Realistically, it should be updating the window when it attempts to link a new item, then again when it succeeds/fails, so it should be `n+1`.

I believe the problem lies in `src\tagstudio\qt\mixed\relink_entries_modal.py::RelinkUnlinkedEntries,repair_entries`, where `displayed_text` is passed the index, `x` (which is from `src\tagstudio\core\library\alchemy\registries\unlinked_registry.py::UnlinkedRegistry.fix_unlinked_entries`)

Changing this to `n+1` on the display would work as a fix, however it might be worth updating the window twice, as I mentioned before.

### Steps to Reproduce

1. Start relinking entries where all entries will be relinked successfully 
2. See issue

### Logs

_No response_

ludvig-sandh:
### Summary

Fixes #1216 by addressing a more deeply rooted problem affecting all progress bars that are counting something (in total 6 in TagStudio). Also fixed a minor text bug where something like "Completed: 4/{unknown key}" because the formatter expected the parameter "total", not "count".

See my comment in #1216 for all details, but essentially this fix introduces a standard for how to use counting progress bars, and fixes some surrounding bugs regarding when to update the progress bar UI and with what counts. This turns all the progress bars from looking like this...

---

 (during processing of file 1: empty but should show 0%)
before_1_of_3

---

(during processing of file 2: should say 1 file completed but says 2, and should be 33% completed)
before_2_of_3

---

(during processing of file 3 (last): empty but should show 2 files completed and 66%)
before_3_of_3

... to this:

---

(during processing of file 1)
after_1_of_3

---

(during processing of file 2)

after_2_of_3

---

(during processing of file 3 (last))
after_3_of_3

The fix, in words, was to remove an off-by-one error where 1 is always added to the progress => corrects the percentage counts, as well as fixes the error where the last step in the progress window was always empty.
Secondly, I documented that the progress count handed to the UI element should always be the current number of completed elements, nothing else. This removed the need for off-by-one cases scattered around the project => logic simplified.
Thirdly, the progress UIs are updated in the beginning of the loop iterations instead of at the end, to avoid that first blank view when no progress has been made yet.

### Tasks Completed

-   Platforms Tested:
    -   [X] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [X] Basic functionality


CyanVoxel (approved):

PR #1294, opened 2026-01-18
minor refactor: cleanup parameters of open_library and open_sqlite_library

Computerdores:
### Summary

A minor refactor of the `Library.open_library` method and the parameters of the `Library.open_sqlite_library` method.
No functional changes were made; a pure refactor.

Note: Ignore the invisible unicode char in the head ref; windows was doing windows things.

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality
    -   [ ] PyInstaller executable

PR #1295, opened 2026-01-18
refactor: sql migrations

Computerdores:
### Summary

This PR refactors the SQL migrations for old DB versions. Previously these migrations happened out of order and were split between schema changes and data updates. This was done because some SQLAlchemy read operations would otherwise fail due to an incompatible schema[^1]. This refactor rearranges these migrations into chronological order. This was done, because
- the code was very entangled thus hard to understand und verify,
- the chronological step-wise way is the typical way of migrating DBs,
- and because this way, new migrations can be added easily and with confidence that none of the previous steps will interfere or be interfered with

Note for reviewers: The changes were intentionally split into small commits that should be easily verifiable individually, because they make small changes that should change nothing about the correctness of the migration steps.

Note for reviewers 2: This is based on #1294.

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality

[^1]: this was fixed in f28adfd6f9e718aca0b3e69778153a56e856194b

### DIFF
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -536,8 +536,7 @@ class Library:
                     self.save_library_backup_to_disk()
                     self.library_dir = None
 
-                # NOTE: Depending on the data, some data and schema changes need to be applied in
-                # different orders. This chain of methods can likely be cleaned up and/or moved.
+                # migrate DB step by step from one version to the next
                 if loaded_db_version < 7:
                     # changes: value_type, tags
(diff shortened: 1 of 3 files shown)


CyanVoxel (approved):
I believe this all looks good, thank you so much for your work on this!

PR #1297, opened 2026-01-21
fix: empty ORLists now return false

### LINKED ISSUE #1287: [Bug]: Search OR returns everything when no tag exists
### Checklist

### TagStudio Version

Alpha 9.5.6

Windows 11

### Description

When searching with an OR operator, if none of the tags used exist in the library, the or statement gets ignored.

### Expected Behavior

Because no entry has the requested tags (because they don't exist), i would asume that nothing get's returned.

### Steps to Reproduce

1. Open a library
2. Enter an OR search query, where every tag is not in the library (example: "tag:a or tag:b", where tag a and b do not exist)
3. Everything get's displayed

### Logs

```
2026-01-06 17:39:36 [info     ] SQL Expression Builder finished (0 seconds)
2026-01-06 17:39:36 [info     ] searching library              filter=BrowsingState(page_index=0, sorting_mode=<SortingModeEnum.DATE_ADDED: 'file.date_added'>, ascending=False, random_seed=0, show_hidden_entries=False, query='tag:a or tag:b') query_full='SELECT DISTINCT entries.id \nFROM entries \nWHERE NOT (EXISTS (SELECT 1 \nFROM tags, tag_entries \nWHERE entries.id = tag_entries.entry_id AND tags.id = tag_entries.tag_id AND tags.is_hidden)) ORDER BY entries.id DESC'
2026-01-06 17:39:36 [info     ] SQL Execution finished (0.1 seconds)
2026-01-06 17:39:36 [info     ] items to render                count=2047
```

TheBobBobs:
### Summary
If an `ORList` has only invalid tags return `False`.

The default query is an empty `ORList` so return `True` if the `ORList` contained no elements before seperating tags.

fixes #1287 
### Tasks Completed
-   Platforms Tested:
    -   [X] Linux x86
-   Tested For:
    -   [X] Basic functionality

### DIFF
--- a/src/tagstudio/core/library/alchemy/visitors.py
+++ b/src/tagstudio/core/library/alchemy/visitors.py
@@ -6,7 +6,7 @@ import re
 from typing import TYPE_CHECKING, override
 
 import structlog
-from sqlalchemy import ColumnElement, and_, distinct, func, or_, select
+from sqlalchemy import ColumnElement, and_, distinct, false, func, or_, select, true
 from sqlalchemy.orm import Session
 from sqlalchemy.sql.operators import ilike_op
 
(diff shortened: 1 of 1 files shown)


Computerdores:
While this might[^1] fix the specific symptom reported in #1287 this doesn't address the root cause of this issue, which is likely that `__separate_tags` incorrectly handles list of only incorrect tag names. And since `__separate_tags` is used elsewhere too from what I remember, it would be better to fix that method.

[^1]: I am not doubting that it does, I just didn't test it myself.

PR #1324, opened 2026-03-01
test: add numeric field tests REQ-01& REQ-01d to feat branch

sumeyayasir:
Moving tests  to the correct feature branch. Refs #1


CyanVoxel:
Looks like this was meant for your own fork, similar to #1327

PR #1326, opened 2026-03-01
fix(nix): relax send2trash dependency

Ambossmann:
### Summary

send2trash has recently increased it's major version to 2 and 2.1.0 has now landed in nixpkgs. This breaks the build as it doesn't fulfil the ~=1.8 constraint in pyproject.toml. From the [changelog](https://github.com/arsenetar/send2trash/blob/master/CHANGES.rst) the only breaking change should be dropped python 2 support, which doesn't affect us.

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality

### DIFF
--- a/nix/package/default.nix
+++ b/nix/package/default.nix
@@ -82,6 +82,7 @@ python3Packages.buildPythonApplication {
     "rarfile"
     "requests"
     "semver"
+    "send2trash"
     "structlog"
     "typing-extensions"
   ];


xarvex:
Whoops, apologies for missing this, but I have addressed this and `rawpy` (which also updated recently) in https://github.com/TagStudioDev/TagStudio/commit/c38fbb034da0ee9e8c745a73542cfd92d74ccda0. Thank you regardless!

Issue #287, opened 2024-06-14
[Bug]: Dark Mode won't activate on Fedora 40 


Takanu:
### Checklist

- [X] I am using an up-to-date version.
- [X] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/doc/index.md).
- [X] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

Alpha 0.9.3

### Operating System & Version

Fedora 40 Workstation (Linux)

### Description

Dark Mode doesn't activate when the OS is in Dark Mode.  

I tried finding various ways to "force enable" the dark mode through available QT options and environment variables that can be modified on Linux to no effect.

### Expected Behavior

I expected either:
- The app to recognize the OS theme mode and change itself to match
- The app to offer manual Dark Mode / Light Mode controls

### Steps to Reproduce

No steps required, opening the app as is provided in Fedora 40 Workstation creates the issue.

### Logs

While the logs don't appear to provide anything useful on inspection ill post the opening log just in case:

```
[QT DRIVER] Config File not specified, defaulting to /home/takanu/.config/TagStudio/TagStudio.ini
qt.svg: Cannot open file './tagstudio/resources/pause.svg', because: No such file or directory
qt.svg: Cannot open file './tagstudio/resources/volume_muted.svg', because: No such file or directory
Calling NavForward with Content:False, Index:0, PageCount:0
[ENTRY PANEL] UPDATE WIDGETS ([])
```

Please let me know if there's anything else I can try.

ooTruffle:
On kde neon it also does not use dark mode for some objects, Causeing a Unique look

![image](https://github.com/user-attachments/assets/a70f65fc-7b4a-4ce9-aa32-c302289cfc97)

RokeJulianLockhart:
@Takanu, you're using GNOME 40, right?

Takanu:
> > [#287 (comment)](https://github.com/TagStudioDev/TagStudio/issues/287#issue-2352251057)
> 
> @Takanu, you're using GNOME, right? This bug isn't uncommon for Qt applications running on GNOME 40+, especially if they don't use KDE Frameworks’ workarounds.

I am.  Is there a workaround for this?

beedell-roke:
> https://github.com/TagStudioDev/TagStudio/issues/287#issuecomment-2365457325

[...]

PR #1328, opened 2026-03-02
fix: MacOS system theme fix (#999)

### LINKED ISSUE #999: [Bug]: `System` theme seems to not work
### Checklist

### TagStudio Version

Alpha 9.5.2

macOS 15.5

### Description

Using the `System` theme within settings seems to keep the app always in dark mode even when in light mode

Image

### Expected Behavior

If the system theme is set to light it should be light and if the system theme is set to dark then use dark

### Steps to Reproduce

1. Set system theme to light
2. Set TagStudio theme to System
3. Restart TagStudio

### Logs

_No response_

terahidro2003:
### Summary
This short PR introduces a fix to a bug related to the issue #999 where system theme was not enforced in MAC OS. 

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [X] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [X] Basic functionality
    -   [X] Added basic unit test for my changes

### DIFF
--- a/src/tagstudio/qt/ts_qt.py
+++ b/src/tagstudio/qt/ts_qt.py
@@ -306,13 +306,13 @@ class QtDriver(DriverMixin, QObject):
             sys.argv += ["-platform", "windows:darkmode=2"]
         self.app = QApplication(sys.argv)
         self.app.setStyle("Fusion")
-        if self.settings.theme == Theme.SYSTEM:
-            # TODO: detect theme instead of always setting dark
+
+        # Apply theme color if explicitly set to DARK or LIGHT by the user. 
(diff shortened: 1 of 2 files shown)


CyanVoxel:
*Tested on macOS 15, Fedora 42 (KDE, X11), and Windows 10*

Thank you so much for the fix! As a bonus this seems to fix the issue both on macOS and on Windows. This also fixes a related issue where light theme elements would incorrectly render when in dark mode on Linux (possibly related to #287)

Issue #207, opened 2024-05-22
[Feature Request] Sorting options


Qronikarz:
Sorting is always good to have. Pain to implement, but useful. Anyway here are some sorting methods that may be considered useful:

* random - when you have massive collection and don't particularly care about sorting or want to enjoy a little chaos.
* sort by width,height - for images, videos and gifs
* sort by size - for every file - Czkawka program may be checked for implementation reference,  it has option to show N-amount of biggest files in the scanned folder/disk
* sort by duration - for videos and gifs
* sort by resolution - I don't know if anything uses that and if it even is correct, but I had it for my Work-In-Progress project because I wanted to sort files by their quality. Basically do the multiplication of width X height and sort the results from other files. It should theoretically show you smallest quality content first and the 4K last.
* sort by various dates - modification date, added date, created date etc.

Tishj:
+1 for random, when there are many results for a search it can be useful to randomize them

Qronikarz:
You can use the thumbs up react instead of +1
This way it counts when you sort the issues by the reactions. Unless you only want the random sort to be added.

simjeehza:
I'll be working on it starting with the created date as its on the Roadmap and related to Issue #68.

CyanVoxel:
> random - when you have massive collection and don't particularly care about sorting or want to enjoy a little chaos.

Random sorting added in #1029

Doragon-S:
Also Sort by filetype?
It kind of feels like there should be a sorting for every file category 
sorting by author/artist for books/music/etc.
etc.

Qronikarz:
I can convert the original post into this tasklist if you want. Could be easier for tracking down the progress. Just say so and I'll edit it.

EDIT: done

Computerdores:
sounds like a good idea to do that @Qronikarz

PR #1330, opened 2026-03-05
feat: add sorting by file size

baracchande:
### Summary

`core/library/alchemy/enums.py`: Added `SIZE = "file.size"` to `SortingModeEnum`

`core/library/alchemy/library.py`
- `search_library()`: when SIZE mode is active, skips SQL ordering (falls back to Entry.id) and calls `_sort_ids_by_file_size()` after the query returns.
- New method `_sort_ids_by_file_size()`: bulk-fetches all entry paths in one query, calls stat().st_size on each file, returns the IDs sorted by size. Missing/unlinked files get size -1 and sort to the front (ascending).

Added "file.size" key with a proper translation in each translation file.

`tests/test_search.py`: Four new tests.

This implements the file size sort option requested in [207](https://github.com/TagStudioDev/TagStudio/issues/207)

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
-   Tested For:
    -   [x] Basic functionality


src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1086,6 +1089,9 @@ def search_library(
             end_time = time.time()
             logger.info(f"SQL Execution finished ({format_timespan(end_time - start_time)})")
 
+            if is_size_sort:
+                ids = self._sort_ids_by_file_size(ids, search.ascending)
+
```
Computerdores, line 1097: This breaks when the query has a LIMIT due to the page size.

Computerdores (changes requested):

Computerdores:
Also, unless you speak 19 languages those are AI/machine translations and I would prefer not having those in here.

Because while I'm not sure whether we have an official policy on AI/machine translations, I would personally prefer a high quality of translations over a high quantity (and people are less likely to contribute when translations are suboptimal rather than missing), so I would prefer a missing translation over an AI one (also should we redecide in the future we can do the AI translations ourselves, where as we won't have any idea how good our translations are if we allow AI translations).

src/tagstudio/core/library/alchemy/library.py
baracchande: Changed the sort logic to handle the query having page size.

baracchande (commented):
size sort now handles paginated search_library calls

baracchande (commented):
Changed the sort logic to handle the query having page size.

baracchande (commented):
add sorting by file size

baracchande (commented):
Changed the sort logic to handle the query having page size.

Issue #1339, opened 2026-04-25
[Bug]+security: drop shell=True from subprocess calls


reddraconi:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

9.5.6

### Operating System & Version

Fedora 43 x86_64

### Description

I've gotten some time back from work and wanted to finish up the features I started back around December. As I did, I ran some security checks and found that, on the Windows path, the `shell=True` can lead to process injections when the command string is being built and sent to `cmd.exe`. One the POSIX path, it looks like we're  vulnerable to backtick substitution.

Since Windows has weird rules about valid characters in file and path names (&, ^, %) that get interpreted by `cmd.exe`, this can lead to command injections when TagStudio opens a mal-named file, path, library, or when ripgrep is called during a library scan.

To fix this up, I updated the subprocess commands to use lists of arguments and removed `shell=True`. I didn't touch anything for regular file names, just stuff that was being passed to the shell.

Demo attached.

### Expected Behavior

Don't have command injection issues due to malformed/malicious file paths or file names.

### Steps to Reproduce

Pre fix:
Windows:
  1. In Settings, enable the "use Windows start command" toggle
  2. Create a folder name containing `&`:
     `C:\Users\<you>\Pictures\album & calc & test`.
  3. Add an image to the folder and import it into a TagStudio library.
  4. Click the file's "Open" action.
  5. Calc.exe should open

Linux:
1. Create a library directory containing a `$(…)` command substitution:
     `mkdir -p '/tmp/My$(gnome-calculator)Library/.TagStudio'`.
  2. Open that directory as a TagStudio library and trigger a refresh
  3. gnome-calc should open

Post-fix:
Both of these should fail since the paths are now lists of strings rather than direct commands.

### Logs

I wrote this up on my Linux box, since my Windows machine died (again).

I ensured:
- TagStudio still  opened files scans libraries, opens files in my file manager, and refreshes my library via ripgrep
- The existing test suite passed everything that was already passing
- ruff and mypy came back clean.

I have a PR coming for this now.

[exploit_demo.py](https://github.com/user-attachments/files/27072075/exploit_demo.py)

PR #1340, opened 2026-04-25
fix(security): drop shell=True from subprocess invocations

reddraconi:
Removes `shell=True` from the three Windows-platform subprocess calls in qt/utils/file_opener.py and the ripgrep invocation in core/library/refresh.py.

Each call now passes its arguments as a list and goes straight to execvp / CreateProcess, so filenames with shell metacharacters (&, ^, %, (), $(...), backticks) aren't parsed by cmd.exe or /bin/sh.

    - "Show in Explorer" passes ["explorer", f"/select,{path}"].
    - The Windows-start fallback passes ["cmd", "/c", "start", "", path].
    - The default Windows open uses `os.startfile`.
    - ripgrep is invoked with its arguments as a list.

Behavior should be unchanged for ordinary filenames.

We passed 231 of 231 tests. ruff and mypy came back clean.

Refs TagStudioDev/TagStudio#1339

### Summary

#1339 contains the vulnerability description, the Windows and POSIX reproduction steps, and the standalone reproducer script.

This PR removes `shell=True` from the four subprocess invocations noted in #1339 and replaces each with the equivalent list-argument form to bypass the shell entirely:

- `src/tagstudio/qt/utils/file_opener.py`
  - "Show in Explorer" passes `["explorer", f"/select,{path}"]` directly.
  - Windows-start fallback passes `["cmd", "/c", "start", "", path]`.
  - Default Windows open switches to `os.startfile`.
- `src/tagstudio/core/library/refresh.py`
  - ripgrep is invoked with its arguments as a list; the manual quoting of the ignore-file path is also removed.

No new dependencies. No user-visible change for valid filenames.

### Tasks Completed

- Platforms Tested:
  - [ ] Windows x86 
  - [ ] Windows ARM
  - [ ] macOS x86
  - [ ] macOS ARM
  - [x] Linux x86
  - [ ] Linux ARM
- Tested For:
  - [x] Basic functionality (open file, reveal in file manager, library refresh via ripgrep on Linux)
  - [ ] PyInstaller executable
  - [x] Existing test suite: 231/231 passing
  - [x] `ruff check` clean
  - [x] `mypy --config-file pyproject.toml .` clean

### DIFF
--- a/src/tagstudio/qt/utils/file_opener.py
+++ b/src/tagstudio/qt/utils/file_opener.py
@@ -2,6 +2,7 @@
 # Licensed under the GPL-3.0 License.
 # Created for TagStudio: https://github.com/CyanVoxel/TagStudio
 
+import os
 import shutil
 import subprocess
 import sys
@@ -40,39 +41,24 @@ def open_file(path: str | Path, file_manager: bool = False, windows_start_comman
         if sys.platform == "win32":
             normpath = str(Path(path).resolve())
             if file_manager:
-                command_name = "explorer"
-                command_arg = f'/select,"{normpath}"'
-
-                # For some reason, if the args are passed in a list, this will error when the
-                # path has spaces, even while surrounded in double quotes.
+                # /select, takes the path as a single comma-joined argument.
                 silent_popen(
-                    command_name + command_arg,
-                    shell=True,
+                    ["explorer", f"/select,{normpath}"],
                     close_fds=True,
                     creationflags=subprocess.CREATE_NEW_PROCESS_GROUP
                     | subprocess.CREATE_BREAKAWAY_FROM_JOB,
                 )
             else:
                 if windows_start_command:
-                    command_name = "start"
-                    # First parameter is for title, NOT filepath
-                    command_args = ["", normpath]
+                    # First "" is the start-command title slot, not the filepath.
                     subprocess.Popen(
-                        [command_name] + command_args,
-                        shell=True,
+                        ["cmd", "/c", "start", "", normpath],
                         close_fds=True,
                         creationflags=subprocess.CREATE_NEW_PROCESS_GROUP
                         | subprocess.CREATE_BREAKAWAY_FROM_JOB,
                     )
                 else:
-                    command = f'"{normpath}"'
-                    silent_popen(
-                        command,
-                        shell=True,
-                        close_fds=True,
-                        creationflags=subprocess.CREATE_NEW_PROCESS_GROUP
-                        | subprocess.CREATE_BREAKAWAY_FROM_JOB,
-                    )
+                    os.startfile(normpath)
         else:
             if sys.platform == "darwin":
                 command_name = "open"
--- a/src/tagstudio/core/library/refresh.py
+++ b/src/tagstudio/core/library/refresh.py
@@ -93,19 +93,16 @@ class RefreshTracker:
                 pattern_file.write("\n".join(ignore_patterns))
 
             result = silent_run(
-                " ".join(
-                    [
-                        "rg",
-                        "--files",
-                        "--follow",
-                        "--hidden",
-                        "--ignore-file",
-                        f'"{str(compiled_ignore_path)}"',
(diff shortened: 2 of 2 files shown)


src/tagstudio/qt/utils/file_opener.py
```diff
@@ -40,39 +41,24 @@ def open_file(path: str | Path, file_manager: bool = False, windows_start_comman
...
                 )
             else:
                 if windows_start_command:
-                    command_name = "start"
-                    # First parameter is for title, NOT filepath
-                    command_args = ["", normpath]
+                    # First "" is the start-command title slot, not the filepath.
                     subprocess.Popen(
-                        [command_name] + command_args,
-                        shell=True,
```
Computerdores: I'm not on windows so I can't check, but it seems to me like removing `shell = True` and then calling the shell manually instead would still cause shell meta chars to be parsed (as it would with `bash -c "<cmd>"`. (Although the syntax of `/c` is different from bash so maybe not)

reddraconi: Craaaaaap. You're right. I think it'll still expand stuff like `%VAR%`, too after `/c`.

If we have to keep the edge case and still use `cmd`, we could wrap the whole path in double quotes and parse the guts of the string for internal double quotes and percents, then toss them? Or just reject the string. It wouldn't be perfect, but it's better than sending raw stuff to `cmd`.

Another option is PowerShell, but that could run into ExecutionPolicy issues, depending on how the host is configured. I know *I* lock down ExecutionPolicy for my users via group policy to make sure they don't get up to shenanigans.

Something like this with `-LiteralPath` could work since it won't expand into anything weird:

```python
  def _open_with_powershell(normpath: str) -> None:
      escaped = normpath.replace("'", "''")
      ps_script = f"Start-Process -FilePath '{escaped}'"
      # -EncodedCommand wants base64 of UTF-16LE.
      encoded = base64.b64encode(ps_script.encode("utf-16-le")).decode("ascii")
      silent_popen(
          ["powershell", "-NoProfile", "-NonInteractive", "-EncodedCommand", encoded],
          close_fds=True,
          creationflags=subprocess.CREATE_NEW_PROCESS_GROUP
          | subprocess.CREATE_BREAKAWAY_FROM_JOB,
      )

  if windows_start_command:
      _open_with_powershell(normpath)
  else:
      os.startfile(normpath)
```

Otherwise, if we don't need the edge case, `os.startfile` path would still be the way to do it.

Thoughts?

Computerdores: I think we should be able to only use `os.startfile` here?
There was a [discussion](https://discordapp.com/channels/1229183630228848661/1232037962116759603/1349454071454761042) on the discord last year, however, even after re-reading the discussion now, I am still not quite sure where the difference is between `os.startfile` and manually calling the cmd-builtin `start`.
@Xarvex do you perhaps remember?

xarvex: > I think we should be able to only use `os.startfile` here? There was a [discussion](https://discordapp.com/channels/1229183630228848661/1232037962116759603/1349454071454761042) on the discord last year, however, even after re-reading the discussion now, I am still not quite sure where the difference is between `os.startfile` and manually calling the cmd-builtin `start`. @xarvex do you perhaps remember?

@Computerdores can't quite remember the operational differences, from my understanding `os.startfile` calls the `start` built-in internally anyway. However, what I did find is a discussion about a contextual difference, where it was stated that `os.startfile` executes `start` in a way where it is a subprocess of TagStudio (see [this message](https://discord.com/channels/1229183630228848661/1229309667528806420/1232902165006385234) and the surrounding context).

Supposedly when `os.startfile` was used, whenever TagStudio exited, whatever application was opened for viewing the file/folder also exited. For `popen`, we can use `subprocess.CREATE_NEW_PROCESS_GROUP` and `subprocess.CREATE_BREAKAWAY_FROM_JOB` to prevent that behavior, but I don't see any similar controls for `os.startfile`.

Issue #1341, opened 2026-04-25
[Bug]+security: cap XML/archive reads to bound memory use


reddraconi:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### TagStudio Version

9.5.6

### Operating System & Version

Fedora 43 x86_64

### Description

There's a few spots where we don't have guardrails that could lead to TagStudio crashing or chewing up all of a system's memory if something malicious or malformed is encountered. Specifically in XML- and archive-based formats.

I understand this is a local app and it's the user's fault if they knowingly feed it bad data.

Since TagStudio uses Python's native `xml.etree.ElementTree` to parse XML, that leaves it open to Billion Laughs and XXE attacks. Several callers read an untrusted length straight into `f.read()` or `archive.read()` before parsing the file, so even a better parser like `defusedxml` wouldn't save us.

This needs a two-part fix:
1. Swap `xml.etree.ElementTree` for `defusedxml.ElementTree` anywhere we touch untrusted XML. 
2. Size-gate those reads with a sane cap.

This is currently a problem in:
1. ePub/CB* ComicInfo.xml and _epub_cover
2. Legit `.mpd` file headers are a max of 32-bits, but have an unbounded read
3. Legit `.pdn` file headers are a max of 24-bits, but have an unbounded read
4. DupeGuru XML results are only bounded by system memory.

PR coming with more details and tests.

### Expected Behavior

Parse XML and archives without being vulnerable to Billion Laughs / XXE / OOM issues.

### Steps to Reproduce

A maliciously crafted `.mdp`:

```python
  import struct
  # Magic + bin_header claiming a 1 GiB XML header that doesn't exist.
  with open("/tmp/bomb.mdp", "wb") as f:
      f.write(b"mdipack\x00")
      f.write(struct.pack("<LLL", 0, 1024 * 1024 * 1024, 0))
```
A maliciously crafted `.pdn`:

```python  
  # Magic "PDN3" + 24-bit little-endian header_size. The format caps the
  # field at ~16 MiB (0xFFFFFF), so the worst case is bounded by the spec,
  # but that's still 16-ish MiB allocated before defusedxml runs, and
  # nothing stops a user from receiving a file that hits that ceiling.
[...]

PR #1342, opened 2026-04-25
fix(security): Cap untrusted XML/archive reads to bound memory use

reddraconi:
Two-part fix to TagStudio's XML/archive parsing through two commits to address #1341

NOTE: This adds a new dependency, `defusedxml.ElementTree` and updates the settings UI to include a new **Advanced** page for these (probably-not-often-touched settings).

These are to keep the app from crashing if it eats some bad XML data. Probably a low-threat case since this is all local anyway.

1. Swap `xml.etree.ElementTree` to `defusedxml.ElementTree`

The stdlib parser is known to be unsafe on untrusted input. It happily expands DTD entities (billion-laughs /
OOM bombs), follows external-entity references that read local files or fetch over HTTP (XXE), and applies
no internal bounds. `defusedxml` is the recommended drop-in replacement that disables those parser behaviors by default. The  `Element` import is kept from stdlib since `defusedxml` only ships replacement parsers, not data classes.

We left the legacy `core/library/json/library.py` `xml.etree.ElementTree.parse` alone, per
`CONTRIBUTING.md` guidance. If we need to swap that too, we can.

2. Cap attacker-controlled XML reads

`defusedxml` only protects the parser. Since there's a handful of calls that happen *before* the parser can protect us, we'd like to cap those read sizes so a malicious or accidentally-large input doesn't OOM TagStudio

Each path now checks the declared/on-disk size against a byte cap before allocating, raising `ValueError` with a message that names the exact GlobalSettings field to raise. Caps live as fields on `GlobalSettings` (canonical defaults in `core/constants.py`) so users can raise them if they need to:

| Setting | Default MiB | Protects |
|---|---|---|
| `comic_info_max_mb` | 1 | ePub / CB\* `ComicInfo.xml` zip-bomb members |
| `mdp_header_max_mb` | 1 | `.mdp` (MDIPACK) headers — 32-bit declared length |
| `pdn_header_max_mb` | 16 | `.pdn` (Paint.NET) headers — 24-bit declared length (matches format ceiling) |
| `dupe_results_max_mb` | 128 | DupeGuru results XML before `ET.parse` |

A new  **Advanced** tab in the settings panel exposes these settings. The renderer paths' `ValueError` is caught by each method's exception handler and throws as `Couldn't render thumbnail error=ValueError` log line, so users hitting their cap get useful messages. The dupe registry raises `ValueError` before `ET.parse` runs
and bubbles up to the caller in `fix_dupe_files.py`.

`_epub_cover`, `_mdp_thumb`, and `_pdn_thumb` are no longer `@staticmethod` because they now read `self.driver.settings`. All existing call sites already used `self.<method>(...)`, so this should be a no-op for callers.

### Tasks Completed

- Platforms Tested:
  - [ ] Windows x86
  - [ ] Windows ARM
  - [ ] macOS x86
  - [ ] macOS ARM
  - [X] Linux x86
  - [ ] Linux ARM
- Tested For:
  - [x] Basic Functionality
  - [ ] PyInstaller Executable
  - [x] `pytest`: All 231 tests still pass. I added 5 new tests 
  - [x] `ruff` and `mypy` came back clean.
  - [x] Manual: Opened the new Advanced tab, changed values, save + reloaded to ensure values persisted.


CyanVoxel:
Closing due to the closure of #1341.

I appreciate your desire to contribute here, however the niche nature of the original issue as well as the work required to mitigate the "issue" would arguably fall under a recent section in our CONTRIBUTING.md's "Unnaceptable Code" section:
> Code that adds a drastic amount of complexity with minimal utility

PR #1198, opened 2025-11-11
fix(Mirror Entries): repaired Entry mirroring (#1197)

louisl4696:
### Summary

Closes #1197

Tags and fields are now correctly mirrored across Entries when clicking the _Mirror Entries_ button in the _Fix Duplicate Files_ menu 

 - Updated DupeFilesRegistry method refresh_dupe_files() to use library.get_entry_full() in _dupe_files_registry.py_
 _entries previously returned with get_entries() were incomplete and causing errors later down the line_
 - Created mirror_entry_tags() in _library.py_
 - fixed mirror_entry_fields() in _library.py_
 _field mirroring was inconsistent, skipping all fields present on the reference Entry. Now, already present fields are skipped on a per-Entry basis._
 - updated mirror_entries_runnable() to also call lib.mirror_entry_tags() in _mirror_entries_modal.py_

NOTE: this will be difficult to test until #1196 is fixed. See [my thread](https://discord.com/channels/1229183630228848661/1436449668413263953/1436457499204128899) in discord for a quick hack to get around that bug and be able to test the functionality of the changes in this PR.

### Tasks Completed

-   Platforms Tested:
    -   [x] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality
    -   [ ] PyInstaller executable

PR #1354, opened 2026-05-08
refactor!: uncouple fields from hard-coded values

CyanVoxel:
> [!WARNING]
> This PR bumps the **major** version of the library database to **200**. Please backup any existing libraries before testing.

### Summary
This PR uncouples fields from hard-coded values such as enums and immutable SQLite tables, allowing for the *foundation* of independently customizable field names and user field templates (intended for a future PR).

Legacy "TEXT_LINE" and "TEXT_BOX" field types are now a single `TextField` type, distinguished by an `is_multiline` boolean attribute. This allows for a future UI change where all text fields can be edited in a single modal that adapts to the attributes present and allows for any text attributes to be changed for the field.

I've also reduced the set of default field "templates", in preparation for a future PR that allows for the creation of user created field templates. Unlike the built-in tags, these fields are merely a jumping-off point for users to customize or remove as they see fit.
image

These changes also open the door for additional field type attributes, for example `is_rich_text` or `is_hyperlink` for `TextFields` or `is_time` for `DatetimeFields`.

**NOTE:** I was unable to remove the now-unused `type_key` columns from the `text_fields` and `datetime_fields` tables for existing libraries due to them containing foreign key relationships. This is not an urgent issue, but is something that should be addressed in the future. See this issue with dropping columns foreign keys in SQLite: https://www.sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_changes

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [x] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality


src/tagstudio/core/library/alchemy/library.py
```diff
@@ -296,24 +312,48 @@ def migrate_json_to_sqlite(self, json_lib: JsonLibrary):
...
+                                    ),
+                                )
+                            elif LEGACY_FIELD_MAP[legacy_field_id]["type"] == DatetimeField:
+                                self.add_datetime_field_to_entry(
+                                    entry_id=(
+                                        entry.id + 1
+                                    ),  # NOTE: JSON IDs start at 0 instead of 1
+                                    name=str(LEGACY_FIELD_MAP[legacy_field_id]["name"]),
+                                    value=value,
+                                )
```
Computerdores:
```suggestion
                            field_info = LEGACY_FIELD_MAP[legacy_field_id]
                            if field_info["type"] == TextField:
                                self.add_text_field_to_entry(
                                    entry_id=(
                                        entry.id + 1
                                    ),  # NOTE: JSON IDs start at 0 instead of 1
                                    name=str(field_info["name"]),
                                    value=value,
                                    is_multiline=bool(field_info["is_multiline"]),
                                )
                            elif field_info["type"] == DatetimeField:
                                self.add_datetime_field_to_entry(
                                    entry_id=(
                                        entry.id + 1
                                    ),  # NOTE: JSON IDs start at 0 instead of 1
                                    name=str(field_info["name"]),
                                    value=value,
                                )
```

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -454,6 +488,22 @@ def open_sqlite_library(
...
+                            session.add(
+                                TextFieldTemplate(name=ft.name, is_multiline=ft.is_multiline)
+                            )
+                        elif type(ft) is DatetimeFieldTemplate:
+                            session.add(DatetimeFieldTemplate(name=ft.name))
+
+                        session.commit()
+                    except IntegrityError:
+                        logger.info("[Library] FieldTemplate already exists", field_template=ft)
+                        session.rollback()
```
Computerdores, line 500: I don't really see the point in recreating the templates here, so I would suggest either:
```suggestion
            if is_new:
                for template in get_default_field_templates():
                    try:
                        session.add(template)
                        session.commit()
                    except IntegrityError:
                        logger.info("[Library] FieldTemplate already exists", field_template=template)
                        session.rollback()
```
or if there is a reason for that, I would still improve the variable name here:
```suggestion
            if is_new:
                for template in get_default_field_templates():
                    try:
                        if type(template) is TextFieldTemplate:
                            session.add(
                                TextFieldTemplate(name=template.name, is_multiline=template.is_multiline)
                            )
                        elif type(template) is DatetimeFieldTemplate:
                            session.add(DatetimeFieldTemplate(name=template.name))

                        session.commit()
                    except IntegrityError:
                        logger.info("[Library] FieldTemplate already exists", field_template=template)
                        session.rollback()
```

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -736,16 +763,112 @@ def __migrate_sql_to_ts_ignore(self, library_dir: Path):
...
+                try:
+                    if type(ft) is TextFieldTemplate:
+                        session.add(TextFieldTemplate(name=ft.name, is_multiline=ft.is_multiline))
+                    elif type(ft) is DatetimeFieldTemplate:
+                        session.add(DatetimeFieldTemplate(name=ft.name))
+
+                    session.commit()
+                except IntegrityError:
+                    logger.info("[Library] FieldTemplate already exists", field_template=ft)
+                    session.rollback()
```
Computerdores, line 847: This should be a single transaction imo, because otherwise if one of the steps fails for whatever reason the DB will be left in a corrupt state where some of the migration steps were already done and will thus fail in the future, but some haven't been done. So every future attempt a the migration would fail. However, as one transaction none of the steps would be saved to disk and thus the user could report the bug, pull the fix and try again.

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1169,165 +1292,136 @@ def remove_tag(self, tag_id: int) -> bool:
...
-                if rows:
-                    session.commit()
-
     def remove_entry_field(
         self,
         field: BaseField,
         entry_ids: list[int],
     ) -> None:
-        FieldClass = type(field)  # noqa: N806
+        field_ = type(field)
```
Computerdores: this naming seems worse than before to me

PS: there's multiple occurrences of this var name

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1867,39 +1961,75 @@ def set_version(self, key: str, value: int) -> None:
...
+                            value=field.value,
+                            is_multiline=field.is_multiline,
+                        )
+                    elif type(field) is DatetimeField:
+                        self.add_datetime_field_to_entry(
+                            entry_id=entry.id, name=field.name, value=field.value
+                        )
+            logger.info(
+                "[Library][mirror_fields]", entry_id=entry.id, field_count_after=len(entry.fields)
+            )
```
Computerdores: Wouldn't it make more sense to just clone (with a mind for sqlalchemy bs) the original field instance irrespective of what type of field it is?

src/tagstudio/core/ts_core.py
```diff
@@ -27,170 +21,6 @@ class TagStudioCore:
     def __init__(self):
         self.lib: Library = Library()
 
-    @classmethod
-    def get_gdl_sidecar(cls, filepath: Path, source: str = "") -> dict:
```
Computerdores, line 31: I would suggest throwing a note into the pr body about removing this, so that it is searchable in case someone wants to reimplement gallery dl support in the future

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1169,165 +1292,136 @@ def remove_tag(self, tag_id: int) -> bool:
...
             session.execute(update_stmt)
             session.commit()
 
-    @property
-    def field_types(self) -> dict[str, ValueType]:
-        with Session(self.engine) as session:
-            return {x.key: x for x in session.scalars(select(ValueType)).all()}
+    def add_text_field_to_entry(
+        self, entry_id: int, name: str, value: str | None = None, is_multiline: bool = False
+    ) -> bool:
```
Computerdores: I think it would make more sense to directly accept a `TextField` here (and `DatetimeField` in the other method) as there are quite a lot of places where this is called by decomposing a template which is constructed in the method anyways.

src/tagstudio/qt/mixed/field_containers.py
```diff
@@ -250,36 +266,45 @@ def write_container(self, index: int, field: BaseField, is_mixed: bool = False):
...
             self.containers.append(container)
             self.scroll_layout.addWidget(container)
         else:
             container = self.containers[index]
 
-        if field.type.type == FieldTypeEnum.TEXT_LINE:
-            container.set_title(field.type.name)
+        # Set field title
+        field_name_key: str = FIELD_TYPE_KEYS.get(field.__class__.__name__, "field_type.unknown")
+        title = f"{field.name} ({Translations[field_name_key]})"
```
Computerdores: there are multiple places where this exact code is used, move this to a method (perhaps of the the base class for the fields?)

src/tagstudio/qt/mixed/field_containers.py
```diff
@@ -460,26 +485,24 @@ def remove_field(self, field: BaseField):
...
+        """Update a datetime field across selected entries."""
+        entry_ids = [e.id for e in self.cached_entries]
         assert entry_ids, "No entries selected"
-        self.lib.update_entry_field(
-            entry_ids,
-            field,
-            content,
-        )
+
+        self.lib.update_datetime_field(entry_ids, field, dt.fromisoformat(value))
```
Computerdores, line 488: Doesn't this break when there are multiple instances of the same field type on the same entry? (e.g. multiple authors)

src/tagstudio/resources/translations/en.json
```diff
@@ -71,6 +71,9 @@
     "entries.unlinked.unlinked_count": "Unlinked Entries: {count}",
     "ffmpeg.missing.description": "FFmpeg and/or FFprobe were not found. FFmpeg is required for multimedia playback and thumbnails.",
     "ffmpeg.missing.status": "{ffmpeg}: {ffmpeg_status}<br>{ffprobe}: {ffprobe_status}",
+    "field_type.datetime": "Datetime",
+    "field_type.text": "Text",
+    "field_type.unknown": "Unknown Type",
```
Computerdores, line 76: If you want you can throw this into the german translation:
```
    "field_type.datetime": "Datum - Uhrzeit",
    "field_type.text": "Text",
    "field_type.unknown": "Unbekannter Typ",
```

tests/macros/test_dupe_files.py
```diff
@@ -19,13 +20,13 @@ def test_refresh_dupe_files(library: Library):
     entry = Entry(
         folder=folder,
         path=Path("bar/foo.txt"),
-        fields=library.default_fields,
+        fields=[TextField(name="Title", value="I'm a Test Title")],
```
Computerdores: dedup the list perhaps?

tests/test_library.py
```diff
@@ -207,13 +207,18 @@ def test_remove_entry_field(library: Library, entry_full: Entry):
     assert not entry.text_fields
 
 
-def test_remove_field_entry_with_multiple_field(library: Library, entry_full: Entry):
+def test_remove_text_field_entry_with_multiple_field(library: Library, entry_full: Entry):
```
Computerdores:
```suggestion
def test_remove_text_field_entry_with_multiple_fields(library: Library, entry_full: Entry):
```

Computerdores (commented):
welp that was a huge PR ^^
didn't test it locally and probably won't have time to do so either

Computerdores:
I think this conflicts with #1198 btw

src/tagstudio/core/library/alchemy/library.py
CyanVoxel: The issue I was facing was the extra extraneous values as a part of the fields (namely `id` and `entry_id`) getting in the way of determining what I would consider to be "unique" fields for this case (a test case I've also added now, should've had that before)

However, by overriding the `__eq__()` and/or `__hash__()` methods for the fields, I can get rid of all the tuple nonsense here. The question is whether or not other parts of the code might rely on the original equality checks and hashes, but off the top of my head I can't think of anything...

```py
# TextField class
@override
    def __eq__(self, other: object) -> bool:
        if not isinstance(other, TextField):
            return False

        return (self.name, self.value, self.is_multiline) == (
            other.name,
            other.value,
            other.is_multiline,
        )

    @override
    def __hash__(self) -> int:
        return hash((self.name, self.value, self.is_multiline))
```

src/tagstudio/core/ts_core.py
CyanVoxel: I've already re-implemented it on the [macros](https://github.com/TagStudioDev/TagStudio/tree/macros) branch, I just need to rebase that and get a PR going sooner or later

src/tagstudio/core/library/alchemy/library.py
CyanVoxel: In that case, I could also recombine these two methods back into a single `add_field_to_entry()` method

src/tagstudio/qt/mixed/field_containers.py
CyanVoxel: It's only used in this one spot (minus a couple related log statements). I'll keep that in mind in case it changes though

src/tagstudio/qt/mixed/field_containers.py
CyanVoxel: I haven't run into anything like that, are you able to produce that?

src/tagstudio/resources/translations/en.json
CyanVoxel: I'll wait to do that on the Weblate side afterwards 👍

src/tagstudio/core/library/alchemy/library.py
CyanVoxel: I believe I've got it into a single transaction now, please let me know if something doesn't look right

Computerdores: looks good to me

src/tagstudio/core/library/alchemy/library.py
Computerdores: I can't think of anything either and I wouldn't have relied on it myself either, after all, who knows what kind of fields sqlalchemy might be adding that would be part of the default equality check...

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1366,61 +1349,51 @@ def update_datetime_field(
...
-            except IntegrityError as e:
-                logger.error(e)
-                session.rollback()
-                return False
+            field = TextField(
+                entry_id=entry_id,
+                name=field.name,
+                value=field.value,
+                is_multiline=field.is_multiline,
+            )
```
Computerdores: It might be nicer to give BaseField a virtual method `clone_with_id` that the subclasses implement, because this function shouldn't need to be changed when a new field type is added or an existing gets or looses a property.

Also, two things this would fix that should be fixed either way:
- the `with` statement for DB add is duplicated and this would fix that for free.
- I would add an else with a `NotImplementedError` to make sure this gets updated when a new field is introduced

src/tagstudio/qt/mixed/field_containers.py
```diff
@@ -220,23 +220,18 @@ def add_field_to_selected(self, field_list: list[QListWidgetItem]):
...
-                    self.lib.add_datetime_field_to_entry(
-                        entry_id=entry_id,
-                        name=field_.name,
-                    )
+                if type(template) is TextFieldTemplate:
+                    text_field = TextField(name=template.name, is_multiline=template.is_multiline)
+                    self.lib.add_field_to_entry(entry_id, text_field)
+                elif type(template) is DatetimeFieldTemplate:
+                    datetime_field = DatetimeField(name=template.name)
+                    self.lib.add_field_to_entry(entry_id, datetime_field)
```
Computerdores: I would suggest adding a virtual `make_field` method to BaseFieldTemplate that `TextFieldTemplate`, etc. implement. Otherwise I would add an else with a `NotImplementedError` here to make sure the method gets updated when a new Field is introduced.

src/tagstudio/qt/mixed/field_containers.py
Computerdores: Hmm technically you're correct, although, see `AddFieldModal.show` — the code there is extremely similar

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1169,165 +1275,126 @@ def remove_tag(self, tag_id: int) -> bool:
...
         self,
         entry_ids: list[int] | int,
-        field: BaseField,
-        content: str | datetime,
+        field: DatetimeField,
+        value: datetime,
     ):
+        """Update a DatetimeField field on one or more Entries."""
         if isinstance(entry_ids, int):
             entry_ids = [entry_ids]
```
Computerdores, line 1332: `entry_ids` is never used, I suspect that the update statement is supposed to filter on it

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1169,165 +1275,126 @@ def remove_tag(self, tag_id: int) -> bool:
...
             session.commit()
 
-        # recalculate the remaining positions
-        # self.update_field_position(type(field), field.type, entry_ids)
+    def update_text_field(
+        self, entry_ids: list[int] | int, field: TextField, value: str, is_multiline: bool
+    ):
+        """Update a TextField field on one or more Entries."""
+        if isinstance(entry_ids, int):
+            entry_ids = [entry_ids]
```
Computerdores, line 1310: `entry_ids` is never used, I suspect that the update statement is supposed to filter on it

src/tagstudio/qt/mixed/field_containers.py
Computerdores: nvm, misread

src/tagstudio/core/library/alchemy/library.py
Computerdores: even better 👍

src/tagstudio/qt/mixed/field_containers.py
CyanVoxel: I went ahead and added a `class_name` property for both the `BaseField` and `BaseFieldTemplate` classes

Computerdores (approved):
lgtm

CyanVoxel:
@Computerdores Thank you so much for the review!!

PR #1362, opened 2026-05-11
docs/chore: add .editorconfig and .prettierrc.toml configs

CyanVoxel:
### Summary

This PR adds [EditorConfig](https://editorconfig.org/#example-file) and [Prettier](https://prettier.io/) config files for use in formatting non-Python files (Markdown, JSON, YAML, HTML, CSS, etc.) and updates the STYLE.md to reference them.

If approved, I can follow this up with an additional PR that applies the solidified formatting rules across all applicable files and put that squashed commit inside `.git-blame-ignore-revs`. 

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [ ] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [ ] Basic functionality

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -18,8 +18,5 @@ trim_trailing_whitespace = true
 [*.{css,json,md}]
 indent_size = 4
 
-[*.md]
-trim_trailing_whitespace = false
-
 [*.{yaml,yml}]
 indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,25 @@
+# SPDX-FileCopyrightText: (c) TagStudio Contributors
+# SPDX-License-Identifier: MIT
+
+# EditorConfig https://editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,25 @@
...
+indent_style = space
+insert_final_newline = true
+max_line_length = 100
+trim_trailing_whitespace = true
+
+[*.{css,json,md}]
+indent_size = 4
+
+[*.md]
+trim_trailing_whitespace = false
```
Computerdores: Why not trim whitespace here?

Computerdores:
Had a look over it, besides that one comment, looked fine to me. I don't know if there are perhaps other options we'd want to set, but we can always do that later too, so it's not that important imo

.editorconfig
CyanVoxel: Whoops, left that in when referencing other projects' config settings. Just took it out now

Computerdores (approved):

PR #1368, opened 2026-05-13
fix(ci): address remaining pyright errors

CyanVoxel:
### Summary

This PR addresses the remaining pyright errors present either by correcting the underlying issues or suppressing the errors where appropriate. Some warnings and basedpyright-specific issues were addressed as well.

Unblocks #1232.

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [x] macOS ARM
    - [ ] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/src/tagstudio/qt/previews/renderer.py
+++ b/src/tagstudio/qt/previews/renderer.py
@@ -43,7 +43,7 @@ from PIL import (
     UnidentifiedImageError,
 )
 from PIL.Image import DecompressionBombError
-from pillow_heif import register_heif_opener
+from pillow_heif import register_heif_opener  # pyright: ignore[reportUnknownVariableType]
 from PySide6.QtCore import (
     QBuffer,
     QFile,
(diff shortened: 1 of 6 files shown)


src/tagstudio/qt/previews/renderer.py
```diff
@@ -1120,8 +1120,8 @@ def _image_raw_thumb(filepath: Path) -> Image.Image | None:
                 )
         except (
             DecompressionBombError,
-            rawpy.LibRawIOError,
-            rawpy.LibRawFileUnsupportedError,
+            rawpy.LibRawIOError,  # pyright: ignore[reportPrivateImportUsage]
+            rawpy.LibRawFileUnsupportedError,  # pyright: ignore[reportPrivateImportUsage]
```
Computerdores: nit: maybe move this to an import like in the other file

src/tagstudio/qt/previews/renderer.py
```diff
@@ -1137,6 +1137,7 @@ def _image_exr_thumb(filepath: Path) -> Image.Image | None:
         try:
             # Load the EXR data to an array and rotate the color space from BGRA -> RGBA
             raw_array = cv2.imread(str(filepath), cv2.IMREAD_UNCHANGED)
+            assert raw_array
```
Computerdores: nit: I would prefer:

```suggestion
            assert raw_array is not None
```
since that is clearer (also `raw_array` could technically also have a different falsy value that shouldn't trigger an AssertionError)
(same with the other asserts)

Computerdores (approved):

PR #1370, opened 2026-05-13
fix: drop type_key column from text_fields and datetime_fields tables

### LINKED ISSUE #1367: [Bug]: Can't Add Fields to Entries in Migrated Libraries
### Checklist

### TagStudio Version

Alpha 9.5.7 (main branch)

macOS 15

### Description

If a library was migrated from a DB version earlier than 200, then fields can not be added to entries due to a `NOT NULL` constraint on the legacy `type_key` column that was left in due to it being difficult to remove because of a `FOREIGN KEY` constraint. Caused by #1354. 

Removing the now-unused `type_key` column entirely as described [here](https://www.sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_changes) should resolve this issue as well.

### Expected Behavior

Fields should be able to be added to entries in all libraries, regardless of the version they were created in.

### Steps to Reproduce

1. Have a library that was created before DB version 200 (9.5.7 will do, not on the main branch)
2. Try adding a field to a file entry
3. Observe that you can not

### Logs

```yaml
2026-05-12 23:06:56 [error    ] (sqlite3.IntegrityError) NOT NULL constraint failed: text_fields.type_key
[SQL: INSERT INTO text_fields (value, is_multiline, name, entry_id) VALUES (?, ?, ?, ?)]
[parameters: (None, 0, 'Title', 230)]
(Background on this error at: https://sqlalche.me/e/20/gkpj)
```

CyanVoxel:
### Summary

This PR fixes #1367 by dropping the leftover `type_key` columns from the `text_fields` and `datetime_fields` tables. Because SQLite does not allow for the removal of columns with foreign key constraints, I've had to create new versions of these tables without the columns and copy the data from the old tables to the new ones, as outlined in the [SQLite documentation](https://www.sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_changes).

I've also made a related change to enforce the ordering of columns inside of both of these tables. SQLAlchemy by default orders these columns strangely as it combines fields from the base `fields` definition, and I did not wish to canonize that ordering in raw SQL statements for the migration.

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [x] macOS ARM
    - [ ] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable


src/tagstudio/core/library/alchemy/library.py
```diff
@@ -863,6 +865,59 @@ def __apply_db200_migrations(self, session: Session):
...
+            session.execute(create_datetime_fields_table)
+            session.flush()
+            session.execute(
+                text("""
+                    INSERT INTO datetime_fields (id, name, entry_id, value)
+                    SELECT id, name, entry_id, value
+                    FROM datetime_fields_old
+                """)
+            )
+            session.execute(text("DROP TABLE datetime_fields_old"))
```
Computerdores: [According to the sqlite docs this can lead to DB corruption.](https://www.sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_changes:~:text=Caution) The docs also describe an alternate way of the same complexity that should work fine though

Computerdores (commented):
haven't tested so far but found one major issue

src/tagstudio/core/library/alchemy/library.py
CyanVoxel: Ahh I see, I missed the importance of this before. I've changed it to now use the recommended renaming ordering
image

Computerdores (approved):

PR #1373, opened 2026-05-14
feat: allow all raster images to be used as archive thumbnails.

### LINKED ISSUE #1364: [Feature Request]: Thumbnail support for WEBP files inside ZIP
### Checklist

### Description

Certain file formats (namely .webp that I could discover) that are currently supported for thumbnails do not generate while inside of zip/cbz structures as you can see below. The zip files in this example all contain the exact same image of their corresponding format. If a working format is added to the zip, it will automatically update the thumbnail to use the first file that is supported as expected.

Image

### Solution

Assuming that the problem lies in the zip forgetting to search for .webp as a thumbnail option, it should be included. Here's an example from my File Explorer:

Image

### Alternatives

_No response_

Sola-ris:
### Summary

Allow all (except `.exr`) raster images to be used as thumbnails for archive-based files.

Fixes https://github.com/TagStudioDev/TagStudio/issues/1364

I've left out `exr` images since they have to be handled separately from the other raster images and I suspect that they would be pretty rare in archives and especially in comic books.
If I'm wrong or if you think they're worth having I can add them.

#### Before
before

#### After
after

#### Files used for testing
[samples.zip](https://github.com/user-attachments/files/27763020/samples.zip)

All shown files zipped because of upload limitations

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [x] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/src/tagstudio/qt/previews/renderer.py
+++ b/src/tagstudio/qt/previews/renderer.py
@@ -949,9 +949,10 @@ class ThumbRenderer(QObject):
         if cover is not None:
             pages = [f for f in archive.namelist() if f != "ComicInfo.xml"]
             page_name = pages[int(unwrap(cover.get("Image")))]
-            if page_name.endswith((".png", ".jpg", ".jpeg", ".gif", ".bmp", ".svg")):
+            ext = ThumbRenderer.__suffix(page_name)
+            if MediaCategories.IMAGE_RASTER_TYPES.contains(ext):
                 image_data = archive.read(page_name)
-                im = Image.open(BytesIO(image_data))
+                im = ThumbRenderer.__load_raster_image(BytesIO(image_data))
 
         return im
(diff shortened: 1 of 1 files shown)


CyanVoxel:
> I've left out exr images since they have to be handled separately from the other raster images and I suspect that they would be pretty rare in archives and especially in comic books.
If I'm wrong or if you think they're worth having I can add them.

I agree with the decision to leave them out, I think that'd be a bit overkill to support compared to the other formats

CyanVoxel (approved):
Thank you so much for the addition!

PR #1382, opened 2026-05-29
fix: match against the correct path in the ignore registry

TrigamDev:
### Summary

Fixes the "Ignore Entries" tool not correctly finding ignored entries.

This issue was caused by it matching against the path `{library_dir}/{entry_path}`, instead of just the entry path like is matched against elsewhere.

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [x] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/src/tagstudio/core/library/alchemy/registries/ignored_registry.py
+++ b/src/tagstudio/core/library/alchemy/registries/ignored_registry.py
@@ -4,14 +4,12 @@
 
 from collections.abc import Iterator
 from dataclasses import dataclass, field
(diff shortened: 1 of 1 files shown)


CyanVoxel (approved):
As far as I've been able to test I believe this should fix the issue, thank you for you work on this!

PR #1383, opened 2026-06-11
feat(refresh): implement adding file creation and modification date to database when creating new entries 

JonathanGlixman:
### Summary
Implements adding file creation and modification date into the database for new entries during a library refresh, aligning with the v9.6.0 roadmap.

The entry generation logic has been extracted into a private helper method (__create_entry) to cleanly resolve absolute paths for OS stat extraction while keeping the list comprehension scannable.

### Tasks Completed

- Platforms Tested:
    - [x] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [ ] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/src/tagstudio/core/library/refresh.py
+++ b/src/tagstudio/core/library/refresh.py
@@ -1,7 +1,6 @@
 # SPDX-FileCopyrightText: (c) TagStudio Contributors
 # SPDX-License-Identifier: GPL-3.0-only
-
-
+import platform
 import shutil
 from collections.abc import Iterator
 from dataclasses import dataclass, field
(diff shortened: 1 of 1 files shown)


CyanVoxel:
Thank you for taking a look at the roadmap and wanting to contribute!

This is actually something I'm already actively working on in the `file-metadata-table` branch. That implementation also restructures the database to split off file metadata into its own tables, populates file metadata for all existing entries on top of new ones, and keeps that data updated on new scans. Here the file metadata is only added once for new entries, which is still incomplete as a feature implementation. But again, this is something I'm already working on and covering in that other branch, but I appreciate you looking to contribute!

PR #1388, opened 2026-06-15
fix: read and write text config files as UTF-8

### LINKED ISSUE #1345: [Bug]+io: settings files written under system default encoding can fail on non-UTF-8 codepages
### Checklist

### TagStudio Version

9.5.6

Fedora 43 x86_64

### Description

There's a few places where we don't specify UTF-8 as the format for files that TagStudio owns. For Linux and Mac it's probably not a problem since they should default to UTF-8 but on Windows systems that can have non-UTF-8 codepages, this will cause problems.

Examples:
- A user on Japanese Windows (`cp932`) saves a library path containing `写真` to `settings.toml`. The file's written in cp932. If that user opens TagStudio on system with a different Windows codepage, or on Linux, `read_settings()` blows up with `UnicodeDecodeError`.
- TOML and JSON specs both mandate UTF-8, so anything we write that isn't UTF-8 is nonconforming and won't
  round-trip through other tools. (Probably pretty minor, in the grand scheme of things)
- The compiled ignore file (`.compiled_ignore`) passed to `ripgrep` is opened with default encoding, so a user with non-ASCII glob patterns can poison the ripgrep input.
- `ResourceManager.read_text()` reads bundled resource files (English translations, etc.) under the system encoding.

PR inbound.

### Expected Behavior

TagStudio files (configurations like settings.toml, .ts_ignore, etc.) should use a standard codepage to ensure files moved between systems are readable by those systems if one or the other(s) are using a non-UTF-8 codepage.

If a TagStudio-owned file is written with non-UTF-8 formatting, it should be rewritten in UTF-8 on next save.

### Steps to Reproduce

```python
  # Build a settings.toml whose ASCII structure is intact but whose string *value*
  # contains raw cp932 bytes (invalid as UTF-8).
  import pathlib
  legacy = "写真".encode("cp932")  # b"\x8e\xca\x90^"
  payload = b'language = "ja"\ndate_format = "' + legacy + b'"\n'
  pathlib.Path("/tmp/settings.toml").write_bytes(payload)
```

Use that settings file in TagStudio on a non-UTF-8 system with a non-cp932 codepage and watch it fall over.

### Logs

_No response_

yosofbadr:
### Summary

Draft / waiting on broader validation.

Fixes #1345 by reading and writing TagStudio text configuration/resource files with explicit encodings instead of relying on the system default encoding.

This keeps the change focused on text encoding handling: settings are read as UTF-8 first with a fallback for legacy non-UTF-8 files, then saved as UTF-8. Related generated ignore/resource text reads and writes now also specify UTF-8. I removed an unrelated generated `uv.lock` change from this PR.

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [x] macOS ARM - targeted local test/lint
    - [ ] Linux x86
    - [ ] Linux ARM
      - Not manually tested on a non-UTF-8 Windows desktop; covered locally by a cp932 regression fixture.
      - GitHub Actions for the latest push are awaiting maintainer approval on the fork.
- Tested For:
    - [x] Basic functionality - targeted settings encoding regression tests
    - [ ] PyInstaller executable
      - Local: `uv run --extra pytest python -m pytest tests/qt/test_global_settings.py -q`
      - Local: `uv run --extra ruff ruff check src/tagstudio/qt/global_settings.py src/tagstudio/core/library/refresh.py src/tagstudio/qt/resource_manager.py src/tagstudio/core/library/alchemy/library.py tests/qt/test_global_settings.py`
      - Missing before marking ready again: latest GitHub Actions run and/or Windows non-UTF-8 codepage validation.


yosofbadr:
Closing this after a stricter audit. The issue is Windows/codepage-specific and the PR only has a synthetic local fixture, not real Windows validation. I do not want to leave maintainers with a PR I cannot confidently stand behind.

Sorry for the noise.

PR #1414, opened 2026-07-02
fix: make tagstudio python package executable

Computerdores:
### Summary

Originally proposed in #912, but the author closed it.
So I reimplemented it here since I don't know whether they want their changes to be used.
Allows using `python -m tagstudio` to launch tagstudio.

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [x] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/src/tagstudio/__main__.py
+++ b/src/tagstudio/__main__.py
@@ -1,0 +1,4 @@
+from .main import main
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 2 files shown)


src/tagstudio/__main__.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: (c) TagStudio Contributors
+# SPDX-License-Identifier: GPL-3.0-only
+
+
+from .main import main
```
xarvex:
```suggestion
from tagstudio.main import main
```
Sorry for the nitpick: any reason not to do the absolute import here? That follows how the imports are done elsewhere in the program.
Otherwise, this follows as done in other projects, thanks!

PR #1415, opened 2026-07-02
fix(nix): add pyright to devshell

Computerdores:
### Summary

If you don't have pyright installed on the system, the commit hook in the dev shell will fail.
This PR adds pyright to the devshell to fix that issue.

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [x] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -1,7 +1,6 @@
 # SPDX-FileCopyrightText: (c) TagStudio Contributors
 # SPDX-License-Identifier: GPL-3.0-only
 
-
 {
   lib,
   pkgs,
@@ -78,10 +77,10 @@ pkgs.mkShellNoCC {
     coreutils
     uv
 
+    pyright
     ruff
   ];
   buildInputs = [
-    pkgs.pyright
     python3Wrapped
   ]
   ++ (with pkgs; [
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -81,6 +81,7 @@ pkgs.mkShellNoCC {
     ruff
   ];
   buildInputs = [
+    pkgs.pyright
     python3Wrapped
   ]
   ++ (with pkgs; [


xarvex:
Just did a quick switch-a-roo to put this in `nativeBuildInputs` (see [this answer](https://discourse.nixos.org/t/use-buildinputs-or-nativebuildinputs-for-nix-shell/8464/2) for why). Thanks!

Computerdores:
> Just did a quick switch-a-roo to put this in `nativeBuildInputs` (see [this answer](https://discourse.nixos.org/t/use-buildinputs-or-nativebuildinputs-for-nix-shell/8464/2) for why). Thanks!

Could you elaborate on that? Because I always use `buildInputs` for all my devshells and have never had problems with that and the post seems to suggest that it just doesn't make a difference

xarvex:
Sorry, totally missed this when it came in!

So, my answer was *about* to be that outside of cross-compiling, there are other distinctions when it comes to `*Inputs` such as in regard to shell hooks. However, that means I missed the literal first line which states clearly that for `mkShell` it doesn't matter. Looking in the source for `mkShell`, that tracks, everything is outputted together: [nixpkgs/pkgs/build-support/mkshell/default.nix](<https://github.com/NixOS/nixpkgs/blob/2f87ae01e829895aa90254d07eb4a5ab431fe8f7/pkgs/build-support/mkshell/default.nix#L57-L59>)

This also reminded me of the existence of `inputsFrom` and `packages`. `packages` is the `mkShell`-specific attribute that I would wager was added in order to side-step the confusion of `*Inputs` in a shell, I would recommend using that for most cases.
`inputsFrom` is very special in that it lets you use the dependencies from a package, the reason we don't use it in this shell is because it doesn't include the transitive dependencies needed for Qt to work (and which we have to wrap the Python executable for, which that wouldn't do either). That, and what it *would* include is all the Python dependencies from nixpkgs, which in this case is undesirable. Otherwise, I would recommend it.

TL;DR: I was wrong, and I am switching the shell to use a single `packages` list.

PR #1421, opened 2026-07-03
fix: correctly display duration statistic in preview panel

### LINKED ISSUE #1037: [Bug]: File Duration Label Non-Functional
### Checklist

### TagStudio Version

Alpha 9.5.3

macOS 15.5

### Description

The file duration label (length in time for audio files and videos) in the preview panel appears to be non-functional as of 9.5.3 and shows a blank spot instead of the correct label (on first click only): 
Image
Image

Then when clicking the same file thumbnail again, a widely incorrect duration is now shown:
Image

Note: I caught this while preparing the release of 9.5.3 but wasn't able to get it functioning myself in time for the patch. I believe I've traced this back to #952, but the apparent cause and solution to this bug are not so straightforward. My *impression* is that this is caused by a potential race condition when loading a media file for the preview panel. This is indicated by the fact that clicking the file again properly loads the movie's length as something other than "0" internally, and how in my tests that moving the length check to a later part in the code started to produce positive values some of the time on the first try. The wildly high values on the other hand I haven't looked into too deeply yet.

As a minor knock-on effect, it appears that a duration value of "0" breaks the string formatting and just causes the final line to be blank. This, in comparison, is an easy fix that just requires that edge case to be fixed. I decided not to push this fix before the patch as to give "no information over false information" about the file duration.

### Expected Behavior

The correct duration of audio and video files should show in the preview panel, as they did in 9.5.2.

### Steps to Reproduce

1. Open a TagStudio library with audio and/or video files
2. Click on one of these file thumbnails to inspect it in the preview panel and see the empty space where the duration belongs

### Logs

N/A

ludvig-sandh:
### Summary

Fixes #1037, where previewing a video/audio didn't display the duration until selected again. Once selected again, duration was displayed, but a million times larger than expected.

The "million times larger" problem came from multiplying the duration in milliseconds by 1000. The fix was to divide by 1000 instead, to get seconds.

The problem of the duration not showing up was due to QMediaPlayer loading the duration asynchronously but the UI's trying to access it immediately. I fixed it by adding a signal that is triggered when the duration is available by QMediaPlayer. This signal causes the UI to update again via a callback, this time including the duration.

Some extra details:
- I return early in the callback if the selected preview file has been changed by the user. This avoids displaying stale preview information.
- The function updating the preview stats (in the UI) is called both immediately and upon the callback, making sure the immediately available data still shows up on screen even if QMediaPlayer would be slow/fail to load the duration.
- Renamed PreviewThumbView.__filepath to PreviewThumbView.__thumb_filepath since we now store both the path to the thumb and media file in that class, and we want to distinguish them. Just makes everything clearer.
- Previously this solution was used for formatting the duration:
```dur_str = str(timedelta(seconds=float(stats.duration)))[:-7]```
I think the idea was to strip out the microseconds (6 digits plus the decimal point). However, converting a timedelta to a string doesn't include microseconds. So this solution incorrectly pruned 7 characters from the valid duration string. I ended up replacing it with my own formatter for clarity.

### Tasks Completed

Evidence of correct duration stats:
image

Also tried quickly selecting different files and verified that always the expected duration showed up on screen.

- Platforms Tested:
    - [X] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [X] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [X] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/src/tagstudio/qt/mixed/file_attributes.py
+++ b/src/tagstudio/qt/mixed/file_attributes.py
@@ -7,7 +7,6 @@ import platform
 import typing
 from dataclasses import dataclass
 from datetime import datetime as dt
-from datetime import timedelta
 from pathlib import Path
 
 import structlog
@@ -37,6 +36,21 @@ class FileAttributeData:
     width: int | None = None
     height: int | None = None
     duration: int | None = None
+
+
+def _format_duration(duration: int | float) -> str:
+    """Format a duration in seconds as M:SS or H:MM:SS."""
+    try:
+        seconds = int(float(duration))
+        hours, seconds = divmod(seconds, 3600)
+        minutes, seconds = divmod(seconds, 60)
+        return (
+            f"{hours}:{minutes:02}:{seconds:02}"
+            if hours
+            else f"{minutes}:{seconds:02}"
+        )
+    except (OverflowError, ValueError):
+        return "-:--"
 
 
 class FileAttributes(QWidget):
@@ -224,15 +238,7 @@ class FileAttributes(QWidget):
 
             if stats.duration is not None:
                 stats_label_text = add_newline(stats_label_text)
-                try:
-                    dur_str = str(timedelta(seconds=float(stats.duration)))[:-7]
-                    if dur_str.startswith("0:"):
-                        dur_str = dur_str[2:]
-                    if dur_str.startswith("0"):
-                        dur_str = dur_str[1:]
-                except OverflowError:
-                    dur_str = "-:--"
-                stats_label_text += f"{dur_str}"
+                stats_label_text += _format_duration(stats.duration)
 
             if font_family:
                 stats_label_text = add_newline(stats_label_text)
--- a/src/tagstudio/qt/mixed/file_attributes.py
+++ b/src/tagstudio/qt/mixed/file_attributes.py
@@ -19,6 +19,7 @@ from tagstudio.core.enums import ShowFilepathOption
 from tagstudio.core.library.alchemy.library import Library
 from tagstudio.core.library.ignore import Ignore
 from tagstudio.core.media_types import MediaCategories
+from tagstudio.core.utils.str_formatting import format_duration
 from tagstudio.core.utils.types import unwrap
 from tagstudio.qt.models.palette import ColorType, UiColor, get_ui_color
 from tagstudio.qt.translations import Translations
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/tagstudio/qt/views/preview_thumb_view.py [hunk 1]
```diff
@@ -296,14 +311,15 @@ def _display_image(self, filepath: Path):
     def hide_preview(self) -> None:
         """Completely hide the file preview."""
         self.__switch_preview(None)
-        self.__filepath = None
+        self.__thumb_filepath = None
+        self.__media_filepath = None
```
CyanVoxel: Why split the filepath into two variables? Shouldn't these always be the same value?

ludvig-sandh: Oops, you are absolutely right. Mistook the existing filepath as being the path to the cached thumb, not the original media file. Will fix

ludvig-sandh: Fixed. And also solved a bug where the thumb was rerendered upon resize even when the preview was hidden.

src/tagstudio/qt/views/preview_panel_view.py
```diff
@@ -51,6 +51,8 @@ def __init__(self, library: Library, driver: "QtDriver") -> None:
         self._containers = FieldContainers(
             self.lib, driver
         )  # TODO: this should be name mangled, but is still needed on the controller side atm
+        self.__current_stats: FileAttributeData | None = None
+        self.__current_stats_filepath: Path | None = None
```
CyanVoxel: There doesn't appear to be a purpose to this `self.__current_stats_filepath` variable, is that correct?

src/tagstudio/qt/mixed/file_attributes.py
```diff
@@ -39,6 +38,17 @@ class FileAttributeData:
...
 
+def _format_duration(duration: int | float) -> str:
+    """Format a duration in seconds as M:SS or H:MM:SS."""
+    try:
+        seconds = int(float(duration))
+        hours, seconds = divmod(seconds, 3600)
+        minutes, seconds = divmod(seconds, 60)
+        return f"{hours}:{minutes:02}:{seconds:02}" if hours else f"{minutes}:{seconds:02}"
+    except (OverflowError, ValueError):
+        return "-:--"
```
CyanVoxel: This function doesn't have to be limited to this file - if it's going to be outside the `FileAttributes` class, it can be made a public function inside the `src/tagstudio/core/utils/str_formatting.py` file

src/tagstudio/qt/views/preview_thumb_view.py
```diff
@@ -34,11 +34,13 @@ class PreviewThumbView(QWidget):
     """The Preview Panel Widget."""
 
     check_ffmpeg = Signal(bool)
+    stats_updated = Signal(object, object)
```
CyanVoxel: These types should be made more specific than `object`, e.g. `Signal(Path, FileAttributeData)`

CyanVoxel (commented):
Approved when remaining comments are addressed 👍

src/tagstudio/qt/views/preview_panel_view.py
ludvig-sandh: Indeed

src/tagstudio/qt/views/preview_thumb_view.py
ludvig-sandh: Fixed this and the other comments.

CyanVoxel (approved):

PR #1423, opened 2026-07-04
fix: remove invalid child_id relationships from tag_parents

### LINKED ISSUE #1419: [Bug]: Tag parent relationship can remain after child is deleted
### Checklist

### TagStudio Version

Alpha v9.6.0/`main`

Windows 11

### Description

*(Reporting on behalf of someone else on Discord)* There's been a case of a tag parent relationship remaining in the database after a child tag in the relationship has been deleted. I can't currently reproduce the issue myself but have verified the extraneous relationship inside the DB. Note that this is the inverse issue that the DB 102 migration corrects for.

Quick info about the affected DB:
- Example of invalid relationship: parent_id 1038, child_id 1107
- Tag 1107 does not exist in the DB
- DB was created with version 102
- The invalid relationship existed **before** v9.6.0 in DB 102 (v9.5.5 or v9.5.6)
     - This means that the issue presented now is a result of this pre-existing invalid relationship going from dormant behavior to causing active issues due to the new structure of `search_tags()` introduced in v9.6.0

This issue presents as strange tag search behavior, which can range from results that don't update to no results at all.
Image
Image

Discord thread: https://discord.com/channels/1229183630228848661/1522240701675606077

Affected library: https://cdn.discordapp.com/attachments/1522240701675606077/1522598055667568791/ts_library.sqlite?ex=6a490d70&is=6a47bbf0&hm=74c161e11cd5109ef2abdae305d900628a109c4543a6372cd2ac7bb3ebf880c0&

While it should be trivial to patch any existing libraries that are affected by this, finding the root cause and determining if it's still an active issue is critically important.

### Expected Behavior

Tag parent relationships should be removed if either the parent or child tags referenced are deleted.

### Steps to Reproduce

*Currently unknown*

### Logs

Taken from my machine on `main`:
```
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/qt/controllers/search_panel_controller.py", line 126, in on_search_query_changed
    self.update_items(query)
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/qt/controllers/search_panel_controller.py", line 170, in update_items
    search_results: tuple[list[T], list[T]] = self.search_items(query_lower)
                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/qt/controllers/tag_search_panel_controller.py", line 145, in search_items
    return self.__lib.search_tags(name=query, limit=self._get_limit()[1])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/core/library/alchemy/library.py", line 1316, in search_tags
    descendant_tags = [hierarchy.pop(id) for id in all_ids]
                       ^^^^^^^^^^^^^^^^^
KeyError: 1107
```

CyanVoxel:
### Summary
This PR applies a migration patch similar to 102 that fixes existing library data where a child tag that is referenced in a `tag_parents` relationship has been deleted but the relationship has remained in the database. This should be enough to fix #1419. Bumps the DB version to **202**.

Since the `parent_id` side of deleting `tag_parent` relationships was corrected in #1107, it shouldn't be necessary to correct for invalid data in this direction as migration 102 did. Meanwhile #1250 ensures that no invalid data like this should be left behind going forward.

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [x] macOS ARM
    - [ ] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable


src/tagstudio/core/library/alchemy/library.py
```diff
@@ -917,6 +921,15 @@ def __apply_db201_migrations(self, session: Session):
 
             session.commit()
 
+    def __apply_db202_migration(self, session: Session):
+        """Migrate DB to DB_VERSION 202."""
+        with session:
+            all_tag_ids = session.scalars(text("SELECT DISTINCT id FROM tags")).all()
+            stmt = delete(TagParent).where(TagParent.child_id.not_in(all_tag_ids))
```
Computerdores: This could be combined into a single query which would make it much more efficient for large libraries, but this will be quite fast overall so this is more of a nitpick

Computerdores (approved):

src/tagstudio/core/library/alchemy/library.py
CyanVoxel: I went ahead and combined both this and the similar statement in the 102 migration into single statements

Computerdores: uhm you have to remove the `session.scalars` stuff, because that causes python to execute a separate query and then build a new sql query with all the results from the first one embedded, so instead of:
```
stmt = delete(TagParent).where(TagParent.child_id.not_in(session.scalars(select(Tag.id).distinct()).all()))
```
this should work (untested):
```
stmt = delete(TagParent).where(TagParent.child_id.not_in(select(Tag.id).distinct()))
```

CyanVoxel: Oh nice, didn't realize that could be done. Just tested and pushed it 👍

Computerdores (approved):

PR #1427, opened 2026-07-04
fix/refactor: use generic ModuleStatus class for getting status of optional modules

### LINKED ISSUE #1405: [Bug]: ripgrep inaccessible in macOS builds
### Checklist

### TagStudio Version

Alpha v9.6.0

macOS 15.7.7

### Description

In macOS builds, ripgrep is inaccessible due to GUI applications on macOS not having access to the system's PATH environment variable. This is the same issue we faced with FFmpeg until manually building a solution for it. I was already going to build in a parallel check system for ripgrep and other optional modules going forward, but in the meantime the addition of the ripgrep check in the UI added in #1398 has helped illuminate this silent issue.
Image

### Expected Behavior

macOS builds of TagStudio should be able to access any necessary library or module in the builds as it can when run from the terminal, including ripgrep. 

### Steps to Reproduce

1. Download the macOS build form the v9.6.0 release or build one yourself
2. Install ripgrep, probably through homebrew
3. Launch TagStudio via the desktop environment (double-click on the icon), go to "About TagStudio" in the menu and see the missing ripgrep warning
4. For comparison, launch TagStudio from the terminal (built version or developer environment) and see how ripgrep is found

### Logs

N/A

CyanVoxel:
### Summary
This PR introduces a new generic ModuleStatus class to get the locations and versions of FFmpeg, FFprobe, ripgrep, and any other future optional modules. I've also updated the "About" screen with the ripgrep version, and put these optional modules under their own section to better indicate that they're technically optional. Fixes #1405.

image

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [x] macOS ARM
    - [ ] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [x] PyInstaller executable


src/tagstudio/core/utils/ffmpeg_status.py
```diff
@@ -0,0 +1,69 @@
...
+    """Class for getting the location and version of FFmpeg, if it exists."""
+
+    @override
+    @classmethod
+    def which(cls):
+        if cls._cached_location:
+            return cls._cached_location
+
+        cls._cached_location = cls._which("ffmpeg")
+        return cls._cached_location
```
Computerdores: All of the implementations of `which` and `version` have the same caching logic, we should probably move that to the parent class.
Could alternatively be modeled as an abstract subclass of `ModuleStatus` (`CachedModuleStatus`?) to allow for modules where caching doesn't make sense for some reason (but I am not sure whether that would ever be the case).

Futhermore, the implementations of `which` and `version` each only differ in a single string, but are entirely duplicated, so using a factory pattern or an abstract class that only expects subclasses to define these parts (name of the command and args to get the version) would make sense

src/tagstudio/resources/translations/en.json
```diff
@@ -4,6 +4,7 @@
     "about.description": "TagStudio is a photo and file organization application with an underlying tag-based system that focuses on giving freedom and flexibility to the user. No proprietary programs or formats, no sea of sidecar files, and no complete upheaval of your filesystem structure.",
     "about.documentation": "Documentation",
     "about.module.found": "Found",
+    "about.modules.title": "Optional Modules",
```
Computerdores, line 7: You can add "Optionale Module" as the german version

src/tagstudio/qt/ts_qt.py
```diff
@@ -1755,3 +1761,32 @@ def select_to_entry(self, entry_id: int):
...
+            logger.warning("[QtDriver] FFprobe not found")
+
+        if RipgrepStatus.which():
+            logger.info(
+                "[QtDriver] ripgrep found",
+                which=RipgrepStatus.which(),
+                version=RipgrepStatus.version(),
+            )
+        else:
+            logger.warning("[QtDriver] ripgrep not found")
```
Computerdores: nit: could deduplicate by using a loop here

src/tagstudio/qt/previews/vendored/probe.py
```diff
@@ -0,0 +1,45 @@
...
+
+from tagstudio.core.utils.ffmpeg_status import FfprobeStatus
+from tagstudio.core.utils.silent_subprocess import (
+    silent_popen,  # pyright: ignore[reportUnknownVariableType]
+)
+
+logger = structlog.get_logger(__name__)
+
+
+def probe(filename: Path | str, timeout: int | None = None, **kwargs: ...):
```
Computerdores, line 22: nit: Since this is a library method and we are doing stuff to it anyway, maybe add a return type annotation?

src/tagstudio/core/utils/ffmpeg_status.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @override
+    @classmethod
+    def version(cls):
+        if cls._cached_version:
+            return cls._cached_version
+
+        ffmpeg_cmd = cls._which("ffmpeg")
```
Computerdores: Just call `cls.which()` here, that way we utilise the caching and avoid duplicating the "ffmpeg" string

src/tagstudio/core/utils/ffmpeg_status.py
CyanVoxel: > All of the implementations of which and version have the same caching logic, we should probably move that to the parent class.

It's fairly straightforward for me to move the location caching logic to the parent class inside the `_which()` method, but it's not clear to me how this would be possible with the version caching logic as there's no parent implementation to leverage
```py
class FfmpegStatus(ModuleStatus):
    @override
    @classmethod
    def which(cls):
        return cls._which("ffmpeg")
```
```py
class ModuleStatus:
    __cached_location: str | None = None
    _cached_version: str | None = None
    
    @classmethod
    def version(cls) -> str | None:
        raise NotImplementedError()

    @classmethod
    def _which(cls, cmd: str) -> str | None:
        """Internal method for determining the correct location for which().

        Args:
            cmd (str): The process command to search for.
        """
        if cls.__cached_location:
            return cls.__cached_location

        if platform.system() == "Darwin":
            for loc in _MACOS_BIN_LOCATIONS:
                full_command = which(loc + cmd)
                if full_command:
                    cls.__cached_location = full_command
                    return full_command

        cls.__cached_location = which(cmd)
        return cls.__cached_location
```

Otherwise I'm also able to consolidate the "FF" group's version logic into an intermediary class that also reduces the caching logic and serves as an alternative to a factory pattern:
```py
class _FfModuleStatus(ModuleStatus):
    """A base class that implements common logic for reading FFmpeg/FFprobe version output."""

    @classmethod
    def ff_version(cls, command: str):
        if cls._cached_version:
            return cls._cached_version

        ff_cmd = cls._which(command)
        if ff_cmd:
            out = silent_run([ff_cmd, "-version"], shell=False, capture_output=True, text=True)
            if out.returncode == 0:
                with contextlib.suppress(Exception):
                    cls._cached_version = str(out.stdout).split(" ")[2]
                    return cls._cached_version
```
```py
class FfmpegStatus(_FfModuleStatus):
    @override
    @classmethod
    def version(cls):
        return cls.ff_version("ffmpeg")
```

src/tagstudio/resources/translations/en.json
CyanVoxel: I'll add this on the Weblate side

src/tagstudio/qt/previews/vendored/probe.py
CyanVoxel: The return type is implicitly calculated as `Any | None` and can't be narrowed, and annotating that would require a new important and a pyright ignore comment anyway that also puts it way over the line length limit

Computerdores: good point

src/tagstudio/core/utils/ffmpeg_status.py
Computerdores: I was thinking of something like this:
```python
class ModuleStatus:
    _cached_version: str | None = None
    _cached_location: str | None = None

    @classmethod
    def version(cls) -> str | None:
        raise NotImplementedError()

    @classmethod
    def which(cls) -> str | None:
        raise NotImplementedError()

    @classmethod
    def _raw_which(cls, cmd: str) -> str | None:
        """Internal method for determining the correct location for which().

        Args:
            cmd (str): The process command to search for.
        """
        if platform.system() == "Darwin":
            for loc in _MACOS_BIN_LOCATIONS:
                full_command = which(loc + cmd)
                if full_command:
                    return full_command
        return which(cmd)

class CachedModuleStatus(ModuleStatus):
    @override
    @classmethod
    def version(cls) -> str | None:
        if cls._cached_version is None:
            self._cached_version = self._version()
        return self._cached_version

    @classmethod
    def _version(cls) -> str | None:
        raise NotImplementedError()
    
    @override
    @classmethod
    def which(cls) -> str | None:
        if cls._cached_location is None:
            self._cached_location = self._which()
        return self._cached_location
    
    @classmethod
    def _which(cls) -> str | None:
        raise NotImplementedError()

class SimpleCachedModuleStatus(CachedModuleStatus):
    _VERSION_ARGS: str | None = None
    _VERSION_IDX: int | None = None
    _CMD: str | None = None

    @override
    @classmethod
    def _version(cls) -> str | None:
        assert cls._VERSION_ARGS is not None
        assert cls._VERSION_IDX is not None
        cmd = cls.which()
        if cmd is not None:
            out = silent_run([cmd] + _VERSION_ARGS, shell=False, capture_output=True, text=True)
            if out.returncode == 0:
                with contextlib.suppress(Exception):
                    return str(out.stdout).split(" ")[cls._VERSION_IDX]

    @override
    @classmethod
    def which(cls):
        assert cls._CMD is not None
        return cls._which(cls._CMD)
```
(not sure if this is entirely correct, concrete names are just examples, etc.)

CyanVoxel: Okay cool, I think I see what you're getting at here with this - I don't see a need right now or going forward to provide a class that doesn't implement the caching, but version caching here should work well

CyanVoxel:
@Computerdores Implemented feedback and added a small new addition where tooltips show the module locations on the "About" screen

src/tagstudio/core/utils/ffmpeg_status.py
Computerdores: yeah same thought with the caching, figured I'd split it up for clarity

src/tagstudio/core/utils/ffmpeg_status.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class _FfModuleStatus(ModuleStatus):
+    """A base class that implements common logic for reading FFmpeg/FFprobe version output."""
+
+    _FFMPEG = "ffmpeg"
+    _FFPROBE = "ffprobe"
+
+    @classmethod
+    def ff_version(cls, command: str):
```
Computerdores, line 25: nit: I would set this private `_ff_version`

src/tagstudio/core/utils/module_status.py
```diff
@@ -0,0 +1,75 @@
...
+    __cached_location: str | None = None
+    __cached_version: str | None = None
+
+    @classmethod
+    def which(cls) -> str | None:
+        raise NotImplementedError()
+
+    @classmethod
+    def _cache_location(cls) -> str | None:
+        raise NotImplementedError()
```
Computerdores, line 45: unused?

Computerdores (approved):
Looking good besides two nits.
The new about window looks really good btw

PR #1429, opened 2026-07-05
feat: implement custom STL preview renderer from scratch

ludvig-sandh:
### Summary

This PR experiments with a custom renderer for STL file previews (#351).

STL files only contain triangle information. No shading, material or light, which makes them pretty easy to parse. STL files are pretty rigid and come in two main formats: binary and ASCII. I handle both, then parse the triangles, compute normals (for lighting), project onto 2d, then draw onto a PIL image.

Pros:
- Multiple STL files can be rendered concurrently (previous attempts had issues with this: (#1424 & #1231)
- No dependencies.
- Small binary footprint
- Should work cross-platform since I am just using regular python + numpy + PIL.

Cons:
- Slow. The rasterization step is sort of a bottleneck since we need to loop through all triangles and draw a polygon() on the image with each of them.

For now it is still an experiment (hence draft), so I have included some benchmarking code for tracking rendering times. My investigation shows roughly:
small files (0-2000 triangles): <80ms
medium files (2000-100,000): <1000ms
large files (100,000+): several seconds

For now I set a tri count cap of 100,000 and don't render anything above. Eventually the cap should be configurable.
It is also possible to approximate the rendering by reducing the triangle counts (only use a subset of them), but this didn't look good as the objects had visible holes in them.

This is what it looks like (two of the bottom files are not rendered due to the cap): 
image

From my very short and limited testing the renders looks good enough, and the rendering seems performant enough for users to benefit from the feature. Especially with the tri count cap, as well as concurrently rendering many files at a time. Obviously we may need some more rigorous testing to make sure it works well (eg. on low-end systems).

Although I haven't done any research, I can image most STL files being pretty small. Especially if you have many enough to use TagStudio to organize them. If that's true, a slower renderer that by default only renders smaller STL files might work well for now.

### Tasks Completed

- Platforms Tested:
    - [X] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [X] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [X] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/src/tagstudio/qt/previews/stl_renderer.py
+++ b/src/tagstudio/qt/previews/stl_renderer.py
@@ -1,0 +1,312 @@
+# SPDX-FileCopyrightText: (c) TagStudio Contributors
+# SPDX-License-Identifier: GPL-3.0-only
+
+from __future__ import annotations
+
+import math
+import re
+import struct
+import threading
+from pathlib import Path
+from time import perf_counter
+
+import numpy as np
+from PIL import Image, ImageDraw
+
+_BINARY_STL_HEADER_SIZE = 84
+_BINARY_STL_TRIANGLE_SIZE = 50
+_BINARY_STL_DTYPE = np.dtype(
+    [
+        ("normal", "<f4", (3,)),
+        ("vertices", "<f4", (3, 3)),
+        ("attribute_byte_count", "<u2"),
+    ]
+)
+_ASCII_FLOAT_PATTERN = rb"[-+]?(?:(?:\d+(?:\.\d*)?)|(?:\.\d+))(?:[eE][-+]?\d+)?"
+_ASCII_VERTEX_RE = re.compile(
+    rb"(?im)^\s*vertex\s+("
+    + _ASCII_FLOAT_PATTERN
+    + rb"\s+"
+    + _ASCII_FLOAT_PATTERN
+    + rb"\s+"
+    + _ASCII_FLOAT_PATTERN
+    + rb")"
+)
+_MODEL_PADDING = 0.86
+_MIN_TRIANGLE_AREA = 1e-12
+_BENCHMARK_STL_RENDERER = True
+_benchmark_print_lock = threading.Lock()
+
+
+class StlRenderError(ValueError):
+    """Raised when an STL file cannot be loaded or rendered."""
+
+
+def render_stl_thumbnail(
+    filepath: Path,
+    size: int,
+    bg_color: str,
+    max_file_size: int,
+    max_triangles: int,
+) -> Image.Image:
+    """Render an STL file to a square thumbnail image."""
+    file_size = filepath.stat().st_size
+    if file_size > max_file_size:
+        raise StlRenderError("STL file is too large")
+
+    start_time = perf_counter()
+    header = _read_stl_header(filepath)
+    read_time = perf_counter()
+    triangles, source_triangle_count, stl_kind = _load_stl_triangles(
+        filepath, header, file_size, max_triangles
+    )
+    load_time = perf_counter()
+    loaded_triangle_count = len(triangles)
+    triangles, normals = _prepare_triangles(triangles)
+    prepare_time = perf_counter()
+    if len(triangles) == 0:
+        raise StlRenderError("STL file contains no renderable triangles")
+
+    projected, depths, normals = _project_triangles(triangles, normals, size)
+    project_time = perf_counter()
+    image, drawn_triangle_count = _rasterize(projected, depths, normals, size, bg_color)
+    raster_time = perf_counter()
+
+    if _BENCHMARK_STL_RENDERER:
+        _print_benchmark(
+            filepath=filepath,
+            stl_kind=stl_kind,
+            file_size=file_size,
+            source_triangle_count=source_triangle_count,
+            loaded_triangle_count=loaded_triangle_count,
+            renderable_triangle_count=len(triangles),
+            drawn_triangle_count=drawn_triangle_count,
+            read_seconds=read_time - start_time,
+            load_seconds=load_time - read_time,
+            prepare_seconds=prepare_time - load_time,
+            project_seconds=project_time - prepare_time,
+            raster_seconds=raster_time - project_time,
+            total_seconds=raster_time - start_time,
+        )
+
+    return image
+
+
+def _read_stl_header(filepath: Path) -> bytes:
+    with filepath.open("rb") as file:
+        return file.read(_BINARY_STL_HEADER_SIZE)
+
+
+def _load_stl_triangles(
+    filepath: Path, header: bytes, file_size: int, max_triangles: int
+) -> tuple[np.ndarray, int, str]:
+    if len(header) < _BINARY_STL_HEADER_SIZE:
+        raise StlRenderError("STL file is too small")
+
+    triangle_count = struct.unpack_from("<I", header, 80)[0]
+    expected_size = _BINARY_STL_HEADER_SIZE + (triangle_count * _BINARY_STL_TRIANGLE_SIZE)
+
+    if expected_size == file_size:
+        if triangle_count > max_triangles:
+            raise StlRenderError("STL file contains too many triangles")
+        triangles = _load_binary_stl_triangles(filepath, triangle_count)
+        return triangles, triangle_count, "binary"
+
+    data = filepath.read_bytes()
+    trailing = data[expected_size:] if expected_size <= file_size else b""
+    if expected_size < file_size and not trailing.strip(b"\x00\r\n\t "):
+        if triangle_count > max_triangles:
+            raise StlRenderError("STL file contains too many triangles")
+        triangles = _load_binary_stl_triangles(filepath, triangle_count)
+        return triangles, triangle_count, "binary"
+
+    triangles, source_triangle_count = _load_ascii_stl_triangles(data, max_triangles)
+    return triangles, source_triangle_count, "ascii"
+
+
+def _load_binary_stl_triangles(filepath: Path, triangle_count: int) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


src/tagstudio/qt/previews/stl_renderer.py
```diff
@@ -0,0 +1,312 @@
...
+        return file.read(_BINARY_STL_HEADER_SIZE)
+
+
+def _load_stl_triangles(
+    filepath: Path, header: bytes, file_size: int, max_triangles: int
+) -> tuple[np.ndarray, int, str]:
+    if len(header) < _BINARY_STL_HEADER_SIZE:
+        raise StlRenderError("STL file is too small")
+
+    triangle_count = struct.unpack_from("<I", header, 80)[0]
```
Computerdores: move the `80` to a constant to document what it represents

src/tagstudio/qt/previews/stl_renderer.py
```diff
@@ -0,0 +1,312 @@
...
+            raster_seconds=raster_time - project_time,
+            total_seconds=raster_time - start_time,
+        )
+
+    return image
+
+
+def _read_stl_header(filepath: Path) -> bytes:
+    with filepath.open("rb") as file:
+        return file.read(_BINARY_STL_HEADER_SIZE)
```
Computerdores, line 90: this seems unnecessary to me

src/tagstudio/qt/previews/stl_renderer.py
```diff
@@ -0,0 +1,312 @@
...
+) -> tuple[np.ndarray, int, str]:
+    if len(header) < _BINARY_STL_HEADER_SIZE:
+        raise StlRenderError("STL file is too small")
+
+    triangle_count = struct.unpack_from("<I", header, 80)[0]
+    expected_size = _BINARY_STL_HEADER_SIZE + (triangle_count * _BINARY_STL_TRIANGLE_SIZE)
+
+    if expected_size == file_size:
+        if triangle_count > max_triangles:
+            raise StlRenderError("STL file contains too many triangles")
```
Computerdores: this is duplicated and seems like it should be right after the triangle count is retrieved

src/tagstudio/qt/previews/stl_renderer.py
```diff
@@ -0,0 +1,312 @@
...
+        triangles = _load_binary_stl_triangles(filepath, triangle_count)
+        return triangles, triangle_count, "binary"
+
+    data = filepath.read_bytes()
+    trailing = data[expected_size:] if expected_size <= file_size else b""
+    if expected_size < file_size and not trailing.strip(b"\x00\r\n\t "):
```
Computerdores: this way of detecting that it's an ascii variant seems scuffed

Computerdores (commented):
read over the loading code out of curiosity and wrote down some thoughts I had
Note that I didn't think about it particularly long, so take these for the not-thought-through thoughts that they are :)

ludvig-sandh:
@Computerdores thanks for the review! Good points :)

Well, do we want to move forward with this approach? There's no point in polishing experimental code until we aim to use it.

CyanVoxel:
Approach-wise this is the furthest anything has come by far, and is actually the first one to get some useful renders (and no crashes so far!). I think this is definitely worth refining as an approach, and I'm shocked that the performance seems decent (it's at least snappy with my 5-50 KB files).
image
There's definitely some visual quirks to iron out as well as what Computerdores has already mentioned, but I think this has my vote to continue on with. I'm having some fun playing with different angles and lighting in the meantime :)

Computerdores:
yeah my take is pretty much the same as Cyan's, only thing I am unsure about is the performance
80ms seems fairly good at first glance (and better then I would have expected), but these are also fairly small models and realistically in practice many if not most are going to be larger than 2k tris[^1] and especially during initial generation of the thumbnails we want the render times to be as small as possible
So imo this either needs significant optimisations (I've got no clue about numpy and co so maybe that is feasible?) or it would need to be written in faster (i.e. compiled) language which would have to be rust, because in the future the plan is to at least partially move to rust

[^1]: libraries likely either have no/very few 3D models or they have a ton because the person is using TS to manage their models. And in the latter case they are also likely to be larger so the effect would be amplified even more. (this is speculation of course, but I am fairly confident in my guess)

src/tagstudio/qt/previews/stl_renderer.py
ludvig-sandh: It is necessary to keep loading times small by avoiding expensive full file reads. In the common case for a correctly formatted binary STL file, the full file is never read into a python byte object, which is how loading times are kept below 1ms.

src/tagstudio/qt/previews/stl_renderer.py
ludvig-sandh: This is not the way we detect ascii. Rather it is a second chance for the binary format, in case it ends with null bytes or whitespaces after its regular content (it is not impossible for some exporters to add such things at the end of STL files). I clarified this with a comment, thx :)

src/tagstudio/qt/previews/stl_renderer.py
ludvig-sandh: Extracted the duplicate check now. And for the other part, the triangle count is extracted from the file where it's supposed to be *if* it is in binary format. In the binary case, that gives us the actual triangle count. In the ascii case, it is arbitrary text reinterpreted as uint32, so checking against max_triangles that early would falsely reject valid ascii files (where the noise value happens to exceed max_triangles). Not until we know for sure that the format is binary can we use the triangle count as a cutoff.

The way we know if it's binary is by checking triangle_count against the file size. For a given triangle count N, the file size should be 84 + 50N bytes. If this is the case, it means we read the format correctly. Otherwise we default to ascii parsing.

src/tagstudio/qt/previews/stl_renderer.py
ludvig-sandh: Fixed

ludvig-sandh:
I did three things:

1. Polish the code, with the review comments in mind. I think it's much clearer now.
2. Optimizations: (i) PIL .polygon() calls were expensive. Rasterizing ourselves was 3x faster, even in regular interpreted python. (ii) Removed regex parsing for ascii format to a custom solution. 1.5x faster ascii loading.
3. Experimented with STL rendering in c++ (never used Rust lol) to investigate an upper bound on the speedup achievable by using bindings with a statically typed language.

Results for ~100k triangle render (edit: that's a 5MB file if binary format):
| Approach  | Render time |
|-------|----------|
| PIL (original) | ~500ms |
| Custom rasterizer (current) | ~150ms |
| C++ (same alg as current) | ~5ms |

Two conclusions:
1. With the recent optimizations, I think what we have now is much better (I'm thinking of @Computerdores perf concerns).
2. C++ yielded another 30x on top of the optimizations I already made. So, if we wanted to use bindings with rust, we could expect another 10-30x speedup.

Thoughts?

src/tagstudio/qt/previews/stl_renderer.py
Computerdores: mb on the phrasing, what I meant was having it as a separate function and passing the result around

src/tagstudio/qt/previews/stl_renderer.py
Computerdores: I was referring more generally to doing it based on the file length, but I checked the blender source and they do it the same way. I guess STL parsing is just scuffed lol

src/tagstudio/qt/previews/stl_renderer.py
ludvig-sandh: Ah I see. Just thought it was clean. It's small but it has its very clear purpose. In general I like small functions bc they're easy to test. We can change it though. Curious, is there any downside to it that I'm not seeing perhaps?

CyanVoxel:
> I did three things:
>  [...]
>     3. Experimented with STL rendering in c++ (never used Rust lol) to investigate an upper bound on the speedup achievable by using bindings with a statically typed language.

> Two conclusions:
>     1. With the recent optimizations, I think what we have now is much better (I'm thinking of Computerdores perf concerns).
>     2. C++ yielded another 30x on top of the optimizations I already made. So, if we wanted to use bindings with rust, we could expect another 10-30x speedup.

> Thoughts?

So just to be clear, these C++ experiments referencing bindings aren't present in the most recent commits? Or is this referencing the changes made in 04b11e50f5bb3e62fe270a65a09c5defbd0ac369?

Regardless, I feel it's probably best to have something that works fine first, and then optimize from there in future PRs. On my machine at least, the render time of the STLs I have is extremely reasonable and is comparable to PDFs. I think otherwise the biggest concerns are the minor graphical glitches that occur with some of the faces, which appear to either render in the wrong order, with inverted normals, or otherwise strange stretched triangles. These models all appear normally when viewing in other programs. I have several examples of these:

***
Lastly, I apologize for the merge conflict created with the renaming of the `renderer.py` file to `file_renderer.py`, and a potential conflict with #1498 if that gets merged before this. The thumbnail system is one of the most frequently touched areas of the codebase and is receiving some long-overdue reworking, so it's been a little disruptive to open renderer PRs and I've tried to time it with a period of lower activity.

ludvig-sandh:
Correct about the c++ experiments not being present here. iirc I just wrote a standalone cpp script with the same logic, without any intention of actually ever using that code. Never actually wrote any bindings

I'm glad you think the version we have now is sufficient. I will look into the merge conflicts. They can't be that bad right (famous last words...)

Also I should be able to fix the visual artifacts, and then we should be good to go!

May be a bit busy right now since I just started my first real software engineering job and moved countries, but I'm passionate enough about this that I will find the time somehow.

PR #1430, opened 2026-07-05
feat: render `.pxd` thumbnails.

### LINKED ISSUE #1409: [Feature Request]: Thumbnail support for .pxd Documents
### Checklist

### Description

Add a thumbnail preview for .pxd, the file format used in Pixelmator Pro (Apple's Photoshop Alternative) 

### Solution

The file is in a .zip format and has a thumbnail under QuickLook/Preview.heic

Sample File:
[sample.pxd.txt](https://github.com/user-attachments/files/29579857/sample.pxd.txt)

### Alternatives

_No response_

Sola-ris:
### Summary
Add support for rendering `.pxd` thumbnails.
Fixes https://github.com/TagStudioDev/TagStudio/issues/1409.

As mentioned in the issue, `.pxd` files store their thumbnails in the `QuickLook` folder, similar to the other iWork files, they just use `.heic` images instead of `.jpg`.
`Preview.heic` is checked before `Thumbnail.heic` since its resolution is a bit higher (2000x2000 instead of 1024x1024).

#### Before
before

#### After
after

#### File used for testing
[sample.pxd.txt](https://github.com/user-attachments/files/29579857/sample.pxd.txt) from https://github.com/TagStudioDev/TagStudio/issues/1409.

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [x] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/src/tagstudio/qt/previews/renderer.py
+++ b/src/tagstudio/qt/previews/renderer.py
@@ -1218,8 +1218,8 @@ class ThumbRenderer(QObject):
         return im
 
     @staticmethod
-    def _iwork_thumb(filepath: Path) -> Image.Image | None:
-        """Render a thumbnail for an Apple iWork (Pages, Numbers, Keynote, Pixelmator) file.
+    def _apple_embedded_thumb(filepath: Path) -> Image.Image | None:
+        """Extract and render an apple embedded thumbnail (iWork, Apple Creative Studio).
 
         Args:
             filepath (Path): The path of the file.
@@ -1317,10 +1317,10 @@ class ThumbRenderer(QObject):
             page_size *= size / page_size.height()
         else:
             page_size *= size / page_size.width()
-        # Enlarge image for antialiasing
+        # Enlarge image for anti-aliasing
         scale_factor = 2.5
         page_size *= scale_factor
-        # Render image with no antialiasing for speed
+        # Render image with no anti-aliasing for speed
         render_options: QPdfDocumentRenderOptions = QPdfDocumentRenderOptions()
         render_options.setRenderFlags(
             QPdfDocumentRenderOptions.RenderFlag.TextAliased
@@ -1868,8 +1868,11 @@ class ThumbRenderer(QObject):
                 ):
                     image = self._open_doc_thumb(_filepath)
                 # Apple iWork Suite ============================================
-                elif MediaCategories.is_ext_in_category(ext, MediaCategories.IWORK_TYPES):
-                    image = self._iwork_thumb(_filepath)
+                elif (
+                    MediaCategories.is_ext_in_category(ext, MediaCategories.IWORK_TYPES)
+                    or ext == ".pxd"
+                ):
+                    image = self._apple_embedded_thumb(_filepath)
                 # Plain Text ===================================================
                 elif MediaCategories.is_ext_in_category(
                     ext, MediaCategories.PLAINTEXT_TYPES, mime_fallback=True
--- a/src/tagstudio/qt/previews/renderer.py
+++ b/src/tagstudio/qt/previews/renderer.py
@@ -623,7 +623,7 @@ class ThumbRenderer(QObject):
             image (Image.Image): The image to apply the edge to.
             edge (tuple[Image.Image, Image.Image]): The edge images to apply.
                 Item 0 is the inner highlight, and item 1 is the outer shadow.
-            faded (bool): Whether or not to apply a faded version of the edge.
+            faded (bool): Whether to apply a faded version of the edge.
                 Used for light themes.
         """
         opacity: float = 1.0 if not faded else 0.8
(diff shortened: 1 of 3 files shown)


src/tagstudio/qt/previews/renderer.py
```diff
@@ -1317,7 +1320,7 @@ def _pdf_thumb(filepath: Path, size: int) -> Image.Image | None:
         # Enlarge image for antialiasing
         scale_factor = 2.5
         page_size *= scale_factor
-        # Render image with no anti-aliasing for speed
+        # Render image with no antialiasing for speed
```
CyanVoxel: ? https://en.wikipedia.org/wiki/Anti-aliasing

src/tagstudio/qt/previews/renderer.py
```diff
@@ -1219,13 +1219,17 @@ def _image_vector_thumb(filepath: Path, size: int) -> Image.Image:
 
     @staticmethod
     def _iwork_thumb(filepath: Path) -> Image.Image | None:
-        """Extract and render a thumbnail for an Apple iWork (Pages, Numbers, Keynote) file.
```
CyanVoxel, line 1222: "Extract" is correct here since it's pulling an existing image from an archive, while changing it to simply "Render" is misleading as that gives the impression that the file is being processed and rendered for an image, similar to how the native applications in question would render them, which this is not doing

CyanVoxel (changes requested):
Functionally everything appears to work, though I have a couple nits for the documentation and a big nit for the categorization: Pixelmator is not a part of the [iWork](https://www.apple.com/iwork/index.html) suite. iWork has always consisted only of Pages, Keynote, and Numbers. Pixelmator Pro has recently become a part of the [Apple Creative Studio](https://en.wikipedia.org/wiki/Apple_Creator_Studio), which includes iWork apps along side their other professional grade software and miscellaneous apps like Freeform, but that does not retroactively make any of these part of the iWork suite.

I understand that the code for extracting an embedded thumbnail from a typical iWork file as well as a .pxd file is incredibly similar and seems to share many of the possible internal filenames, so I'd recommend renaming this class to something more generic like `_apple_creative_suite_thumb()` or `_apple_embedded_thumb()`. I know that typical [iLife](https://en.wikipedia.org/wiki/ILife) files don't have compatible structures with this, and can't speak for how the rest of the files in the creative suite operate, so it may be best to play it as generic as possible. For now I'd also just keep the `.pxd` extension in the `_DOCUMENT_SET`, I need to give these sets some attention and do an overall cleanup and reorganization.

Sola-ris:
> Pixelmator is not a part of the iWork suite. iWork has always consisted only of Pages, Keynote,
> and Numbers. Pixelmator Pro has recently become a part of the Apple Creative Studio, which
> includes
> iWork apps along side their other professional grade software and miscellaneous apps like
> Freeform, but that does not retroactively make any of these part of the iWork suite.

Got it, I just put it there since Apple lists it alongside Pages and friends on the Pixelmator
Website.

> I understand that the code for extracting an embedded thumbnail from a typical iWork file as well
> as a .pxd file is incredibly similar and seems to share many of the possible internal filenames,
> so
> I'd recommend renaming this class to something more generic like `_apple_creative_suite_thumb()`
> or
`_apple_embedded_thumb()`

Makes sense, I'll go with `_apple_embedded_thumb()` for now.

> For now I'd also just keep the `.pxd` extension in the `_DOCUMENT_SET`

Wouldn't that look something like this, grouping wise?

```python
elif (MediaCategories.is_ext_in_category(ext, MediaCategories.IWORK_TYPES)
      or MediaCategories.is_ext_in_category(ext, MediaCategories.DOCUMENT_TYPES)):
    image = self._apple_embedded_thumb(_filepath)
```

It doesn't seem right, to treat all documents as potential apple formats.
I could add a MediaType for Apple Creative Studio and use it like this

```python
elif (MediaCategories.is_ext_in_category(ext, MediaCategories.IWORK_TYPES)
      or MediaCategories.is_ext_in_category(ext, MediaCategories.APPLE_CREATIVE_STUDIO)):
    image = self._apple_embedded_thumb(_filepath)
```

src/tagstudio/qt/previews/renderer.py
Sola-ris: I should probably stop applying PyCharm's spell-check suggestions

src/tagstudio/qt/previews/renderer.py
Sola-ris: I changed that to keep the line length below 100 characters I'll fix it while switching over to `_apple_embedded_thumb()`

CyanVoxel:
For the check here you could either just use a check against the extension itself, similar to the PowerPoint one:
> ```python
> elif MediaCategories.is_ext_in_category(ext, MediaCategories.IWORK_TYPES) or ext == ".pdx":
>     image = self._apple_embedded_thumb(_filepath)
> ```
Since there's a growing dissonance between the media categories and what they're used for (e.g. searching for documents vs rendering document thumbnails) this'll be just fine for now. Eventually I'll rework the media types to better separate the groups depending on what they're used for

CyanVoxel (approved):

PR #1432, opened 2026-07-05
refactor: split out sql migrations from library

Computerdores:
### Summary

Currently creation of newly libraries, opening of existing libraries and migration of old libraries are all intertwined in the `Library.open_sqlite_library` method.
In order to make this code more maintainable, I want to split it such that:
- open library triggers migration if necessary and then opens normally either way
- creation of new libraries is handled separately
- the migration code
  - is outside of the `Library` class
  - doesn't use SQLAlchemy
  - is modular with a Migration class that has a list of all migrations (each migration would be a separate class with some metadata)

Since this will inevitably end up being a rather large refactor, I will endeavor to make commits that can be independently reviewed and be found to make sense.
This way the reviewer can review in smaller chunks.
In order to assist with this, I plan to add descriptions to the commit when I think it is necessary or helpful for understanding the intention behind the commit.
Also, I won't force-push on this PR at all, so reviewing the existing progress while the PR is still a draft and in progress shouldn't be a problem.

I tested each migration on this test library:  [ts_library.sqlite.zip](https://github.com/user-attachments/files/29713576/ts_library.sqlite.zip)

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [ ] Linux x86
    - [ ] Linux ARM
- Tested For:
    - [ ] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -587,9 +587,14 @@ class Library:
             )
 
         logger.info(f"[Library] Library DB version: {loaded_db_version}")
+        # TODO: this is very sketchy; blindly creating all tables the newest DB version should have
+        # without considering what version the DB is currently on and then doing all of the
+        # migrations after that seems like it could cause problems in some scenarios.
+        # instead only have this on creation and create new tables as part of migrations
         make_tables(self.engine)
 
         with Session(self.engine) as session:
+            # TODO ASSURANCE 1: version check + reordering
             # Ensure version rows are present
             with catch_warnings(record=True):
                 try:
@@ -605,6 +610,7 @@ class Library:
                 except IntegrityError:
                     session.rollback()
 
+            # TODO ASSURANCE 2: version check + reordering
             # check if folder matching current path exists already
             self.folder = session.scalar(select(Folder).where(Folder.path == library_dir))
             if not self.folder:
@@ -617,9 +623,6 @@ class Library:
                 session.commit()
                 self.folder = folder
 
-            # Apply any post-SQL migration patches.
-            assert loaded_db_version >= 6
-
             # save backup if patches will be applied
             if loaded_db_version < DB_VERSION:
                 self.library_dir = library_dir
@@ -627,6 +630,7 @@ class Library:
                 self.library_dir = None
 
             # migrate DB step by step from one version to the next
+            # TODO: every migration step should either complete sucessfully or be aborted fully
             if loaded_db_version < 7:
                 # changes: value_type, tags
                 self.__apply_db7_migration(session)
@@ -658,6 +662,7 @@ class Library:
                 # changes: tag_parents
                 self.__apply_db202_migration(session)
 
+            # TODO ASSURANCE 3: version check + reordering
             session.execute(
                 text("CREATE INDEX IF NOT EXISTS idx_tags_name_shorthand ON tags (name, shorthand)")
             )
@@ -672,6 +677,7 @@ class Library:
                 )
             )
 
+            # TODO: instead update DB version after every migration and abort on first fail
             # Update DB_VERSION
             if loaded_db_version < DB_VERSION:
                 logger.info(f"[Library] Library migrated to DB version {DB_VERSION}")
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -427,7 +427,9 @@ class Library:
             connection_string=connection_string,
             poolclass=poolclass,
         )
-        return create_engine(connection_string, poolclass=poolclass)
+        return create_engine(
+            connection_string, poolclass=poolclass, connect_args={"autocommit": False}
+        )
 
     def create_sqlite_library(
         self, library_dir: Path, in_memory: bool, sql_filename: str = SQL_FILENAME
@@ -615,6 +617,7 @@ class Library:
         """Migrate DB from DB_VERSION 6 to 7."""
         logger.info("[Library][Migration][7] Applying patches to DB_VERSION: 6 library...")
         # Repair tags that may have a disambiguation_id pointing towards a deleted tag.
+        # TODO: combine into single sql statement
         all_tag_ids = session.scalars(text("SELECT DISTINCT id FROM tags")).all()
         disam_stmt = (
             update(Tag)
@@ -682,7 +685,7 @@ class Library:
         logger.info("[Library][Migration][9] Added filename column to entries table")
 
         # Populate the new filename column.
-        for entry in self.all_entries():
+        for entry in self.__all_entries(session):
             session.merge(entry).filename = entry.path.name
         session.flush()
         logger.info("[Library][Migration][9] Populated filename column in entries table")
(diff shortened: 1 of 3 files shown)


src/tagstudio/core/library/alchemy/library.py
```diff
@@ -420,89 +426,170 @@ def open_sqlite_library(
...
-                            f"{mismatch_text}\n"
-                            f"{found_text} v{loaded_db_version}, "
-                            f"{expected_text} v{DB_VERSION}"
-                        ),
-                    )
+            # Add default tag color namespaces.
+            namespaces = default_color_groups.namespaces()
+
+            # TODO: do these try excepts even make sense? Shouldn't library creation just fail
+            # in these cases?
```
CyanVoxel: Yeah I think all these should be under a single try/except, any failure should abort the whole thing

Computerdores:
Ok, the spaghetti code has been largely untangled.
And from what I have found I only made one small semantics mistake that wasn't in the migration code and would have only affected the tests, so I am rather happy in that regard so far.

Computerdores:
@CyanVoxel I have now tested the full chain of migrations step by step and fixed two issues.
The one fixed in 81734c1373b3363289ded8d63bb7e81c71f53368 was actually already present before this PR...

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -596,337 +512,396 @@ def open_sqlite_library(
...
+                (self.__apply_db103_migration, 103, None),  # changes: tags
+                (self.__apply_db104_migration, 104, None),  # changes: deletes preferences
+                (self.__apply_db200_migration, 200, None),  # changes: field tables
+                (self.__apply_db201_migration, 201, 200),  # changes: field tables
+                (self.__apply_db202_migration, 202, None),  # changes: tag_parents
+            ]
+            for migration, v, iv in migrations:
+                if loaded_db_version < v and (iv is None or initial_db_version < iv):
+                    logger.info(f"[Library][Migration][{v}] Starting DB Migration")
+                    with session:
```
CyanVoxel: This is already inside an outer `with session`

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -596,337 +512,396 @@ def open_sqlite_library(
...
+            self.folder = session.scalar(select(Folder).where(Folder.path == library_dir))
+            if not self.folder:
+                folder = Folder(
+                    path=library_dir,
+                    uuid=str(uuid4()),
+                )
+                session.add(folder)
+                session.expunge(folder)
+                session.commit()
+                self.folder = folder
```
CyanVoxel, line 618: This folder stuff has never actually had a purpose and the PR that was going to finish implementing it was abandoned shortly after this came along with the migration to SQLite. Really this whole table should be removed in a future migration, but for now there's at least no need to enforce any logic or data since whatever it's doing is completely unused.

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -418,169 +421,82 @@ def open_sqlite_library(
...
-                            f"{found_text} v{loaded_db_version}, "
-                            f"{expected_text} v{DB_VERSION}"
-                        ),
-                    )
+            # Add default tag color namespaces.
+            namespaces = default_color_groups.namespaces()
 
-            logger.info(f"[Library] Library DB version: {loaded_db_version}")
-            make_tables(self.engine)
+            # TODO: are all of these commits necessary?
```
CyanVoxel, line 453: These commits can likely be replaced with flushes

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -596,337 +512,396 @@ def open_sqlite_library(
...
+            )
+
+        logger.info(f"[Library] Library DB version: {loaded_db_version}")
+        # TODO: this is very sketchy; blindly creating all tables the newest DB version should have
+        # without considering what version the DB is currently on and then doing all of the
+        # migrations after that seems like it could cause problems in some scenarios.
+        # instead only have this on creation and create new tables as part of migrations
+        # Note: this actually produces an error and fails to initialise built-in tags when opening
+        # a library that doesn't yet have the is_hidden property on the tags table
+        make_tables(self.engine)
```
CyanVoxel, line 565: Agreed

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -596,337 +512,396 @@ def open_sqlite_library(
...
+                session.expunge(folder)
+                session.commit()
+                self.folder = folder
 
         # everything is fine, set the library path
         self.library_dir = library_dir
         return LibraryStatus(success=True, library_path=library_dir)
 
-    def __apply_db7_migration(self, session: Session):
+    def __apply_db7_migration(self, session: Session, __library_dir__: Path):
```
CyanVoxel: Why surround the `library_dir` parameter in double underscores in the migration methods?

src/tagstudio/core/library/alchemy/library.py
Computerdores: fixed in ce64fb7773be77ec81299969cdc4be555d46be41

src/tagstudio/core/library/alchemy/library.py
Computerdores: totally forgot about that ^^
fixed in 4c3e0a3

src/tagstudio/core/library/alchemy/library.py
Computerdores: this avoids the highlighting as unused by pyright

src/tagstudio/core/library/alchemy/library.py
Computerdores: yeah good point, but I will keep that out of this PR (it's already big enough ^^)

Computerdores:
Considering how big this PR already is and that the current state of things seems quite good to me on its own, I think it would be a good idea to limit this PR to the current progress and continue the work in subsequent PRs.

Remaining todos for following PRs (mostly in that order):
- remove `Folder` stuff
- in `open_sqlite_library` replace `make_tables` with code that creates the tables in the appropriate migration
- split migrations out of opening procedure
  - split migrations out of library class
- make migrations sqlalchemy independent

src/tagstudio/core/library/alchemy/library.py
CyanVoxel: If that's the intention, then I'd prefer adding a `pyright: ignore` comment on these lines to be explicit about it:
```py
def __apply_db7_migration(self, session: Session, library_dir: Path):  # pyright: ignore[reportUnusedParameter]
```

Also when it comes to names with leading and trailing underscores, [PEP 8](https://peps.python.org/pep-0008/#descriptive-naming-styles) states to "Never invent such names; only use them as documented":
> `__double_leading_and_trailing_underscore__`: “magic” objects or attributes that live in user-controlled namespaces. E.g. `__init__`,` __import__` or `__file__`. Never invent such names; only use them as documented.

Computerdores: Hmm, this seems to be a setup problem on my end.
I'm using pyright as an LSP and it keeps reporting that the param is not accessed no matter what I add to ignore it (neither ignore comment nor global rule in pyproject.toml is respected), meanwhile running pyright normally doesn't even report that. No clue what's going on, but I have removed the dunder naming and will just have to ignore this issue on my end

src/tagstudio/core/library/alchemy/library.py
Computerdores: will be addressed in future PR

CyanVoxel (approved):
As far as I can tell this seems to be looking good so far, and throwing a bunch of old DB versions at it didn't seem to produce anything weird. I'll be keeping an eye on #1442 as it'll be very important to get the rest of this refactor done as soon as it can be.

Thank you so much for your work on all this so far! It's real nice seeing the migrations being cleaned up, especially after so many have started to pile up.

PR #1433, opened 2026-07-06
fix: Thumbnail Grid not resizing to fill width

### LINKED ISSUE #1431: [Bug]: Thumbnail Grid Doesn't Update Properly After Search
### Checklist

### TagStudio Version

TagStudio v9.6.0/`main`

macOS 15.7.7

### Description

After initially loading a library, performing a search or opening a new library will cause an issue with the thumbnail grid where it will no longer expand to fill the available window space. It can be shrunk when resizing the window smaller than the existing icon space, but resizing it larger will only let the icons take up a limited amount of space. Reapplying the search will update the icons to fill the space at first, but the issue reappears when attempting to resize again. It also appears that the scrollbar is appearing and disappearing based on where it expects the icons to normally be, not where they actually are on the screen.
Image

### Expected Behavior

The thumbnail grid should always reorganize to fit the available space, and the scrollbar should not disappear when items are cutoff on the screen. 

### Steps to Reproduce

1. Open a TagStudio library with a decent amount of files (~20+)
2. Resize the window larger and smaller, observing normal behavior
3. Enter a search that also produces a decent amount of items or load a different library
4. Repeat step 2, though this time notice how the thumbnails do not resize properly

### Logs

Occasionally I'm able to get this error from the logs:
```
IndexError: list index out of range
Error calling Python override of QLayout::setGeometry(): Traceback (most recent call last):
  File "/Users/cyanvoxel/Local/Dev/TagStudio/src/tagstudio/qt/thumb_grid_layout.py", line 218, in setGeometry
    self.visible_changed.emit(self._entry_ids[start])
                              ~~~~~~~~~~~~~~~^^^^^^^
IndexError: list index out of range
```

TheBobBobs:
### Summary
Fix #1431 
`ThumbGridLayout` was only updating thumbnails if the visible entries changed. This would cause the grid to not update on pages where despite the row width changing the visible entries were still the same.

### Tasks Completed
- Platforms Tested:
    - [X] Linux x86
- Tested For:
    - [X] Basic functionality

### DIFF
--- a/src/tagstudio/qt/thumb_grid_layout.py
+++ b/src/tagstudio/qt/thumb_grid_layout.py
@@ -48,8 +48,8 @@ class ThumbGridLayout(QLayout):
         self._renderer.updated.connect(self._on_rendered)
         self._render_cutoff: float = 0.0
 
-        # _entry_ids[StartIndex:EndIndex]
-        self._last_page_update: tuple[int, int] | None = None
+        # _entry_ids[StartIndex:EndIndex], per_row
+        self._last_page_update: tuple[int, int, int] | None = None
 
         self._scroll_to: int | None = None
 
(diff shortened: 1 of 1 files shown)


CyanVoxel (approved):
Seems to take care of the issue! Thank you so much for the quick fix!

Issue #1282, opened 2025-12-31
[Feature Request]: Make Github releases Immutable


kitsumed:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

It's a recent safety feature offered by Github to reduce the attack-surface on the supply-chain.
In simpliers terms:
Once Immutable releases are enabled, every new releases and their associated tag will be marked as "Immutable". Assets uploaded in a immutable release cannot be replaced or removed, thus ensuring that once a release is made, the files cannot be altered. You will still be able to edit the actual release text or to delete the whole release, but not modifying the released assets. Draft of releases are not immutable.

You can learn more here : https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/immutable-releases

### Solution

Immutable releases is a checkbox you can find directly by going into the Settings section of this repo.

Issue #1030, opened 2025-08-12
[Feature Request]: Nightly Builds


Computerdores:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

Since TS is still Alpha software there is often a situation where the last release has a bug that is critical for some users, but the git version already has a fix. In these situation I often tell users to try out the git version if they know how to, but obviously not everyone does.

### Solution

I suggest we setup a ci pipeline to create nightly builds, meaning a build of the current git version every day (at the time where merges are least likely). I believe Github allows for marking releases as "experimental" or some such

xarvex:
I already have some workflow revamping that I've been preparing and testing for use here, I will for sure toss in a mechanism for nightly builds. Currently going off other projects that do nightly like Neovim and WezTerm for inspiration.

xarvex:
Seems like I am approaching a year late on this, but wanted to say that the workflows are getting some much-needed TLC right now and that will include nightlies as mentioned here (though will probably come a bit later). So, expect this soon™.

Issue #892, opened 2025-03-29
[CI] Nix Package Build Checks


xarvex:
It has already been on my mind, but #890 spurred some more thought surrounding some kind of workflow to ensure the Nix package can build and functions. One thing that needs to happen is that `pythonRemoveDeps = true` should be removed and replaced with something like the following:
```nix
# ...
{
  # ...
  pythonRemoveDeps = lib.optional (!withJXLSupport) [ "pillow_jxl" ];
  pythonRelaxDeps = [
    # packages here
  ];
  # ...
}
```

Setting `pythonRemoveDeps` the way I did was a mistake as it completely removed dependency checking taking place at all. Here, it can be set to only ignore the Pillow JXL plugin if support is disabled.

I take care of this as part of #886.

One concern about having a workflow is that authors might become confused about this strange failure when working on a contribution. My suggestion is that the workflow should only run for commits and PRs to this “origin” repo. On failure with a PR, a short comment that tags me can be made by the workflow ([peter-evans/create-or-update-comment](https://github.com/peter-evans/create-or-update-comment)), that way things can be taken care of when pulling is ready.

This is something I am working on and tinkering with, this issue only exists for tracking... and for reminding me.

PR #1437, opened 2026-07-07
feat(ci)!: complete workflows revamp

### LINKED ISSUE #1406: [Bug][CI]: macOS x86 build is replaced with Apple Silicon version
### Checklist

### TagStudio Version

9.6.0

MacOS (intel)

### Description

this is a pretty easily solvable issue that apparently in this latest update who ever was making the compiles forgot to double check the x86 zip acculy contained an x86 compatible version as whats in it is identical to the x64 (not compatible) version

it is abit minor though as Mac(intel) is a pretty niche use case at this point

Image

### Expected Behavior

it should just contain the compatible x86 executable

i do wonder if  compiling a "universal" version is in the cards as apple calls it but if its separated already i assume it has its difficulties and probably not worth it now (intel obsolete now)

### Steps to Reproduce

just open  the Mac x86 zip and check the file if at least running on mac it should read it as (Apple silicon) (x64) witch is  not what it should contain

### Logs

_No response_

xarvex:
### Summary

Implements: #1030 (partially, doesn't create a GitHub Release yet)
Fixes: #1406 (macos-15-intel has to be used now)

#### Notable Changes

- Some external actions have been removed to prefer local tools.
- setup-uv has been added to benefit from uv's cache and lightweight installation.
- Single local action for all Python setups, which can take a 'group' argument for extra tools to be installed. This removes the duplicated (and sometimes diverging) setup processes.
- Multi-platform jobs are no longer split up into multiple jobs, but rather take a matrix of runners, with each step checking the OS if relevant. This, as well, avoids duplication of work.
- Steps have been made parallelized thanks to a [recent GitHub change](https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel).
- Nightly builds are now made, albeit without releases. The artifacts will, for now, be available on the action artifacts page.

#### Future Work

The flow of releases still needs to be expanded on, the idea being that the action can be the entry point for creating a new release, be it Nightly, Pre-Release, or full Release. Nightlies will need some concept of cleanup to ensure there won't be many of those piling up. While working on this, #1282 is being taken into consideration.

As well, more checks might want to be added, such as for #892.

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [x] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable


sapioit:
Thank you!

PR #1451, opened 2026-07-18
feat(ui/ux): replace "Add Tag/Field" modals with search+create bar(s) with autocomplete

### LINKED ISSUE #1379: [Feature Request][UX]: Embed "Add Tag" and "Add Field" Modals in Preview Panel
### Checklist

### Description

This is a proposed UX improvement for adding tags and fields quicker to entries. Currently, to add a tag to an entry from the UI, the user must perform the following clicks and key presses:
1. Click "Add Tag" button or use <kbd>Ctrl</kbd> + <kbd>T</kbd>
2. Start typing to search for tag
3. Hit <kbd>Enter</kbd> if tag exists
3a. If tag does not exist, hit <kbd>Enter</kbd> to open tag creation modal
3b. The user could either fill out detailed info now, or hit <kbd>Enter</kbd> again to create and close the "edit tag" modal
4. Hit <kbd>Enter</kbd>/<kbd>Esc</kbd> to close the "add tag" modal

For users who are frequently creating new tags (especially in new libraries), every mouse click and keystroke matters. That's why I'm proposing to ditch the modal approach and prioritize moving the user quickly between searching/creating tags to getting those tags added to searching/creating again. 

### Solution

What I have in mind would work something like this (using tags as an example):
Image
*When adding a tag, the entry data section of the preview panel is now turned into a search preview, similar to how the modal search currently works. The search/input bar for tags is now at the bottom, and the "Add Tag" and "Add Field" buttons are replaced with a single "Cancel" button that reverts the preview panel back to normal (<kbd>Esc</kbd> will also revert the box).*

1. Click "Add Tag" button or use <kbd>Ctrl</kbd> + <kbd>T</kbd>
2. Start typing to search for tag (same as modal).
3. Hit <kbd>Enter</kbd> if tag exists to add it and revert to the preview panel view. Optionally, tags that are already added to the entry can display greyed out.
3a. If tag does not exist, hit <kbd>Enter</kbd> to "Create & Add", which will now *skip* the edit modal and simply make a blank tag with text the user had input and add it to the entry. The preview panel is now reverted back to its normal state, perhaps with the "Add Tag" button highlighted in order to allow for an <kbd>Enter</kbd> press to enter them back into the search. Otherwise, if the user wishes to still edit the tag further at the time of creation, there is a new option before "Create & Add" called "Create & Edit" which opens up the "edit tag" modal, just like the current behavior.

This should allow users to still add existing tags quickly while being able to more quickly *create* tags that don't already exist without having to have the "edit tags" modal fly up in their face if they don't want it. This method should also work for adding fields, building off the work of #1374.

### Alternatives

One aspect of this proposal that could be a hindrance is that previously when adding multiple tags that exist in a row, after hitting <kbd>Enter</kbd> you were then put back into the search bar to start typing for another tag right away. Now, after adding a tag that exists you'll have to hit <kbd>Enter</kbd> again in order to bring the search bar back up.

It's possible that a better behavior would be to keep the search/input bar up after successfully adding or creating a tag, and then requiring the user to hit <kbd>Esc</kbd> or "Cancel" in order to restore the preview panel. It's also possible that this is very subjective and should be an option for the user to set themselves.

In any case, we should know more about this once the feature is developed and we can test run it.

CyanVoxel:
### Summary
This PR changes the functionality of the "Add Tag" and "Add Field" buttons, removing the search modal windows and replacing them with an inline tag search and creation bar. This bar is designed to speed up searching for tags/field templates to add to entries and greatly reduce friction with the tag creation workflow, especially for new libraries. This is intended to cover and enhance *most* tagging workflow needs, however the option for a persistent dockable search list modal is still planned for the future in order to cover additional workflows that aren't covered by an autocomplete search bar.

image

image

image

image

### Full Features
- Click "Add Tag" or press <kbd>Ctrl</kbd>+<kbd>T</kbd> to open the tag search/create bar.
- Type to search for the name of an existing tag, or use that search to create and add a new tag right there.
    - Pressing <kbd>Enter</kbd> with **no results** will **create** a tag from your search query and add it to the selected entries.
        - In the options, or right-click context menu on the search bar, you can toggle the option to open the "Edit Tag" modal after creating new tags if you prefer that extra step.
    - Pressing <kbd>Enter</kbd> with **one or more results** will **add** the first tag (that's not already on your entries) to your selected entries.
    - Holding <kbd>Shift</kbd> while pressing <kbd>Enter</kbd> will force create a tag from your search query, regardless of the search results. This is also indicated by the first tag in the results becoming translucent.
- After adding a tag, the "Add Tag" button is now focused, allowing you to press <kbd>Enter</kbd> or <kbd>Space</kbd> to quickly reactivate the search bar.
- Pressing <kbd>Esc</kbd>, pressing <kbd>Enter</kbd> with nothing typed, or clicking off of the search bar will exit the search and return the "Add Tag" and "Add Field" buttons.
- All of the above is also true for Field Templates, where the hotkey is <kbd>Ctrl</kbd>+<kbd>L</kbd> to open its respective search bar.
- New Field templates are automatically edited after being created, since they have a type that must be chosen before they can be used. This is separate from the option to automatically edit a field's contents once a template has been applied to an entry, which is on by default.

### Project Layout
**Full disclosure**, this PR is also trojan horsing in some MVC-related changes that I've been honing in on while developing this. As someone who's been knee-deep in all flavors of the newer UI paradigms we've been using in the project, I've gotten a real good feel for what's been working and *what hasn't*. Many of the new UI files here ended up spinning off from the search modal/panel classes, which became so heavily modified while I was developing an embedded search that I just split my changes off into new classes. I'd eventually like to go back and retroactively apply the paradigms introduced here into the older classes (and try to reduce code duplication where possible).

The gist of these changes comes down to me using a more traditional MVC setup and tailoring it to Qt's needs, using **views as layouts** and **controllers as final widgets**. This solves every problem I've had with the existing MVC guidelines and implementations while meeting the goals set out for the MVC refactor in the first place, *all while* fitting in cleanly with Qt's widget system. I'm also starting to make some other opinionated changes after working with this part of the codebase so much, like dropping `controller` from controller filenames and using `_single` underscore names for private variables instead of using both `_single` and `__double` names that were trying to mimic a semblance of private/protected variables in a language that just unfortunately doesn't have that functionality (name mangling doesn't count). In my experience working with the areas of the codebase that had accrued these changes, it's just become way more cumbersome than helpful. 

#### This PR
- [x] ~~Do the same for "Add Field", or at least make it functional in the meantime~~
- [x] ~~Fix issue with removing tags not updating the autocomplete while its open~~
- [x] ~~Update documentation~~ (8722c5f83e2b036db1ff527ddd72dd099ab3cf90)
- [x] ~~Fix tag appearances where they used to be wide~~
#### Likely Future PRs
- [ ] Refactor search panels to use same MVC paradigm as this, and reuse code where possible
- [ ] Allow for tabbing between different autocomplete suggestions
    - [x] ~~with an indicator of which one will be added~~
- [ ] Remove Edit menu "Add Tag to Selected" option (once persistent search panel feature is available)

Closes #1379. Progress towards #1460.

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [x] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [ ] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable


Computerdores:
Ok, I've taken a look at the code for the new suggest box to understand the idea behind the MVC approach you have gone for here.
I'll just summarize what I think that idea is and I would ask that you tell me whether I understood it correctly and whether I missing anything important:
- every custom widget class is itself the controller
- the view is the layout used by the widget/controller class
- the controller exposes public methods for the users of the widget to call (in the case of our preview widget, that would be something like `show_from_path(path: Path)`)
- the view/layout exposes public methods for the controller to use to control it (in the same example, that would be something like `hide_video_preview`, `show_video_preview`, `video_preview` (property), and the same for audio and image preview)
- the public controller methods then orchestrate the view to achieve their purpose (in the same example, `show_from_path` would identify what type of file it is, then hide the wrong previews, show the correct one, and use e.g. `audio_preview` to get the preview and tell it to show the audio file at the path)
- the view takes care of styling, arranging, and creating the widgets
- generally, if it isn't meant to be called from outside the class it should be prefixed with `_`

Assuming that this is accurate, this could also serve as a starting point for the updated MVC docs

CyanVoxel:
> Ok, I've taken a look at the code for the new suggest box to understand the idea behind the MVC approach you have gone for here. I'll just summarize what I think that idea is and I would ask that you tell me whether I understood it correctly and whether I missing anything important:
> [...]
> Assuming that this is accurate, this could also serve as a starting point for the updated MVC docs

That's pretty much correct, with the exception that the views/layouts should have **little to no logic in them**, acting purely for arranging and styling the interface elements.  Any logic for modifying the view should be on a controller, even if it's a base controller that other controllers are expected to inherit from. There *might* be exceptions to this if a method *really really* makes sense to go in a layout, but as a rule of thumb I'm trying to separate all logic from them to have a stricter separation that makes the codebase more flexible *(see 9780d68d2cf4f1820c818a75592b4f4cfd8bb3e5 on a future branch where I'm able to axe the `TagSearchPanelView` and `FieldTemplateSearchPanelView` classes, and likewise did not need equivelents to those here for the `SuggestBox`).*

Here `SuggestBoxView` is a great example of that, just a single `__init__()` method that constructs the layout and applies styles. The base controller class `SuggestBox` then sets a `SuggestBoxView` as its layout and establishes the methods for modifying and updating the layout - this hard separation also allows for other base controller classes to come in and work with the view differently, if ever needed. Classes that extend from `SuggestBox` like `TagSuggestBox` and `FieldSuggestBox` then can take advantage of the base logic for the view while adding or overriding their own, where needed.

More examples of this can be found in 9780d68d2cf4f1820c818a75592b4f4cfd8bb3e5, where I do that same hard logic splitting in the `SearchPanel` classes and end up cutting 4 entire classes all while further decoupling the views from the widgets/controllers.

***

I can get some updated docs on all this soon. I've been working ahead with some house-of-cards branches and really putting this paradigm to the test across a lot more of the UI codebase and so far it's been going *(unbelievably)* smoothly, so I'm feeling very confident about the whole approach.

Computerdores:
> Here `SuggestBoxView` is a great example of that, just a single `__init__()` method that constructs the layout and applies styles. The base controller class `SuggestBox` then sets a `SuggestBoxView` as its layout and establishes the methods for modifying and updating the layout - this hard separation also allows for other base controller classes to come in and work with the view differently, if ever needed. Classes that extend from `SuggestBox` like `TagSuggestBox` and `FieldSuggestBox` then can take advantage of the base logic for the view while adding or overriding their own, where needed.

Ah, so the goal is to make it possible to switch out the view while keeping the controller and only methods that aid that should be added to the view?

CyanVoxel:
> Ah, so the goal is to make it possible to switch out the view while keeping the controller and only methods that aid that should be added to the view?

Exactly - instead of something like `TagSuggestBox` being limited to a `SuggestBoxView` just because `SuggestBox` uses it *(i.e. `SuggestBox(SuggestBoxView)` -> `TagSuggestBox(SuggestBox)`)*, it may benefit from a more specific `TagSuggestBoxView(SuggestBoxView)` that contains some additional elements on top of a `SuggestBoxView`. All it would have to do then is override the`__init__()` and then implement any specific logic on top of what is already implemented by `SuggestBox`. ~~I've also got `view` parameters in some constructors at the moment, but I'm looking more into changing those into~~ just raw parameters required for the views and/or setters on the controller that can update the view's widgets, like setting text. I think that would be cleaner and would remain consistent with simpler widgets that can already just create a new view internally inside their `__init__()`s. 

The modularity of these pieces allows them to be incredibly reusable in nature while also being very easy to extend from, and in my experience very easy and unambiguous to follow while developing.

***

Also, a good example for methods that might be appropriate inside views could be ones that setup the appearance modularly, so controllers can selectively control how the view is set up. Not logic or abstract methods dictating how the view should be *used*, but more so just hooks to enable, disable, or change specific parts of how the view is constructed or updated. I don't have any examples of this in the code at the moment, but I imagine I'll start using that sort of modular approach to setting up certain views going forward.

ARBlueSlime:
Please add an option to switch back to the modals, the new bar forces the user into seeing only 3 to 4 tags at a time in a crammed space (wouldn't have been as bad if there was a way to stretch it upwards)

TrigamDev:
> Please add an option to switch back to the modals, the new bar forces the user into seeing only 3 to 4 tags at a time in a crammed space (wouldn't have been as bad if there was a way to stretch it upwards)

The modals can still be used, either by using `Ctrl` + `Shift` + `T`, or by using the "Add Tag to Selected" action.

ARBlueSlime:
> > Please add an option to switch back to the modals, the new bar forces the user into seeing only 3 to 4 tags at a time in a crammed space (wouldn't have been as bad if there was a way to stretch it upwards)
> 
> The modals can still be used, either by using `Ctrl` + `Shift` + `T`, or by using the "Add Tag to Selected" action.

Oh okay, thank you!

Issue #1442, opened 2026-07-10
[Task]: Finish migration cleanup


Computerdores:
Now that #1432 is merged, a couple of things remain to be done on the migrations:
1. [x] remove `Folder` stuff
2. [x] in `open_sqlite_library` replace `make_tables` with code that creates the tables in the appropriate migrations
3. [x] split migrations out of opening function (preparation for 4. and the subpoint)
    - [x] then split migrations out of library class
4. [ ] make migrations sqlalchemy independent

PR #1456, opened 2026-07-23
refactor: almost final migrations cleanup

Computerdores:
### Summary
Progress on #1442.
Aims to complete the migrations cleanup far enough that only independence from sqlalchemy remains.

### Tasks Completed
- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [ ] Linux x86
    - [ ] Linux ARM
- Tested For:
    - [ ] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -528,10 +528,7 @@ class Library:
     ) -> LibraryStatus:
         self.engine = self.__get_engine(library_dir, in_memory, sql_filename)
 
-        logger.info(
-            "[Library] Opening SQLite Library",
-            library_dir=library_dir,
-        )
+        logger.info("[Library] Opening SQLite Library", library_dir=library_dir)
 
         # Don't check DB version when creating new library
         loaded_db_version = self.get_version(DB_VERSION_CURRENT_KEY)
@@ -562,9 +559,7 @@ class Library:
 
         # save backup if patches will be applied
         if loaded_db_version < DB_VERSION:
-            self.library_dir = library_dir
-            self.save_library_backup_to_disk()
-            self.library_dir = None
+            Library.save_library_backup_to_disk(library_dir)
 
         # migrate DB step by step from one version to the next
         # (migration_method, db_version, initial_db_version)
@@ -1859,16 +1854,17 @@ class Library:
                 session.rollback()
                 return None
 
-    def save_library_backup_to_disk(self) -> Path:
-        assert isinstance(self.library_dir, Path)
-        makedirs(str(self.library_dir / TS_FOLDER_NAME / BACKUP_FOLDER_NAME), exist_ok=True)
+    @staticmethod
+    def save_library_backup_to_disk(library_dir: Path) -> Path:
+        assert isinstance(library_dir, Path)
+        makedirs(str(library_dir / TS_FOLDER_NAME / BACKUP_FOLDER_NAME), exist_ok=True)
 
         filename = f"ts_library_backup_{datetime.now(UTC).strftime('%Y_%m_%d_%H%M%S')}.sqlite"
 
-        target_path = self.library_dir / TS_FOLDER_NAME / BACKUP_FOLDER_NAME / filename
+        target_path = library_dir / TS_FOLDER_NAME / BACKUP_FOLDER_NAME / filename
 
         shutil.copy2(
-            self.library_dir / TS_FOLDER_NAME / SQL_FILENAME,
+            library_dir / TS_FOLDER_NAME / SQL_FILENAME,
             target_path,
         )
 
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -14,6 +14,7 @@ from os import makedirs
 from pathlib import Path
 from typing import TYPE_CHECKING
 
+import sqlalchemy
 import structlog
 from humanfriendly import format_timespan  # pyright: ignore[reportUnknownVariableType]
 from sqlalchemy import (
@@ -1745,6 +1746,49 @@ class Library:
                 child_id=tag.id,
             )
             session.add(parent_tag)
+
+    def get_version(self, key: str) -> int:
+        """Get a version value from the DB.
+
+        Args:
+            key(str): The key for the name of the version type to set.
+        """
+        return Library._get_version(self.engine, key)
+
+    @staticmethod
+    def _get_version(engine, key: str) -> int:
+        with Session(engine) as session:
+            engine = sqlalchemy.inspect(engine)
+            try:
+                # "Version" table added in DB_VERSION 101
+                if engine and engine.has_table("versions"):
+                    version = session.scalar(select(Version).where(Version.key == key))
+                    assert version
+                    return version.value
+                # NOTE: The "Preferences" table has been depreciated as of TagStudio 9.5.4
+                # and is set to be removed in a future release.
+                else:
+                    return int(
+                        unwrap(
+                            session.scalar(
+                                text("SELECT value FROM preferences WHERE key == 'DB_VERSION'")
+                            )
+                        )
+                    )
+            except Exception:
+                return 0
+
+    @staticmethod
+    def _set_version(session: Session, key: str, value: int) -> None:
+        """Set a version value to the DB.
+
+        Args:
+            session(Session): The SQLAlchemy DB Session to use.
+            key(str): The key for the name of the version type to set.
+            value(int): The version value to set.
+        """
+        # Insert if key has no value yet, otherwise update the value
+        session.merge(Version(key=key, value=value))
 
     def mirror_entry_fields(self, entries: list[Entry]) -> None:
         """Mirror fields among multiple Entry items."""
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -541,6 +541,13 @@ class Library:
             )
 
         logger.info(f"[Library] Library DB version: {loaded_db_version}")
+
+        # save backup if patches will be applied
+        if loaded_db_version < DB_VERSION:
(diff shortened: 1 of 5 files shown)


src/tagstudio/core/library/alchemy/migrations.py
```diff
@@ -0,0 +1,549 @@
...
+)
+from tagstudio.core.library.alchemy.joins import TagParent
+from tagstudio.core.library.alchemy.models import (
+    Tag,
+    TagColorGroup,
+    Version,
+)
+from tagstudio.core.library.ignore import migrate_ext_list
+from tagstudio.core.utils.types import unwrap
+from tagstudio.qt.translations import Translations
```
CyanVoxel, line 27: nit: these imports can be condensed

src/tagstudio/core/library/alchemy/migrations.py
```diff
@@ -0,0 +1,549 @@
...
+                    self.loaded_db_version = migration.version
+                    try:
+                        Library._set_version(session, DB_VERSION_CURRENT_KEY, migration.version)
+                    except Exception as e:
+                        logger.info(
+                            f"[Library][Migration][{migration.version}] "
+                            "Couldn't update version, continuing without commit",
+                            error=e,
+                        )
+                        session.flush()
```
CyanVoxel, line 122: Why the `session.flush()` here?

src/tagstudio/core/library/alchemy/migrations.py
```diff
@@ -0,0 +1,549 @@
...
+                        self.library_dir,
+                        lambda msg, v=migration.version: f"[Library][Migration][{v}] {msg}",
+                    )
+                    self.loaded_db_version = migration.version
+                    try:
+                        Library._set_version(session, DB_VERSION_CURRENT_KEY, migration.version)
+                    except Exception as e:
+                        logger.info(
+                            f"[Library][Migration][{migration.version}] "
+                            "Couldn't update version, continuing without commit",
```
CyanVoxel, line 119: Shouldn't the rest of the migration process be aborted if a one fails? That way subsequent migrations that expect a specific schema and dataset don't fail, or worse, mangle data

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -2160,7 +1778,8 @@ def get_version(self, key: str) -> int:
             except Exception:
                 return 0
 
-    def set_version(self, session: Session, key: str, value: int) -> None:
+    @staticmethod
+    def _set_version(session: Session, key: str, value: int) -> None:
```
CyanVoxel: These should be public if they're used in `DBMigrations`

src/tagstudio/core/library/alchemy/migrations.py
```diff
@@ -0,0 +1,549 @@
...
+            "Ran all migrations, but the DB is still not on the newest version"
+        )
+        logger.info(f"[Library][Migration] Library migrated to DB version {DB_VERSION}")
+
+
+class MigrationTo7(DBMigration):
+    version = 7
+
+    @classmethod
+    def run(cls, session: Session, library_dir: Path, fmt_log):
```
CyanVoxel, line 149: All the `run()` methods in these `DBMigration` child classes should also have the `@override` decorator

CyanVoxel (commented):
Looks like somethings not quite right with the migrations, here I threw in a DB 9 library and it failed on migration 200:
Image
```
2026-07-24 03:51:02 [info     ] [Library][Migration][200] Starting DB Migration
2026-07-24 03:51:02 [info     ] [Library][Migration][200] Dropping boolean_fields and value_type tables...
2026-07-24 03:51:02 [info     ] [Library][Migration][200] Adding name columns to field tables...
2026-07-24 03:51:02 [info     ] [Library][Migration][200] Dropping position columns to field tables...
2026-07-24 03:51:02 [info     ] [Library][Migration][200] Adding is_multiline column to text_fields...
2026-07-24 03:51:02 [info     ] [Library][Migration][200] Moving values from type_key columns to name...
2026-07-24 03:51:02 [info     ] [Library][Migration][200] Normalizing TextField names...
2026-07-24 03:51:02 [info     ] [Library][Migration][200] Normalizing DatetimeField names...
2026-07-24 03:51:02 [info     ] [Library][Migration][200] Updating is_multiline for legacy TEXT_BOXes...
2026-07-24 03:51:02 [info     ] [Library][Migration][200] Repairing legacy Description fields...
2026-07-24 03:51:02 [info     ] [Library][Migration][200] Repairing legacy Comment fields...
2026-07-24 03:51:02 [info     ] [Library][Migration][200] Adding default field templates...
2026-07-24 03:51:02 [error    ] (sqlite3.OperationalError) no such table: datetime_field_templates
[SQL: INSERT INTO datetime_field_templates (name) VALUES (?)]
[parameters: ('Date',)]
(Background on this error at: https://sqlalche.me/e/20/e3q8)
```
Verified working on `main`

src/tagstudio/core/library/alchemy/migrations.py
Computerdores: this is only about setting the version, if the actual migration fails it doesn't continue.
I introduced this because the earlier versions that we have migrations for don't have the versions table yet and we removed support for the earlier versions from `set_version` at some point (rightly so imo, we have enough code to wrangle as is). This makes it so those can run eventough the version can't be updated immediately, the later migrations (specifically 101 and onwards) will be able to set the version and then also commit.

This is basically the migration process being optimistic that setting the version will succeed in a future migration even if it failed on the current one. If all migrations complete and no commit happens after the last one, the affected migrations will be rolled back and an error will be thrown because of the assert.

src/tagstudio/core/library/alchemy/migrations.py
Computerdores: A fallback in case the migration doesn't flush on its own

src/tagstudio/core/library/alchemy/library.py
Computerdores: I have moved `_set_version` to `DBMigrations.__set_version` as it isn't used anywhere else.

I still think it would be useful to differentiate between `_` and `__`, where `_` essentially means "private, except for a small number of exceptions". Because in this case for example I intended `_get_version` to only be used in the `.alchemy` submodule as it directly touches sql stuff (also there would be a name collision with `get_version` if the `_` were removed). Other examples are for example when a private function needs to be used by the tests, where it doesn't make sense to make it public, but keeping it private (via `__`) isn't possible either.
Another advantage would that looking at a function with `__` immediately tells you that it isn't used outside the class, whereas `_` leaves that open.

Computerdores:
> Looks like somethings not quite right with the migrations, here I threw in a DB 9 library and it failed on migration 200:

Yeah turns out the one time I didn't test, I overlooked something ^^'

I forgot that adding relationships also adds tables that need to be created.
Fyi I'll get around to that, but I don't have time today, I'll let you know when that's done

src/tagstudio/core/library/alchemy/migrations.py
```diff
@@ -0,0 +1,551 @@
...
+
+logger = structlog.get_logger(__name__)
+
+
+class MigrationError(Exception):
+    pass
+
+
+class DBMigration:
+    version: int = None  # pyright: ignore[reportAssignmentType]
```
CyanVoxel: This (should) be able to just be unbound to avoid the pyright error
```suggestion
    version: int
```

src/tagstudio/core/library/alchemy/library.py
CyanVoxel: The reason I ended up souring on the `_` and `__` delineation is because in practice it became very difficult to keep up with, where I found myself constantly refactoring variable names to add or remove underscores across more and more affected files while simultaneously getting frustrated at having to type "<kbd>Shift</kbd>+<kbd>-</kbd>+<kbd>-</kbd>" over and over again *just* to start typing variable names - all for what is a superficial meaning placed onto variables for a language that doesn't actually enforce these concepts at best, and implements them differently in a way we're not respecting (i.e. name mangling) at worst.

It sounds like a good idea on paper, and believe me if Python had actual `private` and `protected` keywords I'd be all over them, but in practice the two types of prefixes is just more of a burden than they're worth imo, and a single underscore will do just fine to indicate intended private usage. If an attribute is meant to be used *sometimes* publicly (tests being an exception), then it should just be public - and if there are aspects of an attribute that shouldn't be allowed publicly, then those should be hidden away somehow, like a read-only getter property.

src/tagstudio/core/library/alchemy/migrations.py
```diff
@@ -0,0 +1,551 @@
...
+                    except Exception as e:
+                        logger.info(
+                            f"[Library][Migration][{migration.version}] "
+                            "Couldn't update version, continuing without commit",
+                            error=e,
+                        )
+                        session.flush()
+                    else:
+                        session.commit()
+                logger.info(f"[Library][Migration][{migration.version}] Completed DB Migration")
```
purpletennisball, line 125: Wouldn't this print even if the migration fails? Just a little nit pick

Computerdores: the method would stop executing on the error like normal, no?

Computerdores:
Ok it works now, and it turns out I was wrong about why I was wrong.
The actual reason was that the `fields.py` file also contains model definitions even though we specifically have a `models.py` file...

src/tagstudio/core/library/alchemy/migrations.py
Computerdores: I have tested it and it does indeed not get executed

CyanVoxel (approved):
> The actual reason was that the fields.py file also contains model definitions even though we specifically have a models.py file...

Gotta love it... 🙃 

Anyway, the issue seems to be fixed now! Thank you for your continued work on this!

PR #1459, opened 2026-07-24
ci(docs): migrate from mkdocs to properdocs

CyanVoxel:
### Summary
This PR migrates from MkDocs to the [ProperDocs](https://github.com/ProperDocs/properdocs) fork, since MkDocs 1.x has been abandoned and has had several bugs accruing, notably the inability to hot reload changes on the site while using `mkdocs serve`. ProperDocs is a fork that implements several bugfixes and is receiving limited maintenance support, which is more than can be said about the situation with MkDocs. A longer-term solution may be to eventually switch to [Zensical](https://zensical.org/about/) once that project matures more.

See: https://github.com/orgs/ProperDocs/discussions/33

### Tasks Completed
- Platforms Tested:
    - [x] `properdocs serve`


.github/workflows/build_docs.yml
```diff
@@ -50,7 +50,7 @@ jobs:
           path: .cache
           restore-keys: |
             mkdocs-material-
-      - name: Run mkdocs
+      - name: Run properdocs
         env:
           DISABLE_MKDOCS_2_WARNING: 'true'
```
xarvex:
```suggestion
```

Redundant/unused with properdocs.

xarvex (changes requested):
Seems to check out fine, just the note I left about the env var.

xarvex (approved):

Issue #1460, opened 2026-07-24
[Task]: Systems Cleanup & Improvement Pass


CyanVoxel:
### Checklist

- [x] I am using an up-to-date version.
- [x] I have read the [documentation](https://github.com/TagStudioDev/TagStudio/blob/main/docs/index.md).
- [x] I have searched existing [issues](https://github.com/TagStudioDev/TagStudio/issues).

### Description

This task tracks an ongoing systems cleanup pass I'm doing to go in and cleanup, fix, and add miscellaneous improvements to various TagStudio systems.

Over time much of the codebase has become out of alignment with its intended functionality and implementations - especially on the UI side. Anti-patterns have arisen to enforce early practices of mine that were never intended to be canonized, some systems have become neglected or bloated since being introduced, and some systems have fallen into complete disrepair. 

I feel that it's very important that I myself take a look at these systems and set them back into alignment to what they should be and how they should operate, not only to fix and improve them for their sake but also to set a proper baseline for how these systems should function. Code should not be built on top of rough systems I added in the initial commit 2 years ago, nor on top of systems that have long outgrown their original purposes or have become a stack of band-aid fixes addressing problems that may not even be relevant anymore.

I'll be working on these at a steady pace over time, but not exclusively. I'll still be giving attention to new features and the ongoing core library changes in addition to what's listed here. 

### Ordered Tasks
- [ ] 0. UI MVC Refactor
    - [ ] Necessary to complete first for several items, as many of the ordered tasks touch the UI code which is either currently fragmented or using the older MVC pattern.
    - See #1458 + #1451 for implementation details
- [ ] 1. Media Type System
    - [ ] Create different mechanisms for file categories that are used for different purposes, such as searching vs. rendering (see comment: https://github.com/TagStudioDev/TagStudio/pull/1430#issuecomment-4898677986)
    - [ ] Create translated media type categories for improved searching capabilities
    - Focus on the usability of this system - do not over-complicate it but still provide the expanded functionality outlined above
- [ ] 2. Thumbnail/Preview Renderer
    - [x] Split into separate classes and files for each renderer type (would replace #1217)
    - [ ] Offload decorative frame rendering and possibly default file icons to Qt instead of using PIL
    - [ ] Have previews and thumbnails use alpha channel for backgrounds to support differing window background colors (progress towards #671)
    - [ ] Respect current color scheme
[...]

PR #1462, opened 2026-07-25
refactor: split ThumbRenderer class

CyanVoxel:
### Summary
Splits up the `ThumbnailRenderer` class into smaller classes and functions inside a new `renderers/` directory and consolidates code inside them (specifically for the functions moved inside the `archive.py` file, the rest is basically untouched)

**NOTE:** I'd like to move the `renderer.py` file into this directory as well, but haven't done so in this PR as a move within the same PR as these changes would result in the moved file losing the git history. That can come later along with several other file renames/moves.

Takes over for #1217, progress towards #1460.

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [x] macOS ARM
    - [ ] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable


src/tagstudio/renderers/archive.py
```diff
@@ -0,0 +1,158 @@
...
+    """Extract and render an apple embedded thumbnail (iWork, Apple Creative Studio)."""
+    image_names: list[str] = [
+        "preview.jpg",
+        "QuickLook/Preview.heic",
+        "QuickLook/Thumbnail.jpg",
+        "QuickLook/Thumbnail.heic",
+        "QuickLook/Thumbnail.webp",
+        "QuickLook/Icon.webp",
+    ]
+    return archive_thumb(filepath, image_names=image_names)
```
Computerdores: Nit: Here and in the following methods:
```suggestion
    return archive_thumb(filepath, image_names)
```

src/tagstudio/renderers/archive.py
```diff
@@ -0,0 +1,158 @@
...
+        image_names: (list[Path] | list[str] | None): List of embedded image names to search for.
+
+    Returns:
+        Image: The first image found in the archive.
+    """
+    try:
+        with open_archive(filepath, ext) as archive:
+            # If no list of image names to search for was provided, default to the first image.
+            if not image_names:
+                return first_image_in_archive(archive)
```
Computerdores, line 120: This should also happen when none of the paths in `image_names` are present, no?

Computerdores (commented):
Mostly looks good

FYI: I reviewed under the assumption, that you only extracted the rendering methods in the first commit and didn't change them

src/tagstudio/renderers/clip_studio.py
```diff
@@ -0,0 +1,80 @@
...
+            if thumbnail:
+                im = Image.open(BytesIO(thumbnail[0]))
+        conn.close()
+    except Exception as e:
+        logger.error("Couldn't render thumbnail", filepath=filepath, error=type(e).__name__)
+
+    return im
+
+
+def paint_dot_net_thumb(filepath: Path) -> Image.Image | None:
```
Sola-ris: Shouldn't this one be in its own file like `medibang_paint_thumb`?

src/tagstudio/renderers/archive.py
CyanVoxel: This was because of the parameter order, but I've reordered them since all but one call doesn't use the `ext` argument so in exchange for cleaning these up there will be one call that looks like this:`image = archive_thumb(filepath_, ext=ext)`

PR #1470, opened 2026-08-08
feat: add per-tag category display options

Sola-ris:
### Summary

Add per-tag category display options

Resolves https://github.com/TagStudioDev/TagStudio/issues/1335
Supersedes https://github.com/TagStudioDev/TagStudio/pull/1336

By default, tags are displayed in all inherited categories, leaving the current behavior unchanged.
This extends the `BuildTagPanel` with the options to exclude a tag from being displayed in one, multiple or all possible categories.

The exclusions are stored in the new `category_exclusions` join table.
Also bumps the database version to 301.

#### Demo

https://github.com/user-attachments/assets/3713377e-247a-4c70-b0ca-90518b0d0a2b

### Tasks Completed

-   Platforms Tested:
    -   [ ] Windows x86
    -   [ ] Windows ARM
    -   [ ] macOS x86
    -   [ ] macOS ARM
    -   [x] Linux x86
    -   [ ] Linux ARM
    
-   Tested For:
    -   [x] Basic functionality
    -   [ ] PyInstaller executable

PR #1471, opened 2026-08-12
fix: close dangling session, migration corrections

### LINKED ISSUE #1447: [Bug]: Database locked after refreshing unlinked
### Checklist

### TagStudio Version

main branch commit 6aa0cf74f9a47845dcb0822139ffeb688aeeb467

NixOS 26.05

### Description

My initial discovery of the issue was that I went to use the "fix unlinked" tool and after pressing "Refresh" and then "Search & Relink" I can see in the logs that it's failing the inserts with "Database is locked" error.
After shutting it down and restarting and the refreshing again I closed the tool and now I can't add any tags because of "database is locked" errors.
I then rolled back to before #1432 as it seemed like the latest change, and now it works. I believe that the issue might stem from the new implementation of `all_entries` in `src/tagstudio/core/library/alchemy/library.py` where I believe it's loosing track of a session that can't be flushed/closed.

This I believe happens in:
```python
def all_entries(self, with_joins: bool = False) -> Iterator[Entry]:
        """Load entries without joins."""
            return self.__all_entries(session, with_joins)
```
where the inner function:
```python
def __all_entries(self, session: Session, with_joins: bool = False) -> Iterator[Entry]:
        """Load entries without joins."""
        stmt = select(Entry)
        if with_joins:
            # load Entry with all joins and all tags
            stmt = (
                stmt.outerjoin(Entry.text_fields)
                .outerjoin(Entry.datetime_fields)
                .outerjoin(Entry.tags)
            )
            stmt = stmt.options(
                contains_eager(Entry.text_fields),
                contains_eager(Entry.datetime_fields),
                contains_eager(Entry.tags),
            )

        stmt = stmt.distinct()

        entries = session.execute(stmt).scalars()
        if with_joins:
            entries = entries.unique()

        for entry in entries:
            yield entry
            session.expunge(entry)
```
is a  generator that holds onto the other session.
This generator is then used in the unlinking refresh to find all the entries, but the session is never closed because it has left the original scope. This then puts the database in a deadlock, and can only be freed by killing the application.

I also think this verifies the suspicion that the test in `tests/macros/test_missing_files.py` doesn't work.

### Expected Behavior

I should be able to refresh and relink entries.

### Steps to Reproduce

1. Open "Tools" -> "Fix Unlinked Entries"
2. Click "Refresh"
3. Close the pop-up
4. Add tag to existing entry

### Logs

2026-07-14 01:35:27 [info     ] [Library] Library DB version: 202
2026-07-14 01:35:27 [info     ] [Library] Creating DB tables...
2026-07-14 01:35:27 [info     ] [Library] Library migrated to DB version 202
2026-07-14 01:35:27 [info     ] [Config] Thumbnail Cache Size: 1.02 GB
2026-07-14 01:35:27 [info     ] [Ignore] Processing the .ts_ignore file... last_mtime=None library=PosixPath('<path-to-project>/TagStudio/test-library') new_mtime=1765618780.0270758
2026-07-14 01:35:27 [info     ] [Ignore]                       glob_patterns=[<ignore-patterns>]
2026-07-14 01:35:27 [info     ] [Ignore] No updates to the .ts_ignore detected last_mtime=1765618780.0270758 library=PosixPath('<path-to-project>/TagStudio/test-library') new_mtime=1765618780.0270758
2026-07-14 01:35:27 [info     ] searching library              filter=BrowsingState(page_index=0, page_positions={}, sorting_mode=<SortingModeEnum.DATE_ADDED: 'file.date_added'>, ascending=False, random_seed=0, show_hidden_entries=False, query=None) query_full='SELECT DISTINCT entries.id \nFROM entries \nWHERE (entries.id NOT IN (SELECT tag_entries.entry_id \nFROM tag_entries \nWHERE tag_entries.tag_id IN (SELECT tags.id \nFROM tags \nWHERE tags.is_hidden))) ORDER BY entries.id DESC'
2026-07-14 01:35:27 [info     ] SQL Execution finished (0 seconds)
2026-07-14 01:35:27 [info     ] items to render                count=8
2026-07-14 01:35:32 [info     ] Closing Library...            
2026-07-14 01:35:32 [info     ] [Library] Creating SQLAlchemy Engine connection_string=sqlite:///<tag-studio-library-path>/.TagStudio/ts_library.sqlite poolclass=<class 'sqlalchemy.pool.impl.NullPool'>
2026-07-14 01:35:32 [info     ] [Library] Opening SQLite Library library_dir=PosixPath('<tag-studio-library-path>')
2026-07-14 01:35:32 [info     ] [Library] Library DB version: 202
2026-07-14 01:35:32 [info     ] [Library] Creating DB tables...
2026-07-14 01:35:32 [info     ] [Library] Library migrated to DB version 202
2026-07-14 01:35:33 [info     ] [Config] Thumbnail Cache Size: 1.02 GB
2026-07-14 01:35:33 [info     ] [Ignore] Processing the .ts_ignore file... last_mtime=1765618780.0270758 library=PosixPath('<tag-studio-library-path>') new_mtime=1763158157.844917
2026-07-14 01:35:33 [info     ] [Ignore]                       glob_patterns=[<ignore-patterns>]
2026-07-14 01:35:33 [info     ] [Ignore] No updates to the .ts_ignore detected last_mtime=1763158157.844917 library=PosixPath('<tag-studio-library-path>') new_mtime=1763158157.844917
2026-07-14 01:35:33 [info     ] searching library              filter=BrowsingState(page_index=0, page_positions={}, sorting_mode=<SortingModeEnum.DATE_ADDED: 'file.date_added'>, ascending=False, random_seed=0, show_hidden_entries=False, query=None) query_full='SELECT DISTINCT entries.id \nFROM entries \nWHERE (entries.id NOT IN (SELECT tag_entries.entry_id \nFROM tag_entries \nWHERE tag_entries.tag_id IN (SELECT tags.id \nFROM tags \nWHERE tags.is_hidden))) ORDER BY entries.id DESC'
2026-07-14 01:35:33 [info     ] SQL Execution finished (0.32 seconds)
2026-07-14 01:35:33 [info     ] items to render                count=193275
2026-07-14 01:35:38 [info     ] [UnlinkedRegistry] Refreshing unlinked files...
2026-07-14 01:35:58 [info     ] [QtDriver] Selecting Items:    append=False bridge=False item_id=253255
2026-07-14 01:35:58 [warning  ] [FieldContainers] Updating Selection entry_id=253255
2026-07-14 01:35:58 [info     ] [Library] Time it took to get entry: 0.01 seconds with_fields=True with_tags=True
2026-07-14 01:35:58 [info     ] [emit_badge_signals] Emitting  emit_on_absent=True tag_ids=set()
2026-07-14 01:35:58 [info     ] [QtDriver][update_badges] Updating ItemThumb badges add_tags=False badge_values={<BadgeType.ARCHIVED: 'Archived'>: False} origin_id=0
2026-07-14 01:35:58 [info     ] [QtDriver][update_badges] Updating ItemThumb badges add_tags=False badge_values={<BadgeType.FAVORITE: 'Favorite'>: False} origin_id=0
2026-07-14 01:35:58 [info     ] [QtDriver] Selecting Items:    append=False bridge=False item_id=253255
2026-07-14 01:35:58 [warning  ] [FieldContainers] Updating Selection entry_id=253255
2026-07-14 01:35:58 [info     ] [Library] Time it took to get entry: 0 seconds with_fields=True with_tags=True
2026-07-14 01:35:58 [info     ] [emit_badge_signals] Emitting  emit_on_absent=True tag_ids=set()
2026-07-14 01:35:58 [info     ] [QtDriver][update_badges] Updating ItemThumb badges add_tags=False badge_values={<BadgeType.ARCHIVED: 'Archived'>: False} origin_id=0
2026-07-14 01:35:58 [info     ] [QtDriver][update_badges] Updating ItemThumb badges add_tags=False badge_values={<BadgeType.FAVORITE: 'Favorite'>: False} origin_id=0
2026-07-14 01:35:59 [info     ] [SearchPanel] Updating items   limit=25
2026-07-14 01:36:01 [info     ] searching tags                 limit=25 results=2 search=photo statement="SELECT tag_aliases.tag_id, tag_aliases.name \nFROM tag_aliases \nWHERE (lower(tag_aliases.name) LIKE '%' || lower(:name_1) || '%')"
2026-07-14 01:36:01 [info     ] [SearchPanel] Search results   results=[<Tag ID: 1200 Name: Photo>, <Tag ID: 2637 Name: AI Generated Photo>, <Tag ID: 1207 Name: Avatar>, <Tag ID: 1201 Name: Banner>, <Tag ID: 1243 Name: Selfie>]
2026-07-14 01:36:03 [info     ] [FieldContainers][add_tags_to_selected] selected=[253255] tag_ids=[1200]
2026-07-14 01:36:03 [info     ] [Library][add_tags_to_entries] entry_ids=[253255] tag_ids=[1200]
  File "<path-to-project>/TagStudio/.venv/lib/python3.13/site-packages/sqlalchemy/engine/base.py", line 1147, in _commit_impl
    self.engine.dialect.do_commit(self.connection)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "<path-to-project>/TagStudio/.venv/lib/python3.13/site-packages/sqlalchemy/engine/default.py", line 715, in do_commit
    dbapi_connection.commit()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
sqlite3.OperationalError: database is locked

The above exception was the direct cause of the following exception:

  File "<path-to-project>/TagStudio/src/tagstudio/qt/controllers/preview_panel_controller.py", line 56, in _add_tag_to_selected
    self._containers.add_tags_to_selected(tag_id)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "<path-to-project>/TagStudio/src/tagstudio/qt/mixed/field_containers.py", line 252, in add_tags_to_selected
    self.driver.add_tags_to_selected_callback(tag_ids)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "<path-to-project>/TagStudio/src/tagstudio/qt/ts_qt.py", line 912, in add_tags_to_selected_callback
    self.lib.add_tags_to_entries(selected, tag_ids)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "<path-to-project>/TagStudio/src/tagstudio/core/library/alchemy/library.py", line 1769, in add_tags_to_entries
    session.commit()
    ~~~~~~~~~~~~~~^^
  File "<path-to-project>/TagStudio/.venv/lib/python3.13/site-packages/sqlalchemy/orm/session.py", line 2035, in commit
    trans.commit(_to_root=True)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "<string>", line 2, in commit
  File "<path-to-project>/TagStudio/.venv/lib/python3.13/site-packages/sqlalchemy/orm/state_changes.py", line 137, in _go
    ret_value = fn(self, *arg, **kw)
  File "<path-to-project>/TagStudio/.venv/lib/python3.13/site-packages/sqlalchemy/orm/session.py", line 1323, in commit
    trans.commit()
    ~~~~~~~~~~~~^^
  File "<path-to-project>/TagStudio/.venv/lib/python3.13/site-packages/sqlalchemy/engine/base.py", line 2642, in commit
    self._do_commit()
    ~~~~~~~~~~~~~~~^^
  File "<path-to-project>/TagStudio/.venv/lib/python3.13/site-packages/sqlalchemy/engine/base.py", line 2747, in _do_commit
    self._connection_commit_impl()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "<path-to-project>/TagStudio/.venv/lib/python3.13/site-packages/sqlalchemy/engine/base.py", line 2718, in _connection_commit_impl
    self.connection._commit_impl()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "<path-to-project>/TagStudio/.venv/lib/python3.13/site-packages/sqlalchemy/engine/base.py", line 1149, in _commit_impl
    self._handle_dbapi_exception(e, None, None, None, None)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<path-to-project>/TagStudio/.venv/lib/python3.13/site-packages/sqlalchemy/engine/base.py", line 2365, in _handle_dbapi_exception
    raise sqlalchemy_exception.with_traceback(exc_info[2]) from e
  File "<path-to-project>/TagStudio/.venv/lib/python3.13/site-packages/sqlalchemy/engine/base.py", line 1147, in _commit_impl
    self.engine.dialect.do_commit(self.connection)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "<path-to-project>/TagStudio/.venv/lib/python3.13/site-packages/sqlalchemy/engine/default.py", line 715, in do_commit
    dbapi_connection.commit()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) database is locked
(Background on this error at: https://sqlalche.me/e/20/e3q8)

### LINKED ISSUE #1467: [Bug]: Infinite loading when modifying the library (Adding a file, unlink, ignore, ect)
### Checklist

### TagStudio Version

Alpha 9.6.2 Portable

Windows 11, version 25H2

### Description

On the latest version of Tagstudio, when you try to change anything in the library that gives the loading window, said loading will be infinite.
So far I've tested
- Ignored files
- Unlinked files
- Refreshing the library after adding a file in the folders library

Might be the same for fixing duplicates.
Files are successfully added to the library after closing up the library and opening it again.
Tags have no issue.

Can't upload the evidence directly in github so I'll add a Youtube link instead.
https://www.youtube.com/watch?v=8dATVkRbmIM

### Expected Behavior

Loading screen must end quickly depending on the amount of files then whatever task I had previously asked will finish.

### Steps to Reproduce

1. Open a library in the version 9.6.2 Alpha (Portable).
2. Either add a file, unlink one or ignore a type of file.
3. The bug will naturally happen until the loading window is manually closed.

### Logs

_No response_

CyanVoxel:
### Summary
This PR fixes issues caused by a missing session context manager in the `all_entries()` method in conjunction with the new `autocommit=False` DB parameter, and makes some tweaks to the migration refactors done in #1432 and #1456, most notably removing static methods in the `Library` class and removing local imports <sup>because I'm an idiot and forgot that I didn't completely review #1456 before signing off and pulling it</sup>

Closes #1447, Closes #1467

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [x] macOS ARM
    - [ ] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/src/tagstudio/core/library/alchemy/library.py
+++ b/src/tagstudio/core/library/alchemy/library.py
@@ -221,7 +221,6 @@ class Library:
         if self.engine:
             self.engine.dispose()
         self.library_dir = None
-        self.folder = None
         self.included_files = set()
 
         self.dupe_entries_count = -1
@@ -378,8 +377,7 @@ class Library:
 
         return self.open_sqlite_library(library_dir, in_memory)
 
-    @staticmethod
-    def __get_engine(library_dir: Path, in_memory: bool, sql_filename: str):
+    def _get_engine(self, library_dir: Path, in_memory: bool, sql_filename: str):
         connection_string = URL.create(
             drivername="sqlite",
             database=(
@@ -407,7 +405,7 @@ class Library:
     def create_sqlite_library(
         self, library_dir: Path, in_memory: bool, sql_filename: str = SQL_FILENAME
     ) -> LibraryStatus:
-        self.engine = self.__get_engine(library_dir, in_memory, sql_filename)
+        self.engine = self._get_engine(library_dir, in_memory, sql_filename)
 
         logger.info(
             "[Library] Opening SQLite Library",
@@ -505,21 +503,21 @@ class Library:
     ) -> LibraryStatus:
         logger.info("[Library] Opening SQLite Library", library_dir=library_dir)
 
-        self.engine = self.__get_engine(library_dir, in_memory, sql_filename)
+        self.engine = self._get_engine(library_dir, in_memory, sql_filename)
+        self.library_dir = library_dir
 
         try:
-            migrations = DBMigrations(library_dir, self.engine)
+            migrations = DBMigrations(self)
 
             # save backup if patches will be applied
             if migrations.required:
-                Library.save_library_backup_to_disk(library_dir)
+                self.save_library_backup_to_disk()
 
             migrations.run()
         except MigrationError as e:
+            self.library_dir = None
             return LibraryStatus(success=False, message=e.args[0])
 
-        # everything is fine, set the library path
-        self.library_dir = library_dir
         return LibraryStatus(success=True, library_path=library_dir)
 
     @property
@@ -669,37 +667,32 @@ class Library:
         with Session(self.engine) as session:
             return unwrap(session.scalar(select(func.count(Entry.id))))
 
-    @staticmethod
-    def _all_entries(session: Session, with_joins: bool = False) -> Iterator[Entry]:
-        """Load entries without joins."""
-        stmt = select(Entry)
-        if with_joins:
-            # load Entry with all joins and all tags
-            stmt = (
-                stmt.outerjoin(Entry.text_fields)
-                .outerjoin(Entry.datetime_fields)
-                .outerjoin(Entry.tags)
-            )
-            stmt = stmt.options(
-                contains_eager(Entry.text_fields),
-                contains_eager(Entry.datetime_fields),
-                contains_eager(Entry.tags),
-            )
-
-        stmt = stmt.distinct()
-
-        entries = session.execute(stmt).scalars()
-        if with_joins:
-            entries = entries.unique()
-
-        for entry in entries:
-            yield entry
-            session.expunge(entry)
-
     def all_entries(self, with_joins: bool = False) -> Iterator[Entry]:
         """Load entries without joins."""
         with Session(self.engine) as session:
-            return Library._all_entries(session, with_joins)
+            stmt = select(Entry)
+            if with_joins:
+                # load Entry with all joins and all tags
+                stmt = (
+                    stmt.outerjoin(Entry.text_fields)
+                    .outerjoin(Entry.datetime_fields)
+                    .outerjoin(Entry.tags)
+                )
+                stmt = stmt.options(
+                    contains_eager(Entry.text_fields),
+                    contains_eager(Entry.datetime_fields),
+                    contains_eager(Entry.tags),
+                )
+
+            stmt = stmt.distinct()
+
+            entries = session.execute(stmt).scalars()
+            if with_joins:
+                entries = entries.unique()
+
+            for entry in entries:
+                yield entry
+                session.expunge(entry)
 
     @property
     def tags(self) -> list[Tag]:
@@ -1447,17 +1440,16 @@ class Library:
                 session.rollback()
                 return None
 
-    @staticmethod
-    def save_library_backup_to_disk(library_dir: Path) -> Path:
-        assert isinstance(library_dir, Path)
-        makedirs(str(library_dir / TS_FOLDER_NAME / BACKUP_FOLDER_NAME), exist_ok=True)
+    def save_library_backup_to_disk(self) -> Path:
+        assert isinstance(self.library_dir, Path)
+        makedirs(str(self.library_dir / TS_FOLDER_NAME / BACKUP_FOLDER_NAME), exist_ok=True)
 
         filename = f"ts_library_backup_{datetime.now(UTC).strftime('%Y_%m_%d_%H%M%S')}.sqlite"
 
-        target_path = library_dir / TS_FOLDER_NAME / BACKUP_FOLDER_NAME / filename
+        target_path = self.library_dir / TS_FOLDER_NAME / BACKUP_FOLDER_NAME / filename
 
         shutil.copy2(
-            library_dir / TS_FOLDER_NAME / SQL_FILENAME,
+            self.library_dir / TS_FOLDER_NAME / SQL_FILENAME,
             target_path,
         )
 
@@ -1749,12 +1741,8 @@ class Library:
         Args:
             key(str): The key for the name of the version type to set.
         """
-        return Library._get_version(self.engine, key)
-
-    @staticmethod
-    def _get_version(engine, key: str) -> int:
-        with Session(engine) as session:
-            engine = sqlalchemy.inspect(engine)
+        with Session(self.engine) as session:
+            engine = sqlalchemy.inspect(self.engine)
             try:
                 # "Version" table added in DB_VERSION 101
                 if engine and engine.has_table("versions"):
--- a/src/tagstudio/core/library/alchemy/migrations.py
+++ b/src/tagstudio/core/library/alchemy/migrations.py
@@ -4,11 +4,11 @@
 
 from collections.abc import Callable
 from pathlib import Path
-from typing import override
+from typing import TYPE_CHECKING, override
 
 import structlog
 import ujson
-from sqlalchemy import Engine, and_, delete, select, text, update
+from sqlalchemy import and_, delete, select, text, update
 from sqlalchemy.orm import Session
 
 from tagstudio.core.constants import IGNORE_NAME, TAG_ARCHIVED, TS_FOLDER_NAME
@@ -21,10 +21,13 @@ from tagstudio.core.library.alchemy.constants import (
 )
 from tagstudio.core.library.alchemy.fields import LEGACY_FIELD_MAP, DatetimeField, TextField
 from tagstudio.core.library.alchemy.joins import TagParent
-from tagstudio.core.library.alchemy.models import Tag, TagColorGroup, Version
+from tagstudio.core.library.alchemy.models import Entry, Tag, TagColorGroup, Version
 from tagstudio.core.library.ignore import migrate_ext_list
 from tagstudio.core.utils.types import unwrap
 from tagstudio.qt.translations import Translations
+
+if TYPE_CHECKING:
+    from tagstudio.core.library.alchemy.library import Library
 
 logger = structlog.get_logger(__name__)
 
@@ -38,20 +41,19 @@ class DBMigration:
     initial_version: int | None = None
 
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log: Callable[[str], str]):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]) -> None:  # pyright: ignore[reportUnusedParameter]
         raise NotImplementedError
 
 
 class DBMigrations:
-    def __init__(self, library_dir: Path, engine: Engine) -> None:
-        from tagstudio.core.library.alchemy.library import Library
+    def __init__(self, library: "Library") -> None:
 
-        self.library_dir = library_dir
-        self.engine = engine
+        self.lib = library
+        self.engine = self.lib.engine
 
         # Don't check DB version when creating new library
-        self.loaded_db_version = Library._get_version(engine, DB_VERSION_CURRENT_KEY)
-        self.initial_db_version = Library._get_version(engine, DB_VERSION_INITIAL_KEY)
+        self.loaded_db_version = self.lib.get_version(DB_VERSION_CURRENT_KEY)
+        self.initial_db_version = self.lib.get_version(DB_VERSION_INITIAL_KEY)
 
         # ======================== Library Database Version Checking =======================
         # DB_VERSION 6 is the first supported SQLite DB version.
@@ -107,7 +109,7 @@ class DBMigrations:
                     # any error causes transaction to rollback
                     migration.run(
                         session,
-                        self.library_dir,
+                        self.lib,
                         lambda msg, v=migration.version: f"[Library][Migration][{v}] {msg}",
                     )
                     self.loaded_db_version = migration.version
@@ -146,7 +148,7 @@ class MigrationTo7(DBMigration):
 
     @override
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]):
         """Migrate DB from DB_VERSION 6 to 7."""
         logger.info(fmt_log("Applying patches to DB_VERSION: 6 library..."))
         # Repair tags that may have a disambiguation_id pointing towards a deleted tag.
@@ -166,7 +168,7 @@ class MigrationTo8(DBMigration):
 
     @override
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]):
         """Migrate DB from DB_VERSION 7 to 8."""
         # Add the missing color_border column to the TagColorGroups table.
         session.execute(
@@ -219,7 +221,7 @@ class MigrationTo9(DBMigration):
 
     @override
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]):
         """Migrate DB from DB_VERSION 8 to 9."""
         # Apply database schema changes
         add_filename_column = text(
@@ -230,9 +232,8 @@ class MigrationTo9(DBMigration):
         logger.info(fmt_log("Added filename column to entries table"))
 
         # Populate the new filename column.
-        from tagstudio.core.library.alchemy.library import Library
-
-        for entry in Library._all_entries(session):
+        entries = session.execute(select(Entry).distinct()).scalars()
+        for entry in entries:
             entry.filename = entry.path.name
             session.merge(entry)
         session.flush()
@@ -244,7 +245,7 @@ class MigrationTo100(DBMigration):
 
     @override
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]):
         """Migrate DB to DB_VERSION 100."""
         # Repair parent-child tag relationships that are the wrong way around.
         stmt = update(TagParent).values(
@@ -261,7 +262,7 @@ class MigrationTo101(DBMigration):
 
     @override
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]):
         """Migrate DB to DB_VERSION 101."""
         # Create versions table
         session.execute(
@@ -284,7 +285,7 @@ class MigrationTo102(DBMigration):
 
     @override
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]):
         """Migrate DB to DB_VERSION 102."""
         # delete TagParents with a dangling parent reference
         stmt = delete(TagParent).where(TagParent.parent_id.not_in(select(Tag.id).distinct()))
@@ -298,7 +299,7 @@ class MigrationTo103(DBMigration):
 
     @override
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]):
         """Migrate DB from DB_VERSION 102 to 103."""
         # add the new hidden column for tags
         session.execute(text("ALTER TABLE tags ADD COLUMN is_hidden BOOLEAN NOT NULL DEFAULT 0"))
@@ -316,17 +317,17 @@ class MigrationTo104(DBMigration):
 
     @override
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]):
         """Migrate DB from DB_VERSION 103 to 104."""
         # Convert file extension list to ts_ignore file, if a .ts_ignore file does not exist
-        cls.__migrate_sql_to_ts_ignore(session, library_dir)
+        cls.__migrate_sql_to_ts_ignore(session, library)
         session.execute(text("DROP TABLE preferences"))
         session.flush()
 
     @classmethod
-    def __migrate_sql_to_ts_ignore(cls, session: Session, library_dir: Path):
+    def __migrate_sql_to_ts_ignore(cls, session: Session, library: "Library"):
         # Do not continue if existing '.ts_ignore' file is found
-        ts_ignore = library_dir / TS_FOLDER_NAME / IGNORE_NAME
+        ts_ignore = unwrap(library.library_dir) / TS_FOLDER_NAME / IGNORE_NAME
         if Path(ts_ignore).exists():
             return
 
@@ -349,7 +350,7 @@ class MigrationTo200(DBMigration):
 
     @override
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]):
         """Migrate DB to DB_VERSION 200."""
         # Drop unused 'boolean_fields' and 'value_type' tables
         logger.info(fmt_log("Dropping boolean_fields and value_type tables..."))
@@ -463,7 +464,7 @@ class MigrationTo201(DBMigration):
 
     @override
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]):
         """Migrate DB to DB_VERSION 201."""
         create_text_fields_table = text("""
         CREATE TABLE text_fields_new (
@@ -519,7 +520,7 @@ class MigrationTo202(DBMigration):
 
     @override
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]):
         """Migrate DB to DB_VERSION 202."""
         stmt = delete(TagParent).where(TagParent.child_id.not_in(select(Tag.id).distinct()))
         session.execute(stmt)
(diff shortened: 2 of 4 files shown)


src/tagstudio/core/library/alchemy/library.py
```diff
@@ -378,8 +377,7 @@ def open_library(self, library_dir: Path, in_memory: bool = False) -> LibrarySta
 
         return self.open_sqlite_library(library_dir, in_memory)
 
-    @staticmethod
-    def __get_engine(library_dir: Path, in_memory: bool, sql_filename: str):
+    def _get_engine(self, library_dir: Path, in_memory: bool, sql_filename: str):
```
Computerdores, line 380: Why make that non-static?

src/tagstudio/core/library/alchemy/migrations.py
```diff
@@ -38,20 +41,19 @@ class DBMigration:
...
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log: Callable[[str], str]):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]) -> None:  # pyright: ignore[reportUnusedParameter]
         raise NotImplementedError
 
 
 class DBMigrations:
-    def __init__(self, library_dir: Path, engine: Engine) -> None:
-        from tagstudio.core.library.alchemy.library import Library
+    def __init__(self, library: "Library") -> None:
```
Computerdores, line 49: I had originally passed the library path and engine because the engine argument will be removed when the migrations are made sql alchemy independent (and thus fully independent of the model that the to-be-migrated lib doesn't yet follow). Once that is done the parameter would only be the library path which is a much looser coupling then passing the full library in[^1].

For that reason I feel it is undesirable to pass the entire library to the constructor here.
Also, afaict it isn't necessary.

Note: Afaict the statements setting `library_dir` in ~~`Library.create_sqlite_library`~~ `Library.open_sqlite_library` can also be removed once this is reverted.

[^1]: We want the coupling to be as loose as possible here, because the logic in the Library class will be for the current schema while the migrations deal with old schemas and so we can't (and shouldn't where we still can) rely on that logic.

src/tagstudio/core/library/alchemy/migrations.py
```diff
@@ -230,9 +232,8 @@ def run(cls, session: Session, library_dir: Path, fmt_log):
         logger.info(fmt_log("Added filename column to entries table"))
 
         # Populate the new filename column.
-        from tagstudio.core.library.alchemy.library import Library
-
-        for entry in Library._all_entries(session):
+        entries = session.execute(select(Entry).distinct()).scalars()
+        for entry in entries:
```
Computerdores, line 241: Good change 👍
Tying this to the Library class was not good a decision by me.

src/tagstudio/core/library/alchemy/migrations.py
```diff
@@ -146,7 +148,7 @@ class MigrationTo7(DBMigration):
 
     @override
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log):
+    def run(cls, session: Session, library: "Library", fmt_log: Callable[[str], str]):
```
Computerdores, line 156: I like annotating the methods with the type for `fmt_log`, but I think it would be better to define something like `LoggingMethod = Callable[[str], str]` and to then set `fmt_log: LoggingMethod`

src/tagstudio/core/library/alchemy/library.py
```diff
@@ -1447,17 +1440,16 @@ def delete_color(self, color: TagColorGroup):
                 session.rollback()
                 return None
 
-    @staticmethod
-    def save_library_backup_to_disk(library_dir: Path) -> Path:
```
Computerdores, line 1451: I had made this static since the idea is to call it when the library isn't loaded yet (and can't be loaded at all due to needing to be migrated), that's why it was static and why I still think it should be here.

We could maybe also just move this to the `DBMigrations` to be done as the first step of `DBMigrations.run`.

src/tagstudio/core/library/alchemy/library.py
Computerdores, line 1777: This split existed, because it is being used in the migrations where the existing engine should be used to ensure consistency irrespective of future changes to the migrations.

However, I think this can just be moved to the migrations entirely as the only other use can be removed.
(It is in `LibraryInfoWindow` where `get_version(DB_VERSION_CURRENT_KEY)` is contrasted to `DB_VERSION`, which is useless because the migrations fail unless `get_version(DB_VERSION_CURRENT_KEY) == DB_VERSION` meaning that can be removed.)

src/tagstudio/qt/ts_qt.py
```diff
@@ -1651,7 +1651,8 @@ def open_library(self, path: Path) -> None:
         else:
             self._init_library(path, open_status)
 
-    def _init_library(self, path: Path, open_status: LibraryStatus):
+    def _init_library(self, path: Path, open_status: LibraryStatus, is_test: bool = False):
+        # TODO: Don't have an is_test parameter, the frontend and backend tasks here can be split.
```
Computerdores, line 1655: This todo is meant to be address in a future PR I suppose?
(just making sure it wasn't forgotten)

Computerdores:
> This PR fixes issues caused by a missing session context manager in the `all_entries()` method in conjunction with the new `autocommit=False` DB parameter

RE the actual fix: I am not quite sure why this fixes anything, because the original setup was not missing any context managers at all. The `_all_entries` method was simply shifting the burden of opening the context manager to the caller instead of opening its own (which would have caused errors in the migrations due to two DB sessions being opened at the same time). My best guess is that because the code now used in the migrations instead is much simpler, which makes me suspicious that the same error might still occur where `all_entries` is still used.

CyanVoxel:
> Why make[...] non-static?

For the non-static reversions, this was mostly motivated by the idea that the Library is not intended to be a singleton class - each instance of the Library class should be completely self-contained and methods for the class that affect instanced versions should only affect data for said instance. While making some methods static didn't necessarily turn the Library class into a singleton, it still separated the self-contained nature of the Library instance by requiring a session to be independently supplied (which raised a deeper issue with the `all_entries()` method). This also got rid of the local imports in `migrations.py`, which I had missed were added. 

> RE the actual fix: I am not quite sure why this fixes anything, because the original setup was not missing any context managers at all. The _all_entries method was simply shifting the burden of opening the context manager to the caller instead of opening its own (which would have caused errors in the migrations due to two DB sessions being opened at the same time). My best guess is that because the code now used in the migrations instead is much simpler, which makes me suspicious that the same error might still occur where all_entries is still used.

When turning the `all_entries()` method back into a non-static class an using the session associated with the Library class instance, that fixed the DB lock that was occurring presumably due to a mismatched instance being passed from the migrations that was holding the lock. Following the thread of this fix raised issues with migration 9, where the results from `all_entries()` would not contain the `flush`ed column that gets added in 9, which is where I switched to querying the DB directly via the migration's session instead of using the full `all_entries()` function that didn't contain the flushed changes *(I'm pretty sure this is a reason I had the migrations split in a weird order initially, as a poor workaround).* These changes also enforce using Library instances as a single unit instead of passing around `session` contexts that can become disjointed due to SQLAlchemy being ~~just wonderful to work with.~~

So a lot of the static reversions weren't 100% necessary to pull off this fix, but sort of went hand-in-hand with it and also reflect my own intentions for how the Library class should be used.

src/tagstudio/qt/ts_qt.py
CyanVoxel: Yep, since the presence of the thread initialization (in an unrelated **UI** test) started causing a segfault (in an unrelated **UI** test), I just patched this method with this flag that skirts around that and will save an actual logic/UI split in a future refactor PR that touches this area.

src/tagstudio/core/library/alchemy/library.py
CyanVoxel: > This split existed, because it is being used in the migrations where the existing engine should be used to ensure consistency irrespective of future changes to the migrations.

Where might an inconsistency arise that would be solved by passing a different engine to build a session from?

> However, I think this can just be moved to the migrations entirely as the only other use can be removed.
(It is in LibraryInfoWindow where get_version(DB_VERSION_CURRENT_KEY) is contrasted to DB_VERSION, which is useless because the migrations fail unless get_version(DB_VERSION_CURRENT_KEY) == DB_VERSION meaning that can be removed.)

The current UI-facing use of this is working as intended, where it shows the current loaded DB version in comparison to the program's currently known `DB_VERSION`, which is useful for telling if you have a DB loaded with a minor version greater than the TagStudio version that's opening it. I'd also like to not gate the `_get_version()` method to the migrations, as that would limit future uses in the UI.

Perhaps the "static" version of this that includes the engine parameter and inspection conditional could be moved to the migrations class, while a simpler instance method that just uses the currently expected logic could remain in the library file? Though that's a tiny bit of code duplication

src/tagstudio/core/library/alchemy/library.py
CyanVoxel: The `save_library_backup_to_disk()` method doesn't require a library to be loaded, it just uses the Library instance's `self.library_dir,` which I tweaked to be set before the migrations are run in` open_sqlite_library()`.

> We could maybe also just move this to the DBMigrations to be done as the first step of DBMigrations.run.

This method is also used outside of the migrations, like in the UI - so there still needs to be a non-static version of it present in the Library class for instances to use without passing the argument of the instance's own `library_dir`.

Computerdores:
> > Why make[...] non-static?
> 
> For the non-static reversions, this was mostly motivated by the idea that the Library is not intended to be a singleton class - each instance of the Library class should be completely self-contained and methods for the class that affect instanced versions should only affect data for said instance. 

This only applies to the `all_entries` method though (which I agree shouldn't have been static, no idea why I did that ^^).
Neither `_get_engine` nor `save_library_backup_to_disk` affect (or access) the data of the Library instance...

Also, the semantics of `save_library_backup_to_disk` are to make a backup of the library DB found at a certain path, not to make a backup of the currently opened library. And making the method non-static implies that it could access anything available on the instance when a library is accessed, which could lead to problems, because in order to call it from the migrations only the library path is set and the library is actually opened (which would of course not be possible). This is why I have to object to making `save_library_backup_to_disk` an instance method.

> When turning the `all_entries()` method back into a non-static class an using the session associated with the Library class instance,

The session is in no way associated with the Library instance

> that fixed the DB lock that was occurring presumably due to a mismatched instance being passed from the migrations that was holding the lock.

If you compare how the session is built now and how it was built then, they are exactly identical, so this cannot be the reason afaict.

>  These changes also enforce using Library instances as a single unit instead of passing around `session` contexts that can become disjointed

Not quite sure what you mean here

src/tagstudio/core/library/alchemy/library.py
Computerdores: > > This split existed, because it is being used in the migrations where the existing engine should be used to ensure consistency irrespective of future changes to the migrations.
> 
> Where might an inconsistency arise that would be solved by passing a different engine to build a session from?

The engine is specific to the DB that is currently open, so if we were to e.g. allow a user to bulk migrate libraries while a different library is open, then this get_version method would cause the migrations to see the incorrect version and not migrate anything because it would see the version of the currently open library and not the one of the DB being migrated.

> > However, I think this can just be moved to the migrations entirely as the only other use can be removed.
> > (It is in LibraryInfoWindow where get_version(DB_VERSION_CURRENT_KEY) is contrasted to DB_VERSION, which is useless because the migrations fail unless get_version(DB_VERSION_CURRENT_KEY) == DB_VERSION meaning that can be removed.)
> 
> The current UI-facing use of this is working as intended, where it shows the current loaded DB version in comparison to the program's currently known `DB_VERSION`, which is useful for telling if you have a DB loaded with a minor version greater than the TagStudio version that's opening it.

Good point, I missed that.

> I'd also like to not gate the `_get_version()` method to the migrations, as that would limit future uses in the UI.
> 
> Perhaps the "static" version of this that includes the engine parameter and inspection conditional could be moved to the migrations class, while a simpler instance method that just uses the currently expected logic could remain in the library file? Though that's a tiny bit of code duplication

Yes, I think that would be best. That way `get_version` is basically a one-liner and the complicated DB introspection logic can be relegated to the migrations.
Also, uncoupling the migrations from sqlalchemy would probably require that anyway, so we might as well do it now.

src/tagstudio/core/library/alchemy/library.py
Computerdores: > The `save_library_backup_to_disk()` method doesn't require a library to be loaded, it just uses the Library instance's `self.library_dir,` which I tweaked to be set before the migrations are run in` open_sqlite_library()`.

Yeah this is exactly what I meant; normally the `library_dir` means "the directory of the currently open library" but here it effectively is "the directory of the library that should be backed up" (unless a library is actually open, in which the original meaning is correct again).

> > We could maybe also just move this to the DBMigrations to be done as the first step of DBMigrations.run.
> 
> This method is also used outside of the migrations, like in the UI - so there still needs to be a non-static version of it present in the Library class for instances to use without passing the argument of the instance's own `library_dir`.

I think having a static method and a non-static method that calls the static one (with `self.library_dir` as the param) would be best then.

CyanVoxel:
> Also, the semantics of save_library_backup_to_disk are to make a backup of the library DB found at a certain path, not to make a backup of the currently opened library. And making the method non-static implies that it could access anything available on the instance when a library is accessed, which could lead to problems, because in order to call it from the migrations only the library path is set and the library is actually opened (which would of course not be possible). This is why I have to object to making save_library_backup_to_disk an instance method.

The `save_library_backup_to_disk()` was previously an instance method and used as such outside of migrations (e.g. the `backup_library()` method in `ts_qt.py`). There can be special cases for things that the new migration system needs, but it shouldn't be assumed that these functions aren't being used elsewhere where the expectation is to just have the Library instance, and then tell that library to save a backup using its own `library_dir`.

> The session is in no way associated with the Library instance

I'm referring to how the session context managers in the Library class are constructed using `with Session(self.engine) as session:`, which use the `self.engine` variable inside the particular Library instance. Reverting the `all_entries()` method to be non-static and restoring the original context manager with `self.engine` fixes the lock on the DB created by "something else, somewhere else". 

> If you compare how the session is built now and how it was built then, they are exactly identical, so this cannot be the reason afaict.

If you've got an alternate explanation for #1447 I'm all ears, but from all the investigating and testing I've done it really seems like the session currently is getting detached or desycned or *something* from the rest of the Library, causing a lock on the DB because there's now something outside the scope of the expected session that's using it (in conjunction with using `autocommit=False`). And restoring the original context manager using the instantiated Library class's `self.engine` resolves that lock issue without having to revert the recommended `autocommit` parameter.

Changes made from there were either due to other obscured issues like the DB9 migration flushing and/or due to wanting to enforce a particular pattern for accessing library methods.

CyanVoxel:
@Computerdores Unfortunately my hands are kind of tied with this, where the reversions of (at least some of) the static methods plus the reversion to use the session associated with a Library instance is fixing this critical issue that needs to ship as soon as it can, and I don't have evidence of an unrelated cause for all this.

The closest I've been able to glean is a similar issue I fixed in dd00f4d4a02fd74b24d40c51b55ae3cc7db09203 where colors couldn't be added due to a dormant double session issue that started causing a DB lock, but I couldn't find evidence of a double session going on here. SQLAlchemy doesn't exactly make it straightforward to trace locks, unfortunately...

We can revisit some of the static discussions and migration parameters in the future, but right now there *needs* to be a fix for #1432/#1456 and this is the only one, which was created on a best-effort basis for this nebulous ORM breakage.

CyanVoxel:
I believe `@racerand` on Discord has uncovered the fundamental cause of the issue: when `all_entries()` was split into two methods, the instanced one that wraps a session had `return Library._all_entries(session, with_joins)` when it *should* have had `yield from Library._all_entries(session, with_joins)` to properly differ to the `Iterator` yielded from the static `_all_entries()`.

This helps quell my concerns about passing sessions around, and with seeing the double session issue in dd00f4d4a02fd74b24d40c51b55ae3cc7db09203 plus a clear example of an internal need to shield library logic from double sessions in [#1470](https://github.com/TagStudioDev/TagStudio/pull/1470/changes#diff-a2274aec4b99e18fee41ce3ea4a3bca8a1d246a2755623f8116a3e0240d9d07eR1755-R1771), I've gained a greater context for what kinds of shifts are needed in the Library class design - keeping in mind that the session/cursor concept will likely stay after moving from SQLAlchemy to raw sqlite3.

I think there's a growing split in what the Library class is thought to be used for now, when previously it was purely a *"Class for the Library object, and all CRUD operations made upon it"*, and having static methods that are intended to be used publicly or psuedo-publicly that don't relate to a Library instance felt inherently against this - but more of these types of operations are becoming necessary, including by Library instances themselves. One direction things could take is splitting the static, non-context-managed logic out into a *separate* class (`LibraryOps` or something) that is then referenced by both the Migrations class and Library instances whenever they need to.

***

So with that in mind, I think I'm going to do a quick commit to main with the `yield from` change, make a small PR with some of the small misc fixes and cleanup included in here, and leave the rest of the big ticket changes for a lower priority PR targeting 9.6.4.

Computerdores:
> I believe `@racerand` on Discord has uncovered the fundamental cause of the issue: when `all_entries()` was split into two methods, the instanced one that wraps a session had `return Library._all_entries(session, with_joins)` when it _should_ have had `yield from Library._all_entries(session, with_joins)` to properly differ to the `Iterator` yielded from the static `_all_entries()`.

So I wasn't going insane! Although I would have expected a warning from pyright for this kind of thing, but whatever

> I think there's a growing split in what the Library class is thought to be used for now, when previously it was purely a _"Class for the Library object, and all CRUD operations made upon it"_, and having static methods that are intended to be used publicly or psuedo-publicly that don't relate to a Library instance felt inherently against this - but more of these types of operations are becoming necessary, including by Library instances themselves. One direction things could take is splitting the static, non-context-managed logic out into a _separate_ class (`LibraryOps` or something) that is then referenced by both the Migrations class and Library instances whenever they need to.

I would suggest calling it `_LibraryOps` and putting it somewhere in `tagstudio.core.library.alchemy` to communicate that it isn't meant for use outside of that.

> So with that in mind, I think I'm going to do a quick commit to main with the `yield from` change, make a small PR with some of the small misc fixes and cleanup included in here, and leave the rest of the big ticket changes for a lower priority PR targeting 9.6.4.

Good decision, even before racerand's insight; much as I dislike pushing a fix I don't understand (must be the security part of my brain) I also think that it is more important to have it working than for the code to be beautiful.

PR #1495, opened 2026-08-28
refactor: library scanning

TrigamDev:
### Summary

- Library scanning is reworked to get a list of all entry paths found in the library *before* scanning, rather than checking for each individual scanned file. This cuts down dramatically on the number of SQL queries created and improves scanning performance.
- Previously, each scanner would be responsible for then also handling and running logic on the scanned files. Now, each scanner is only responsible for producing a list of discovered paths. The logic run on each file is instead placed in `refresh_dir()`, which consumes the files produced by which ever scanner is used.
- The logic behind which scanner is used is moved into `scan_for_files()`. This way, any function that needs to scan for files can just blindly call `scan_for_files()` without needing to determine which scanner should be used, and it will receive an iterator of files.
- Each scanner method has been renamed to be clearer.
  - To help with maintainability, I would recommend *not* shortening names arbitrarily. Good lord, the source code is filled with so many 2 or 3 character names.
- Renamed the `wcmatch` scanner to simply "internal". This better reflects its purpose, and helps if the implementation of the internal scanner ever changes down the line.
- Reworked the implementation of the ripgrep scanner to better process the discovered files *as* they're scanned, rather than scanning them all beforehand and only then processing them.
- Changed how scanning is timed to include the time it takes for the ripgrep scanner to.. scan. Before, it would only track the time it took to *process* the files, excluding the time it took to scan them.
- Made the scanner methods (and `scan_for_files()`) static, as they don't rely on any attributes of `RefreshTracker`. 

### Notes

- This conflicts with #1242. If #1242 is merged first, I can rewrite this PR and rebase it on #1242, since this is a fairly simple PR.
- `ignore_to_glob()` produces a *lot* of redundant, recursive patterns that slow down the internal scanner so much that it's quite literally unusable on my primary library (~65k files, with a large ignore file). I was never able to get it to produce any output on my primary library. It may be worth looking into improving the translation from gitignore-style to glob, or replacing the internal scanner implementation altogether.
- It could be worth having `refresh_dir()` emit more detailed progress, rather than just an `int`.
- `refresh_dir()` is set to only emit output every 1/30th of a second.
  - Is there a reason `refresh_dir()` is the one responsible for enforcing this limit? At least to me personally, it'd make more sense to place the responsibility on the UI for limiting the rate of UI updates. Especially if another, rapidly-updating data source were connected to a progress UI.
  - It may be a good idea to create a general utility method for enforcing an update limit (say, it accepts an iterator, and returns an iterator limited to a certain update limit).
- It could be worth having ripgrep use null separators, rather than new lines, to prevent issues with certain file names.

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [x] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/src/tagstudio/core/library/refresh.py
+++ b/src/tagstudio/core/library/refresh.py
@@ -134,23 +134,29 @@ def _scan_with_ripgrep(scan_dir: Path, ignore_patterns: list[str]) -> Iterator[P
         ],
         cwd=scan_dir,
         stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
+        stderr=subprocess.DEVNULL,
         text=True,
     )
 
     assert process.stdout is not None
 
+    completed = False
+    return_code: int | None = None
     try:
         for line in process.stdout:
             yield scan_dir / line.rstrip("\n")
+        completed = True
     finally:
-        stderr = process.stderr.read() if process.stderr else ""
         process.stdout.close()
+
+        if not completed and process.poll() is None:
+            process.terminate()
+
         return_code = process.wait()
         compiled_ignore_path.unlink(missing_ok=True)
 
     if return_code != 0:
-        raise RuntimeError(f"ripgrep scan failed with exit code {return_code}: {stderr}")
+        logger.error(f"[Refresh] ripgrep scan failed with exit code {return_code}")
 
 
 def _scan_with_internal_scanner(scan_dir: Path, ignore_patterns: list[str]) -> Iterator[Path]:
--- a/src/tagstudio/core/library/refresh.py
+++ b/src/tagstudio/core/library/refresh.py
@@ -3,6 +3,7 @@
 
 
 import shutil
+import subprocess
 from collections.abc import Iterator
 from dataclasses import dataclass, field
 from datetime import datetime as dt
@@ -15,7 +16,6 @@ from wcmatch import pathlib
 from tagstudio.core.library.alchemy.library import Library
 from tagstudio.core.library.alchemy.models import Entry
 from tagstudio.core.library.ignore import PATH_GLOB_FLAGS, Ignore, ignore_to_glob
-from tagstudio.core.utils.silent_subprocess import silent_run  # pyright: ignore
 
 logger = structlog.get_logger(__name__)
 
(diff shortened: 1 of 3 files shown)


CyanVoxel (commented):
Oddly this significantly *slows* down scanning performance for me, on my NAS library at least... scanning a local library does seem to be much faster. However this also removes the cache of previously scanned files, so subsequent scans are no longer lightning fast and take up the full scan time again.

I don't recall #1242 + [this interactive subprocess comment](https://github.com/TagStudioDev/TagStudio/pull/1242#pullrequestreview-4621614493) creating a slowdown, so it may be worth comparing what #1242 is doing differently (without creating the same normalization issue there). Also, while I haven't tested this in compiled builds, I'd lean towards using the silent subprocess method we have to avoid spawning CMD windows in builds (I'm not sure if `ripgrep` will do that, but other subprocesses like `ffmpeg` do normally).
***
> `ignore_to_glob()` produces a lot of redundant, recursive patterns that slow down the internal scanner so much that it's quite literally unusable on my primary library

I agree that it produces a lot of redundant patterns. It was the best I could do at the time in order for it to match the functionality of `.gitignore`, and am open to improvements to it.

> Is there a reason refresh_dir() is the one responsible for enforcing this limit? At least to me personally, it'd make more sense to place the responsibility on the UI for limiting the rate of UI updates. Especially if another, rapidly-updating data source were connected to a progress UI.

> It may be a good idea to create a general utility method for enforcing an update limit (say, it accepts an iterator, and returns an iterator limited to a certain update limit).

If there's a good efficient way to restrict the rapid pinging from the backend side on the UI side, I'm open to it

> It could be worth having ripgrep use null separators, rather than new lines, to prevent issues with certain file names.

Sounds good if straightforward

TrigamDev:
Sorry, forgot to respond.

> Oddly this significantly slows down scanning performance for me, on my NAS library at least... scanning a local library does seem to be much faster.

That's.. really interesting, I wonder why that is.

> However this also removes the cache of previously scanned files, so subsequent scans are no longer lightning fast and take up the full scan time again.

.. oops. I can't remember my exact thought process due to it being several weeks ago, but I must've misunderstood the purpose of `library.included_files`/mixed it up with `library.get_paths()` and assumed it was redundant, or just forgot about it.

> Also, while I haven't tested this in compiled builds, I'd lean towards using the silent subprocess method we have to avoid spawning CMD windows in builds (I'm not sure if `ripgrep` will do that, but other subprocesses like `ffmpeg` do normally).

I wasn't aware of the silent subprocess method while working on the PR, and wasn't aware subprocesses created CMD windows on other platforms since I use Linux.

> > It could be worth having ripgrep use null separators, rather than new lines, to prevent issues with certain file names.
>
> Sounds good if straightforward

ripgrep has a `-0` flag for using null separators between file paths, and on the Python side, it should be possible to split on `b'\0'`.

TrigamDev:
In any case, closing this PR since it's being superseded by #1501.

PR #1497, opened 2026-08-31
Fix exceptions in folders to tags

Huluk:
### Summary

When running "Folders to Tags", the folder list is empty and no tags can be created. This change fixes that issue.

### Tasks Completed

- gracefully handle empty parent tag
- avoid database lock by first creating a list of entries, then writing new tags

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [x] macOS ARM
    - [ ] Linux x86
    - [ ] Linux ARM
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable 
      

### Context

I can see that "Folders to Tags" is deprecated, but I don't see how I can do the same with other existing functionality, and this was the very first feature I tried after installing TagStudio, which kind of gave a bad impression. Sorry for maybe not knowing all your conventions, I'm new to the project and only skimmed the style guide for this small change.

### DIFF
--- a/src/tagstudio/qt/mixed/folders_to_tags.py
+++ b/src/tagstudio/qt/mixed/folders_to_tags.py
@@ -74,7 +74,7 @@ def folders_to_tags(library: Library):
         reversed_tag = reverse_tag(library, tag, None)
         add_tag_to_tree(reversed_tag)
 
-    for entry in library.all_entries():
+    for entry in list(library.all_entries(with_joins=True)):
         folders = entry.path.parts[0:-1]
         if not folders:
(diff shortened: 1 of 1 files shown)


CyanVoxel (approved):
Thank you so much for this fix! While this specific feature has historically been under-maintined is slated to be replaced with the [macros](https://github.com/TagStudioDev/TagStudio/tree/macros) feature eventually, macros have continued to be pushed back so it's very nice to have this fix in the meantime.

PR #1499, opened 2026-09-03
refactor(migrations): sqlalchemy be gone

Computerdores:
### Summary

Completely evict SQLAlchemy from the migrations and instead implement it in raw sqlite3.

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [ ] Linux x86
    - [ ] Linux ARM
- Tested For:
    - [ ] Basic functionality
    - [ ] PyInstaller executable

### DIFF
--- a/src/tagstudio/core/library/alchemy/migrations.py
+++ b/src/tagstudio/core/library/alchemy/migrations.py
@@ -315,16 +315,14 @@ class MigrationTo103(DBMigration):
 
     @override
     @classmethod
-    def run(cls, session: Session, library_dir: Path, fmt_log: LoggingMethod):
+    def run(cls, conn: Connection, library_dir: Path, fmt_log: LoggingMethod):
         """Migrate DB from DB_VERSION 102 to 103."""
         # add the new hidden column for tags
-        session.execute(text("ALTER TABLE tags ADD COLUMN is_hidden BOOLEAN NOT NULL DEFAULT 0"))
-        session.flush()
+        conn.execute("ALTER TABLE tags ADD COLUMN is_hidden BOOLEAN NOT NULL DEFAULT 0")
         logger.info(fmt_log("Added is_hidden column to tags table"))
 
         # mark the "Archived" tag as hidden
-        session.query(Tag).filter(Tag.id == TAG_ARCHIVED).update({"is_hidden": True})
-        session.flush()
+        conn.execute("UPDATE tags SET is_hidden = true WHERE id = ?", [TAG_ARCHIVED])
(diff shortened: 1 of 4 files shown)


Computerdores:
I have done all that I want to do in the way of refactors here at this point, so can take a look if you want @CyanVoxel.
Although I have not specifically tested the migrations beyond making sure the tests pass (no f'in clue what's happening on windows again), there should be relatively little potential for bugs in these refactors.

src/tagstudio/core/library/alchemy/migrations.py
```diff
@@ -102,167 +103,152 @@ def run(self):
...
-                        logger.info(
-                            f"[Library][Migration][{migration.version}] "
-                            "Couldn't update version, continuing without commit",
-                            error=e,
-                        )
-                        session.flush()
-                    else:
-                        session.commit()
+                else:
+                    self._connection.commit()
```
CyanVoxel, line 139: There's no context manager or explicit `.close()` statement anywhere, so the connection seems to be staying open. My guess is that's what's causing the Windows permission errors. I'd encourage wrapping this back in a context manager along the lines of `with self._connection as connection:` or otherwise calling `.close()` on the connection when finished, and we can see if that also fixes the tests

Computerdores: the windows failure was not windows bs, colour me genuinely surprised ^^

Computerdores:
~~note to self: fix the todo comments added here~~

Computerdores:
finally got the motivation to finish this.
All todos are done, should be ready for review

src/tagstudio/core/library/alchemy/migrations.py
```diff
@@ -360,11 +358,17 @@ def __migrate_sql_to_ts_ignore(cls, session: Session, library_dir: Path):
...
+            )[0]
         )
         is_exclude_list: bool = unwrap(
-            session.scalar(text("SELECT value FROM preferences WHERE key = 'IS_EXCLUDE_LIST'"))
+            conn.execute("""
+                SELECT value
+                FROM preferences
+                WHERE key = 'IS_EXCLUDE_LIST'
+            """).fetchone()[0]
         )
```
CyanVoxel, line 372: Should the `[0]` be after the `unwrap()` call here?
```suggestion
            """).fetchone()
)[0]
```

CyanVoxel (approved):
Approved when the one nit is addressed, thank you for your work on this!

Computerdores:
done

PR #1506, opened 2026-09-18
feat: Expand tarball thumbnail support

### LINKED ISSUE #1505: [Feature Request]: Expand Tarball Thumbnail support
### Checklist

### Description

TagStudio currently only supports thumbnails for a limited set of `tar` based files, namely
`.tar`, `.cbt` and `.tgz`.

The `tarfile` module used to extract the thumbnail also supports LZMA, BZip2 and Zstandard
compressed tarballs, their extensions just need to be registered with the renderer.

### Solution

Register the extensions with the archive thumbnail renderer

### Alternatives

continue as is

Sola-ris:
### Summary

Add support for more types of tarballs.
Closes https://github.com/TagStudioDev/TagStudio/issues/1505

As discussed on discord, I haven't registered the compressor extensions like `.gz` with the archives in `file_groups.py`.
One side effect of that is, that archives without thumbnails get the default file icon instead of the one for archives.

#### Before
before

#### After
after

#### Files used for testing
[archives.zip](https://github.com/user-attachments/files/32396059/archives.zip)

### Tasks Completed

- Platforms Tested:
    - [ ] Windows x86
    - [ ] Windows ARM
    - [ ] macOS x86
    - [ ] macOS ARM
    - [x] Linux x86
    - [ ] Linux ARM
      
- Tested For:
    - [x] Basic functionality
    - [ ] PyInstaller executable


src/tagstudio/previews/renderers/archive.py
```diff
@@ -32,13 +32,20 @@ class ArchivePreview(BasePreview):
     @classmethod
     def register_types(cls) -> None:
         # NOTE: Filetype equivalents (i.e. ".tar.gz" == ".tgz") are already declared internally.
-        MediaTypes.register("archive", ".7z", RENDER)
+        MediaTypes.register("archive", [".7z", ".s7z"], RENDER)
```
CyanVoxel: I'd go ahead and consolidate these two as equivalent inside `file_groups.py` instead, and then just reference `".7z"` here

